PR #4, opened 2020-09-01
Set PVC to read write many

dprince:
This should allow us to launch multiple glance pods


mdbooth (approved):
Looks like this is what Glance recommends for file backend:

https://docs.openstack.org/kolla-ansible/latest/reference/shared-services/glance-guide.html

stuggi (approved):
LGTM, we only need to make sure/document the pvc is a shared storage

PR #5, opened 2020-09-02
Add config-data-custom to allow customize service

stuggi:
This change allows to customize the glance-api service using a
<cm>-custom configmap. The tool crudini is used to merge the configs
into the default .conf file coming with the OSP service container
image. The flow is:
  container.conf < config-data default < config-data custom

With this:
* 3 CMs per service
  - -scripts, mounted to /usr/local/bin/container-scripts
  - -config-data, which have the rendered templates from operator
  - -config-data-custom, which can be used to customize the service
* per default there is only one resulting config file for the service.
* all modifications to config-data default get reverted by
  the operator
* all modifications to config-data-custom won't get modified
  by the operator and is the admins responsibility for correctness

In the init container -config-data gets mounted to /var/lib/config-data/default,
-config-data-custom to /var/lib/config-data/custom. The init.sh
from -scripts:
1) copies the default service config from /etc/... to config-data/merged
   which is an empty volume.
2) merges all configs from config-data/default in to config-data/merged.
   If the same file already exists and is not a json file, it is expected
   right now that crudini can be used to merge the configs. If it is a json
   file, or the config don't exist, the file gets copied
3) does the same as in 2) but merges config-data/custom into config-data/merged
4) sets secret information in the service config

The main service container in the end only has/requires the config-data/merged
volume to start the service.

This also adds KOLLA_CONFIG_FILE env parameter to the service config to
point to the kolla config in config-data/merged. With this we can even
modify the kolla config using the CM. This has the benefit that we can
- change the command to start the service for debugging, or patching?
  the services
- merge add additional config files into the container using kolla


stuggi:
I have an idea to make this more generic. will post an update later today or on monday

dprince (approved):

dprince:
@stuggi Should the customerConfigMap's parameter be an array? I like the simplicity of having only 1 but there might be cases where having an array would be handy for layering things (last one wins maybe).

stuggi:
requires https://github.com/openstack-k8s-operators/lib-common/pull/8

stuggi:
> @stuggi Should the customerConfigMap's parameter be an array? I like the simplicity of having only 1 but there might be cases where having an array would be handy for layering things (last one wins maybe).

I am not sure I understand the use case to have multiple, also with my new change, we could now add additional config files in the customConfigMap, even could customize the used kolla config. maybe that removes the need? check the updated PR description.

PR #18, opened 2022-07-04
Modernize Glance operator

abays:
Moving old Glance operator to the new structure established by the team


pkg/glance/pvc.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+
+	glancev1beta1 "github.com/openstack-k8s-operators/glance-operator/api/v1beta1"
+	"github.com/openstack-k8s-operators/lib-common/pkg/common"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// Pvc - creates and returns a PVC object for a backing store
+func Pvc(ctx context.Context, r common.ReconcilerCommon, api *glancev1beta1.GlanceAPI, labels map[string]string) (*corev1.PersistentVolumeClaim, error) {
```
stuggi, line 14: We should add generic functions to lib-common, like we have for deployment [1] that you can use it like pvc := NewPVC(pvc, p2, p3) . Then pvc.CreateOrPatch(), pvc.Delete(), ... . As this is required by multiple services would be good to move this to the new way to create/manage resources. In here we'd then add the raw PVC for glance.

[1] https://github.com/openstack-k8s-operators/lib-common/blob/master/pkg/common/deployment.go#L34

templates/glanceapi/config/glance-api-config.json
```diff
@@ -0,0 +1,17 @@
+{
+    "command": "/usr/bin/glance-api --config-file /etc/glance/glance-api.conf",
```
stuggi, line 2: should add --config-dir=/etc/glance/glance.d and then have custom.conf?
```
        {
            "source": "/var/lib/config-data/merged/custom.conf",
            "dest": "/etc/glance/glance.d/custom.conf",
            "owner": "glance",
            "perm": "0600"
        },
```
https://docs.openstack.org/glance/latest/configuration/configuring.html

controllers/glanceapi_controller.go
```diff
@@ -78,250 +107,207 @@ func (r *GlanceAPIReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+func (r *GlanceAPIReconciler) reconcileInit(
+	ctx context.Context,
+	instance *glancev1beta1.GlanceAPI,
+	helper *helper.Helper,
+	serviceLabels map[string]string,
+) (ctrl.Result, error) {
+	r.Log.Info("Reconciling Service init")
+
+	// PVC
+	// TODO: "lib-commonalize" this
```
stuggi, line 196: ah you already have that on the list :)

pkg/glance/volumes.go
```diff
@@ -73,35 +78,22 @@ func getInitVolumeMounts() []corev1.VolumeMount {
 			MountPath: "/var/lib/config-data/default",
 			ReadOnly:  true,
 		},
-		{
-			Name:      "config-data-custom",
-			MountPath: "/var/lib/config-data/custom",
```
stuggi, line 78: shouldn't we keep this, or do you merge also the custom configs into main service conf? can start glance with --config-dir=DIR as the default is none/

templates/glanceapi/config/glance-api-config.json
abays: Somehow I missed copying this pattern from placement.  Will add.

pkg/glance/volumes.go
abays: So I was just following the pattern we're now using the new Keystone and Placement operators, where the custom config is actually provided in-line within the CR and then added to the config data CM like so: https://github.com/abays/glance-operator/blob/af3f68396b87827483fe660600fb6f74516731eb/controllers/glanceapi_controller.go#L504-L542.  I guess I need to mount a volume pointing to the `custom.conf` key in the config data CM, whereas in the old Glance operator there was an actual separate "custom config data" CM.

templates/glanceapi/config/glance-api-config.json
stuggi: yes thats how we used it in cnosp. there was one concern in a meeting some week ago. some openstack services support having multiple same parameters in the conf. nova is one of them. Therefore I changed how we used it there to just merge the required parameters into the main service config and conf snipped provided via CustomServiceConfig into the conf.d of the service. if the service supports multiple same parameters you could do that with that. 

Now I see "If --config-dir is set, then --config-file is ignored." , but it looks like it supports using multiple --config-file parameters. Have you tried that using `/usr/bin/glance-api --config-file /etc/glance/glance-api.conf  --config-file /etc/glance/glance.conf.d/custom.conf` ?

abays: That does work, but I'm not sure how it respects order for overwrites.  Back to your other thought though, re: `conf-dir`, we could just put `00-glance-api.conf` and `01-custom.conf` in `glance.conf.d` dir and use the numbering to ensure processing order.  Thoughts?

stuggi: yes that would work. could then not merge at all and just add the required to 00 and customization to 01 where the service supports a conf dir, but then we should do that for all components where a conf dir is supported to have the same behavior.

re the ordering, from `podman run -it -u root --rm quay.io/tripleotraincentos8/centos-binary-glance-api:current-tripleo glance-api --help` -> "Values in latter files take precedence"

```
  --config-file PATH    Path to a config file to use. Multiple config files
                        can be specified, with values in later files taking
                        precedence. Defaults to None. This option must be set
                        from the command-line.
```

api/v1beta1/glanceapi_types.go
```diff
@@ -17,36 +17,151 @@ limitations under the License.
...
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
-// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+const (
+	// GlanceFinalizer allows GlanceAPIReconciler to clean up resources associated with GlanceAPI before
+	// removing it from the apiserver.
+	GlanceFinalizer = "glanceapi.osp-director.openstack.org"
```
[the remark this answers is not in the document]
abays: Ack, fixed

templates/glanceapi/config/glance-api-config.json
stuggi: ack sounds good. maybe we see what works for all/most of the services and do the same everywhere.

stuggi (approved):
/lgtm

PR #29, opened 2022-08-28
Add external Ceph cluster support

fmount:
This patch introduces an API extension to support an external Ceph cluster in
the glance operator.
The `GlanceBackend` is computed accordingly, and if Ceph is present (which means that
`CephMonHosts`, `CephClusterFSID` and `CephClienKey` have been defined), the `rbd`
backend is selected, otherwise `file`, which is the default, is still used.
The API struct has been extended and now includes the CephBackend parameters that
can be used to set the "rbd" backend when the config file is built.


fmount:
@gfidente FYI

controllers/glanceapi_controller.go
```diff
@@ -639,16 +668,30 @@ func (r *GlanceAPIReconciler) generateServiceConfigMaps(
 
 	keystoneAPI, err := keystone.GetKeystoneAPI(ctx, h, instance.Namespace, map[string]string{})
 	if err != nil {
-		return err
+		errs = append(errs, err)
```
abays, line 671: This and similar checks here should remain `return err`.  We can't continue and assume we have the associated `keystoneAPI` object, for instance, if there was in fact an error.  

Did you see the `k8serrors.NewAggregate` pattern elsewhere (other operators in other projects)?  I have not seen that approach taken in operators before.

fmount: I see your point: I created this error within the `generateServiceConfigMap` function to make sure any issue related to any component in the `ConfigMap` generation is returned and printed (as an error) within the operator log, otherwise I didn't see anything reporting error in the operator.
This can be useful during the development, but also to understand where is the issue.
We used this kind of pattern in [1], but I understand we'd like to rework the way we collect and print error logs.

[1] https://github.com/openshift/csi-driver-manila-operator/

controllers/glanceapi_controller.go
```diff
@@ -609,6 +612,31 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
...
+	// Validate parameters to enable CephBackend
+	if instance.Spec.CephBackend.CephClusterFSID != "" &&
+		r.validateMons(instance.Spec.CephBackend.CephClusterMonHosts) &&
+		instance.Spec.CephBackend.CephClientKey != "" {
+		return "rbd"
+	}
+	/* file is the default backend: if we don't have any ceph
+	parameter set backend to file */
+	return "file"
+}
```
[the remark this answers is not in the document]
fmount: Ack and thanks for your feedback.
Given those are `utility` functions, I guess it's better moving them to `pkg/glance` and keep this section as slim as possible.
As you said, this is useful to make sure the controller has no clues about the underlying CR lifecyle and can be a good example to build other OSP operators prototypes.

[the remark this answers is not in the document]
fmount: I guess this can be resolved, now those functions are moved and documented in `pkg/glance`, where only SetBackend is exposed.
In the future, as long as more backends are supported, we can think about adding a util.go, but that's pure code organization/refactoring that can be a follow up work.

controllers/glanceapi_controller.go
```diff
@@ -671,11 +689,12 @@ func (r *GlanceAPIReconciler) generateServiceConfigMaps(
...
 	}
 	err = configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
+
 	if err != nil {
-		return nil
+		errs = append(errs, err)
 	}
 
-	return nil
+	return k8serrors.NewAggregate(errs)
```
abays, line 697: I think these can be reversed too.  I will look into the operator not logging the individual errors that you mentioned.

api/v1beta1/glanceapi_types.go
```diff
@@ -85,6 +85,11 @@ type GlanceAPISpec struct {
 	// actual action pod gets started with sleep infinity
 	Debug GlanceAPIDebug `json:"debug,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// CephBackend - The ceph Backend structure with all the parameters
+	// TODO: -> implement a backend wrapper in a generic structure
```
dprince, line 90: @fmount what is your though on the generic structure here. Is the JSON here somewhat free-form (could be anything) or is it something we could model in a Golang struct to verify it is in the right form ahead of time?

fmount: Hi @dprince,
IMO in the current status this is free form: I left this TODO to brainstorm around a well defined wrapper where multiple backends can live within the same json structure.
It will be useful (and required) for multi-store support, and the same we'll see the same problem in cinder, but it can be part of a follow up work.

PR #30, opened 2022-08-31
Do not wait for PVC to reach bound state

abays:
We currently wait until the associated PVC is bound before proceeding with further Glance component deployment.  This is a problem for `WaitForFirstConsumer` storage classes, however, as the consumer of the PVC is the PV that is created for the Glance pod -- but that PV won't be created until the pod is being created, and we're blocking that creation by waiting for the PVC to bind.  We should just create the PVC and move on instead, allowing the native k8s PVC/PV/Pod reconciliations to work themselves out.

### DIFF
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -238,11 +238,6 @@ func (r *GlanceAPIReconciler) reconcileInit(
 	} else if (ctrlResult != ctrl.Result{}) {
 		return ctrlResult, nil
 	}
-
-	if pvc.GetPvc().Status.Phase != corev1.ClaimBound {
-		r.Log.Info("Waiting for GlanceAPI PVC to bind")
-		return ctrl.Result{RequeueAfter: time.Second * 5}, nil
-	}
 	// End PVC creation/patch
 
 	//


controllers/glanceapi_controller.go
```diff
@@ -238,11 +238,6 @@ func (r *GlanceAPIReconciler) reconcileInit(
 	} else if (ctrlResult != ctrl.Result{}) {
 		return ctrlResult, nil
 	}
-
-	if pvc.GetPvc().Status.Phase != corev1.ClaimBound {
```
fmount, line 242: LGTM.
This actually solves the issue I saw, and I guess it's safe enough removing this check given the low complexity of the logic that manages the `pvc` (where we don't have particular constraints).
This change also work in a `SoS` environment, and avoid us to change any storageClass parameter (mostly the `VolumeBindingMode` from `WaitForFirstConsumer` to `Immediate`). 
In particular, in a `SoS` environment where `CSO` manages and reconciles the storageClasses parameters,  it's not safe (and it's not even supported) moving the storageClass to an `unmanaged` status, hence this change is important to make sure we can use any other backend.

Here an example of the pvc, bound to the default storageclass,  and the glance pod running:

```
> oc get sc
NAME              PROVISIONER                    RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
csi-manila-ceph   manila.csi.openstack.org       Delete          Immediate              false                  2d15h
local-storage     kubernetes.io/no-provisioner   Delete          WaitForFirstConsumer   false                  24m
standard          kubernetes.io/cinder           Delete          WaitForFirstConsumer   true                   2d15h
standard-csi      cinder.csi.openstack.org       Delete          WaitForFirstConsumer   true                   2d15h
```

```
> oc get pvc
NAME                STATUS    VOLUME             CAPACITY   ACCESS MODES   STORAGECLASS    AGE
glance              Pending                                                standard-csi    13s
mariadb-openstack   Bound     local-storage003   10Gi       RWO,ROX,RWX    local-storage   109m
```

```
NAME                      READY   STATUS    RESTARTS   AGE
glance-647f54679b-xn89s   1/1     Running   0          5m34s
```

fmount (approved):

fultonj:
I tested this PR and I no longer hit the problem I mentioned in the following with it. LGTM. Thanks. 

https://github.com/openstack-k8s-operators/install_yamls/pull/23

PR #31, opened 2022-09-02
Make sure the glance service exists before deleting it

fmount:
There are cases during the reconciliation loop where keystone is not
able to expose the glance service and the related endpoints because
it receives a delete call for the mentioned objects.
This change introduces the same check added in the cinder-operator
that looks for the `APIEndpoints` in the instance `Status` before
deleting/removing the finalizer. 

### DIFF
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -205,6 +205,16 @@ func (r *GlanceAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
 func (r *GlanceAPIReconciler) reconcileDelete(ctx context.Context, instance *glancev1beta1.GlanceAPI, helper *helper.Helper) (ctrl.Result, error) {
 	r.Log.Info("Reconciling Service delete")
 
+	// It's possible to get here before the endpoints have been set in the status, so check for this
+	if instance.Status.APIEndpoints != nil {
+		ks, _ := keystone.GetKeystoneServiceWithName(ctx, helper, glance.ServiceName, instance.Namespace)
+		//Ensure the service exists
+		ksSvc := keystone.NewKeystoneService(ks.Spec, instance.Namespace, map[string]string{}, 10)
+		err := ksSvc.Delete(ctx, helper)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
 	// Service is deleted so remove the finalizer.
 	controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
 	r.Log.Info("Reconciled Service delete successfully")


fmount:
We experienced this weird behavior during the testing/development of this component.
After some debugging we found something weird in keystone logs [0] that has points to
that part of the reconciliation loop. This change basically introduces the same check
already present in the cinder-operator [1] that fixes the same issue.

[0] https://paste.opendev.org/show/bo58ORuHoMp5lfyBR4hE/
[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/master/controllers/cinderapi_controller.go#L253-L274

controllers/glanceapi_controller.go
```diff
@@ -205,6 +205,26 @@ func (r *GlanceAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
 func (r *GlanceAPIReconciler) reconcileDelete(ctx context.Context, instance *glancev1beta1.GlanceAPI, helper *helper.Helper) (ctrl.Result, error) {
 	r.Log.Info("Reconciling Service delete")
 
+	// It's possible to get here before the endpoints have been set in the status, so check for this
+	if instance.Status.APIEndpoints != nil {
```
dprince, line 209: What is the issue you are trying to address here. The comment on this PR says make sure the sure the service and endpoints are exposed. But it looks like here you are more trying to ensure an explicit cleanup order is followed.

Also, in the case of the KeystoneService is the ownership which we already set not enough to delete that resource and trigger it to be reconciled/deleted.

controllers/glanceapi_controller.go
```diff
@@ -205,6 +205,26 @@ func (r *GlanceAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		ksSvcSpec := keystonev1beta1.KeystoneServiceSpec{
+			ServiceType:        glance.ServiceType,
+			ServiceName:        glance.ServiceName,
+			ServiceDescription: "Glance Service",
+			Enabled:            true,
+			APIEndpoints:       instance.Status.APIEndpoints,
+			ServiceUser:        instance.Spec.ServiceUser,
+			Secret:             instance.Spec.Secret,
+			PasswordSelector:   instance.Spec.PasswordSelectors.Service,
+		}
```
dprince, line 219: Rather than explicitly listing all the parameters here if you do need this you could simply lookup the KeystoneService by name right?

controllers/glanceapi_controller.go
fmount: Sorry, I guess I've not clearly explained the issue we found.
During the development / testing process we found that the `image` service has never been created in keystone, and the endpoints never published, while the pod was up and running and both `service` and `routes` reachable via `curl`.
In a  multiple `deploy-and-cleanup` of the `GlanceAPI` `CRD`, we found that this was actually I side effect caused by the `reconcileDelete` loop that somehow doesn't properly cleanup the resources, hence this patch that solves the problem trough an explicit cleanup flow.

fultonj: Thanks @fmount, With this patch I do not hit the problem of missing glance end points that I ran into earlier.

Details:
The issue I originally ran into was missing entries for glance in keystone.

If I install glance by doing the following then everything is fine:
```
cd ~/install_yamls
make glance
make glance_deploy
```
If I then want to test a patch to my clone of glance-operator, I did the following:

Clean up:
```
cd ~/install_yamls
vim Makefile # to preserve ~/install_yamls/out/openstack/glance/cr
make glance_cleanup
oc delete deployment glance -n openstack
oc delete pvc glance
```
Deploy my variation:
```
cd ~/glance-operator
make build
OPERATOR_TEMPLATES=$PWD/templates ./bin/manager
oc kustomize ~/install_yamls/out/openstack/glance/cr | oc apply -f -
```
I then saw glance and glance-operator pods and I could curl to glance but didn't have keystone endpoints. I worked around it by manually creating the end points and could then use glance (e.g. upload an image).

However, when I repeat the above process with the new patch, then I don't have this issue.

abays: This pattern is repeated in several operators (I think), so this fix might be needed elsewhere beyond Glance/Cinder operators

fultonj: Makes sense to me. Here's an example from cinderapi_controller:

https://github.com/openstack-k8s-operators/cinder-operator/blob/master/controllers/cinderapi_controller.go#L253-L274

fultonj (approved):

controllers/glanceapi_controller.go
fmount: > This pattern is repeated in several operators (I think), so this fix might be needed elsewhere beyond Glance/Cinder operators

Thanks @abays @fultonj, I'll look for this pattern and propose a fix on other operators as well.

fmount: After a quick search I see this code is present on all the operators that implement a `reconcileDelete` in their api controller except glance: it can be improved as per @dprince suggestion (lookup the keystone service by name), but it's only missing here.

PR #32, opened 2022-09-06
Add cephUser and cephPools to cephBackend

fultonj:
Extend API to support cephUser and cephPools within the
cephBackend struct. Without this patch the operator can
only configure glance to connect to a Ceph cluster which
has the cephx user set to 'openstack' and the pool name
set to 'images'. Instead, support any user or pool name.

Signed-off-by: John Fulton <fulton@redhat.com>


fultonj:
@fmount @abays would you please review? thanks

templates/glanceapi/config/ceph.conf
```diff
@@ -1,3 +1,6 @@
 [global]
 fsid = {{ .CephClusterFSID }}
 mon_host = {{ .CephClusterMonHosts }}
+
+[client.{{ .CephUser }}]
+keyring = /etc/ceph/ceph.client.openstack.keyring
```
fmount, line 6: s/openstack/{{ .CephUser }}

fmount (changes requested):
+1 in general.
We might want to update the glance operator as long as the Ceph module is in place [1], but for now I think it's good enough.

[1] https://github.com/openstack-k8s-operators/lib-common/pull/51

templates/glanceapi/config/ceph.conf
fultonj: If I s/openstack/{{ .CephUser }} then I'll need to make the filename variable too as it's hard coded as:

  templates/glanceapi/config/ceph.client.openstack.keyring

E.g. If I do change it I'd need to update `templates/glanceapi/config/glance-api-config.json` too. 

My test of this with our external cluster which uses different pool and cephx user names succeeded but required lines 5 and 6. That's how glance figures out which keyring to use. As per the glance doc "A keyring must be set for this user in the Ceph configuration file, e.g. with a user" 

https://docs.openstack.org/glance/latest/configuration/configuring.html#configuring-the-rbd-storage-backend

fmount: @fultonj I did it here [1], and the src file is renamed as expected and accessible from the container. But if not required here, I'm fine with this.

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/master/templates/cinder/config/cinder-volume-config.json#L30

fultonj: @fmount OK, thanks for the suggestion. I've updated accordingly. 

https://github.com/openstack-k8s-operators/glance-operator/pull/32/files#diff-a68e93cd8bb32386db24d673a79424702323b2173163359eead0c2c72cc48bf6

fmount: +1 thanks John!

templates/glanceapi/config/glance-api-config.json
```diff
@@ -27,7 +27,7 @@
       },
       {
         "source": "/var/lib/config-data/merged/ceph.client.openstack.keyring",
-        "dest": "/etc/ceph/ceph.client.openstack.keyring",
+        "dest": "/etc/ceph/ceph.client.{{ .CephUser }}.keyring",
```
fmount, line 30: Nice, so provided that a given client is configured on the external ceph cluster, we can use this kind of approach to render the right file in the right location.

fmount (approved):
LGTM

PR #33, opened 2022-09-09
Rework CephBackend and rely to the lib-common

fultonj:
Change glance-operator's CephBackend. Do not define our own versions of GetCephOsdCaps, GetCephRbdUser, GetCephGlancePool, and validateMons and their
GlanceAPICephBackend and CephPoolSpec structs.
Instead import them from lib-common [1].

[1] https://github.com/openstack-k8s-operators/lib-common/pull/51

Signed-off-by: John Fulton <fulton@redhat.com>


fultonj:
@fmount should we include a change to the CRDs in both the glance and cinder operators with this move? I'd rather the following names were consistent.
```
CephClusterFSID to ClusterFSID
CephClientKey to ClientKey
CephClusterMonHosts to ClusterMons
CephPoolSpec to PoolSpec
CephPoolName to PoolName
```

controllers/glanceapi_controller.go
```diff
@@ -665,15 +666,15 @@ func (r *GlanceAPIReconciler) generateServiceConfigMaps(
 	using the Client Key provisioned on the Ceph cluster
 	**/
 	if gb == "rbd" {
-		templateParameters["CephClusterFSID"] = instance.Spec.CephBackend.CephClusterFSID
-		templateParameters["CephClusterMonHosts"] = instance.Spec.CephBackend.CephClusterMonHosts
-		templateParameters["CephClientKey"] = instance.Spec.CephBackend.CephClientKey
+		templateParameters["CephClusterFSID"] = instance.Spec.CephBackend.ClusterFSID
```
fmount, line 669: By changing do you mean make those names consistent? (s/CephClusterFSID/ClusterFSID  and so forth)
In theory yes, we can do that, and we have occurrences in [1] we could change.

[1] https://github.com/openstack-k8s-operators/glance-operator/tree/master/templates/glanceapi/config

fmount:
> @fmount should we include a change to the CRDs in both the glance and cinder operators with this move? I'd rather the following names were consistent.
> 
> ```
> CephClusterFSID to ClusterFSID
> CephClientKey to ClientKey
> CephClusterMonHosts to ClusterMons
> CephPoolSpec to PoolSpec
> CephPoolName to PoolName
> ```

Yes, best to be consistent and follow a single naming convention. I'm wondering if your idea is to change those occurrences in the glanceapi config as per my previous comment.

controllers/glanceapi_controller.go
fultonj: Yes, that's the idea. I think having this additional mapping is unnecessary.

fmount: +1, something like [1] should be enough.

[1] https://github.com/openstack-k8s-operators/cinder-operator/commit/d9209644d54ff35bf002a87007db28bc265d41f4

controllers/glanceapi_controller.go
```diff
@@ -665,15 +666,15 @@ func (r *GlanceAPIReconciler) generateServiceConfigMaps(
...
+		templateParameters["CephClusterMonHosts"] = instance.Spec.CephBackend.ClusterMonHosts
+		templateParameters["CephClientKey"] = instance.Spec.CephBackend.ClientKey
 		// The pool we write in glance-api.conf
-		templateParameters["CephPool"] = glance.GetCephGlancePool(instance)
+		templateParameters["CephPool"], _ = ceph.GetPool(instance.Spec.CephBackend.Pools, instance.Name)
```
abays, line 673: Might be best to use `"glance"` instead of `instance.Name` here, because the user might name the CR something other than "glance".  Similar to what we're doing for the Cinder operator here: https://github.com/openstack-k8s-operators/cinder-operator/blob/00a7370b3c7453406ab24c9e965cce26819a7769/controllers/cinder_controller.go#L653

fmount: Yeah, as per our previous conversation on the cinder-operator patch (which is basically similar to this one), if the user named the instance with a diff name, the call to `ceph.GetPool` would return an error. 
Considering that we plan to have a single default related to a service type, I guess for now we can just change the ceph.GetPool to: `ceph.GetPool(instance.Spec.CephBackend.Pools, "glance")`, until new defaults are available and provided by lib-common `storage/ceph` module.

fultonj: OK, I've switched `instance.Name` to `"glance"` as suggested.

abays:
@fultonj It might just be my environment, but I can't seem to run the operator with the patch due to dependency issues.  Does `OPERATOR_TEMPLATES=./templates make run` work for you?

fultonj:
> 

@abays I tested it with `OPERATOR_TEMPLATES=$PWD/templates ./bin/manager` (after building) and used it to create an image on a Ceph server. I have to rebase for the new `go.mod` files anyway. I'll make sure it works with `OPERATOR_TEMPLATES=./templates make run` when I re-submit.

fmount (approved):
+1 on my side, @abays to validate the go.mod change and merge!

abays (approved):
/lgtm

PR #34, opened 2022-09-12
Check if the service is not found during reconcileDelete

fmount:
This patch adds an error check to the glance-operator during reconcileDelete: if it's an error and it's not found we can assume the service has been deleted, otherwise we must fail. The same logic has been added to the cinder operator, and now we have a well established pattern to explicitly check and clean the related resources.

### DIFF
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -207,12 +207,16 @@ func (r *GlanceAPIReconciler) reconcileDelete(ctx context.Context, instance *gla
 
 	// It's possible to get here before the endpoints have been set in the status, so check for this
 	if instance.Status.APIEndpoints != nil {
-		ks, _ := keystone.GetKeystoneServiceWithName(ctx, helper, glance.ServiceName, instance.Namespace)
-		//Ensure the service exists
-		ksSvc := keystone.NewKeystoneService(ks.Spec, instance.Namespace, map[string]string{}, 10)
-		err := ksSvc.Delete(ctx, helper)
-		if err != nil {
+		ks, err := keystone.GetKeystoneServiceWithName(ctx, helper, glance.ServiceName, instance.Namespace)
+		if err != nil && !k8s_errors.IsNotFound(err) {
 			return ctrl.Result{}, err
+		}
+		if !k8s_errors.IsNotFound(err) {
+			ksSvc := keystone.NewKeystoneService(ks.Spec, instance.Namespace, map[string]string{}, 10)
+			err = ksSvc.Delete(ctx, helper)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
 		}
 	}
 	// Service is deleted so remove the finalizer.


abays (approved):
/lgtm

fmount:
@abays I'm wondering if (since we're fixing this behavior), we'd like to reflect the same change on the other operators.
Here what I found from a quick look:

- https://github.com/openstack-k8s-operators/neutron-operator/blob/master/controllers/neutronapi_controller.go#L195
- https://github.com/openstack-k8s-operators/placement-operator/blob/master/controllers/placementapi_controller.go#L205
- https://github.com/openstack-k8s-operators/ironic-operator/blob/master/controllers/ironicapi_controller.go#L249

abays:
> @abays I'm wondering if (since we're fixing this behavior), we'd like to reflect the same change on the other operators. Here what I found from a quick look:
> 
>     * https://github.com/openstack-k8s-operators/neutron-operator/blob/master/controllers/neutronapi_controller.go#L195
> 
>     * https://github.com/openstack-k8s-operators/placement-operator/blob/master/controllers/placementapi_controller.go#L205
> 
>     * https://github.com/openstack-k8s-operators/ironic-operator/blob/master/controllers/ironicapi_controller.go#L249

I made a similar remark elsewhere.  It seems to me that we should indeed be consistent with this pattern across all the operators.

controllers/glanceapi_controller.go
```diff
@@ -207,10 +207,16 @@ func (r *GlanceAPIReconciler) reconcileDelete(ctx context.Context, instance *gla
...
 	// It's possible to get here before the endpoints have been set in the status, so check for this
 	if instance.Status.APIEndpoints != nil {
-		ks, _ := keystone.GetKeystoneServiceWithName(ctx, helper, glance.ServiceName, instance.Namespace)
+		ks, err := keystone.GetKeystoneServiceWithName(ctx, helper, glance.ServiceName, instance.Namespace)
+		if err != nil {
+			if k8s_errors.IsNotFound(err) {
+				return ctrl.Result{}, nil
+			}
+			return ctrl.Result{}, err
+		}
```
stuggi, line 216: I just returned from a longer PTO break so maybe I miss something as my brain us slowly coming into work mode :) 

I am not sure I understand why we need this here when we do the same check in ksSvc.Delete() [1]. and return nil if it does not exist. Isn't it correct in that case to proceed and with the remaining delete tasks.

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/master/pkg/external/funcs.go#L211

stuggi: Now I got it, thanks @abays , to not have to use the service definition for ks.
I think there is still an issue. If we return `return ctrl.Result{}, nil` in the k8s_errors.IsNotFound case we won't proceed and remove the finalizer later.

fmount: @stuggi just a quick question: is `k8s_errors.IsNotFound(err)` assumes that the service is not found? (hence we can return earlier?)

stuggi: yes `k8s_errors.IsNotFound(err)` here assumes that the glanceapi keystoneservice CR is not there, so you want to skip the `ksSvc.Delete()` but still proceed the remaining parts of the func for cleanup. Until there are further cleanup steps the removal of the finalizer on the CR `controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())` .

abays: @fmount My bad for not spotting this in the similar PR for Cinder.  I have a follow-up PR open there to address the concern Martin is sharing

fmount: @abays np, let's see if this version is more aligned with your fix on the cinder side, then I guess we can move on.

abays (approved):
/lgtm

PR #37, opened 2022-09-13
Use latest keystone-operator API

fultonj:
The external pkg moved to an api sub module [1]. The glance operator imported both and named them keystonev1beta1 and keystone. Remove keystonev1beta1 references. Redefine keystone import so that it uses the latest keystone-operator API like keystonev1beta1 did.

[1] https://github.com/openstack-k8s-operators/keystone-operator/pull/90


fultonj:
I did initial testing but want to do further testing before we merge. Early input welcome now however.

fmount:
@fultonj the patch looks good, I have a few more things removed in go.sum but maybe it's my local environment which is a bit messed up [1]. Feel free to check.

[1] https://paste.opendev.org/show/bQpewRHlMRdvUUivZz73/

controllers/glanceapi_controller.go
```diff
@@ -370,7 +369,7 @@ func (r *GlanceAPIReconciler) reconcileInit(
 		return ctrl.Result{}, err
 	}
 
-	ksSvcSpec := keystonev1beta1.KeystoneServiceSpec{
+	ksSvcSpec := keystone.KeystoneServiceSpec{
```
fmount, line 372: +1 makes sense

fmount (approved):

fultonj:
I've tested further and it's still working.

fultonj:
I'm going to change `keystone` to `keystonev1` so that we version the API.

dprince (approved):

PR #46, opened 2022-09-27
Copy rendered CRD files dir to /api/bases in make manifests

stuggi:
With the new api sub module, consumer operators just have to import the api submodule. For envtest it is required to install CRD for e.g. keystone-operator and mariadb-operator to be able to use it.

Moving the directory config/crd/bases which holds the generated files in to the api module that they are available with the module and link the dir in the original place does not work as kustomize autodetect fails on symlinked directories [1].

This updates the `generate` Makefile target to copy config/crd/bases to /api on successful generation run.

[1] kubernetes-sigs/kustomize#1886


Makefile
```diff
@@ -88,7 +88,8 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases && \
+	rm -f api/bases/* && cp -a config/crd/bases api/
```
fmount, line 92: I understand we need those files in the api submodule, and the symlink generates issues.
I'm wondering (long term) if we'd like to use a different strategy, like rely on a `lib-common/api` module where we collect crds (just an idea).
However this lgtm for now

fmount (approved):

Makefile
fmount: > > I understand we need those files in the api submodule, and the symlink generates issues. I'm wondering (long term) if we'd like to use a different strategy, like rely on a `lib-common/api` module where we collect crds (just an idea). However this lgtm for now
> 
> we have discussed a couple of different approaches, this one is the quick/easy fix. CRDs are versioned, so the tricky thing is to have a solution which fits it. right now we can check the api module version from go.mod and get the CRD files for that exact version. And it also works for custom repos/forks.
> 
> We discussed to move them in a central location, but the versioning introduces a real complexity. All CRDs of the different services should likely need some same versioning. but if there is just a change/fix to one api it requires a new version for all of them.... Some projects have this central api organization. we might want to check with them how they do it.

Thanks for clarifying this aspect. 
I think in the current stage this is definitely the best approach, and give us time to investigate how other projects addressed this problem.

abays (approved):
/lgtm

PR #48, opened 2022-09-29
Update makefile

Akrog:
These patches change our Makefile to make our life easier as developers:

- Removes the need to have the `kubectl` command, using `oc` instead.
- Add support to pushing to insecure registries
- `make deploy` will work as it is after using the meta-operator.
- `install/uninstall` targets will not only take care of CRDs but also RBACs.

This is the equivalent changes to the one done in Cinder in [PR#37](https://github.com/openstack-k8s-operators/cinder-operator/pull/37) and [PR#50](https://github.com/openstack-k8s-operators/cinder-operator/pull/50)


dprince (approved):

abays (approved):
/lgtm

PR #50, opened 2022-10-03
Defaults for top-level CRD's API specs

abays:
This small change will allow for the optional exclusion of the `GlanceAPIInternal` and `GlanceAPIExternal` fields in the top-level `Glance` CRD.  It simplifies the required `Glance` CR to deploy the basic service if the user is not interested in customizing the individual internal and external APIs.

### DIFF
--- a/api/bases/glance.openstack.org_glances.yaml
+++ b/api/bases/glance.openstack.org_glances.yaml
@@ -120,6 +120,8 @@ spec:
                   in /etc/<service> . TODO: -> implement'
                 type: object
               glanceAPIExternal:
+                default:
+                  replicas: 1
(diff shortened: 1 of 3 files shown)


fultonj:
Seems like a good idea to me. Thanks.

api/bases/glance.openstack.org_glances.yaml
```diff
@@ -120,6 +120,8 @@ spec:
                   in /etc/<service> . TODO: -> implement'
                 type: object
               glanceAPIExternal:
+                default:
+                  replicas: 1
```
fmount, line 124: Thanks, this looks good to me, and it's actually a good idea providing this level of flexibility.

fmount (approved):

PR #51, opened 2022-10-04
Use optional secret (like other operators)

abays:
Until we decide how we truly wish to handle either requiring or not requiring `Secret` in our individual operator CRDs, this simply sets that field to be optional as it currently is in all other operators.  This will ease its integration with the meta operator.

### DIFF
--- a/api/bases/glance.openstack.org_glances.yaml
+++ b/api/bases/glance.openstack.org_glances.yaml
@@ -463,7 +463,6 @@ spec:
             required:
             - glanceAPIExternal
             - glanceAPIInternal
-            - secret
             - storageRequest
             type: object
           status:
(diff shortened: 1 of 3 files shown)


api/bases/glance.openstack.org_glances.yaml
```diff
@@ -463,7 +463,6 @@ spec:
             required:
             - glanceAPIExternal
             - glanceAPIInternal
-            - secret
```
fmount, line 466: lgtm, if this simplify your work on the meta-operator, it's fine by me.
FYI the next change which is coming for both cinder and glance (later on manila operator as well), is related to the introduction of a new API extension that allows to mount `extraVolumes`.
The CRD will include  something like the following:

```
 extraMounts:
    - propagation: All
      extraVolType: Ceph
      volumes:
      - name: ceph
        secret:
          secretName: ceph-client-conf
      mounts:
      - name: ceph
        mountPath: "/etc/ceph"
        readOnly: true
```

but this is unrelated to this change, so I'm go to go on this.

fmount (approved):

fultonj (commented):
LGTM

PR #52, opened 2022-10-05
Mutating webhook for enforcing APIType

abays:
When a `Glance` CR is created, we want to force the embedded `GlanceAPI` specs for the internal and external API to properly designate "internal" and "external", respectively, in their `APIType` fields (regardless of whatever the user may have placed there).  A mutate webhook allows us to do just this.

This PR will also serve as an example to consult for adding webhooks to other operators.

NOTE: after these changes, to run this operator locally (via `make run`), the `ENABLE_WEBHOOKS=false` must be added as an environment variable.  This is a common pattern across operators that use webhooks.


abays:
This is outdated and superseded by other approaches

PR #57, opened 2022-10-11
Makefile: Support http & insecure registry in opm

Akrog:
In a previous patch we added the USE_TLS environmental variable to our Makefile to allow pushing to insecure local registries during development.

This patch extends this functionality to support the catalog building as well, which is convenient when using the `registry` container image for local development.

With podman/docker push using the VERIFY_TLS=false will not only disable the TLS verification but will also allow HTTP connections, which is not the case of the `opm` command.

The `opm` command has 2 different options `--skip-tls-verify` and `--use-http`, so we'll use the existing `VERIFY_TLS` for the first and a new `USE_HTTP` for the second.


abays (commented):
+1 from me

PR #68, opened 2022-10-20
Bump lib-common/common to get endpoint port bug fix

fultonj:
lib-common patch 5a061c1d139c4a0110fdfa8c8ea99dae73719ac4 is necessary for the openstack CLI to not fail to connect to glance because it tries to reach the route on a port hardcoded in the URL.

Bump the version of lib-common/modules/common so that the glance operator creates endpoints without the port appended so that glance end points are created correctly.

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -3,7 +3,7 @@ module github.com/openstack-k8s-operators/glance-operator/api
 go 1.18
 
 require (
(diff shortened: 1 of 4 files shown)


fmount (approved):
This pull down the new `lib-common/modules/common` version with the endpoint port fixes, thanks John!

abays (approved):
/lgtm

PR #69, opened 2022-10-24
Use defer func to update metadata/finalizer

abays:
This is prototype example of how we might handle delaying the update of adding/removing finalizers until the very end of each reconcile cycle.  It is intended to avoid requiring both an explicit `Update` call and an immediate reconcile return for adding/removing finalizers during the course of a reconcile loop.

### DIFF
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -162,6 +162,21 @@ func (r *GlanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			util.LogErrorForObject(helper, err, "Set after and calc patch/diff", instance)
 		}
 
+		if changed := helper.GetChanges()["metadata"]; changed {
+			// Patch wipes out all defaulted non-scalar fields from the spec (such as PasswordSelector)
+			// that were not explicitly included in the instance's CR YAML
+			// if err := r.Patch(ctx, instance, patch); err != nil && !k8s_errors.IsNotFound(err) {
+			// 	util.LogErrorForObject(helper, err, "Update finalizer", instance)
+			// }
+
+			// Update does not exhibit the "wipes out defaulted non-scalar fields" behavior mentioned above
+			if err := r.Update(ctx, instance); err != nil && !k8s_errors.IsNotFound(err) {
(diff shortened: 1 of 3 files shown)


controllers/glance_controller.go
```diff
@@ -162,6 +162,21 @@ func (r *GlanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			// 	util.LogErrorForObject(helper, err, "Update finalizer", instance)
+			// }
+
+			// Update does not exhibit the "wipes out defaulted non-scalar fields" behavior mentioned above
+			if err := r.Update(ctx, instance); err != nil && !k8s_errors.IsNotFound(err) {
+				util.LogErrorForObject(helper, err, "Update metadata", instance)
+			}
+		}
+
+		// Calling "r.Status().Patch()" BEFORE handling the metadata causes any metadata changes to be
```
gibizer, line 178: This is now AFTER the metadata change

controllers/glance_controller.go
```diff
@@ -162,6 +162,21 @@ func (r *GlanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			util.LogErrorForObject(helper, err, "Set after and calc patch/diff", instance)
 		}
 
+		if changed := helper.GetChanges()["metadata"]; changed {
+			// Patch wipes out all defaulted non-scalar fields from the spec (such as PasswordSelector)
+			// that were not explicitly included in the instance's CR YAML
+			// if err := r.Patch(ctx, instance, patch); err != nil && !k8s_errors.IsNotFound(err) {
```
gibizer, line 168: I vaguely remember similar issues. At that time the problem was that the previous r.Update call updated our local instance object with the result of the Update. At that time I could circumvent the issue by passing a DeepCopy of the instance there. We have  r.Update calls before this place at L132. So I'm wondering if that is related.

controllers/glance_controller.go
abays: Right, that's why it's after instead of before (and I just wanted to explain why I put it after instead of before).  Let me clarify the comment.

abays:
Closing this, as we have a new proposed pattern for this concern

PR #71, opened 2022-10-31
Migrate from policy.json to policy.yaml

konan-abhi:
OpenStack is now using yaml files for policy enforcement rather than json files. This patch now makes sure that we are using yaml file and not json.


abays (approved):
/lgtm

PR #72, opened 2022-10-31
Add default policy rules to yaml file

konan-abhi:
Even though glance support policy in code for policy enforcement, the default policy in glance is open for all. Which means unless secure-rbac is enabled we can not restrict glance APIs to any particular user/role/project. This restricts us from testing policy based validations of each API.

Adding policy.yaml with policy rule for each API to enable this policy enforcement without enabling actual secure-rbac enforcement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,74 @@ spec:
 When the service is up and running, it's possible to interact with the glance
 API and upload an image using the Ceph backend.
 
+## Example: Testing Glance policies
+
+Glance's public API calls may be restricted to certain sets of users using a
+policy configuration file. This section explains exactly how to test APIs with
+policies. Glance operator configures default policies in policy.yaml file which
+will be stored at '/etc/glance' section. The following represents how you can
+test API(s) with glance policies.
+
+```
+Create below projects, users and assign member and reader roles to each of them;
+
+openstack project create --description 'project a' project-a --domain default
+openstack project create --description 'project b' project-b --domain default
+
+openstack user create project-a-reader --password project-a-reader
+openstack user create project-b-reader --password project-b-reader
+openstack user create project-a-member --password project-a-member
+openstack user create project-b-member --password project-b-member
+
+openstack role add --user project-a-member --project project-a member
+openstack role add --user project-a-reader --project project-a reader
+openstack role add --user project-b-member --project project-b member
+openstack role add --user project-b-reader --project project-b reader
+
+Create project-a-reader-rc, project-a-member-rc, project-b-reader-rc, project-b-member-rc using below contents;
+
+export OS_AUTH_URL=<auth-url>
+export OS_PASSWORD='project-a-member'
+export OS_PROJECT_DOMAIN_NAME=Default
+export OS_PROJECT_NAME='project-a'
+export OS_USER_DOMAIN_NAME=Default
+export OS_USERNAME='project-a-member'
+export OS_CACERT=/etc/pki/ca-trust/source/anchors/simpleca.crt
+export OS_IDENTITY_API_VERSION=3
+export OS_REGION_NAME=regionOne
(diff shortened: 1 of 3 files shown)


fmount:
@konan-abhi thanks for this patch, this is definitely aligned with the current glance deployment. To get `policy.yaml` applied to the right path (`/etc/glance` as per glance-api.conf), you have to improve this PR and add something like the following:

```
diff --git a/templates/glance/config/glance-api-config.json b/templates/glance/config/glance-api-config.json
index ef38150..6b682a3 100644
--- a/templates/glance/config/glance-api-config.json
+++ b/templates/glance/config/glance-api-config.json
@@ -30,6 +30,12 @@
         "dest": "/etc/ceph/ceph.client.{{ .User }}.keyring",
         "owner": "glance",
         "perm": "0600"
+      },
+      {
+        "source": "/var/lib/config-data/merged/policy.yaml",
+        "dest": "/etc/glance/policy.yaml",
+        "owner": "glance",
+        "perm": "0600"
       }
     ]
 }
 ```
 By doing this, when the pod starts, it copies the file to the right destination, and then it's properly applied.
 Can you try that (and test) before moving forward?

**Review by fmount — CHANGES_REQUESTED** [review]
We need to copy the `policy.yaml` file to the expected destination

**Comment by konan-abhi** [comment]
> 

Ack, will make these changes and check whether policies are applied to API calls and then update the patch.
Thank you for explanation :+1:

**Comment by konan-abhi** [comment]
Simple way to change policies are working is, create 2 more projects, create 2 users for each project (member and reader) and assign member and reader roles to these projects and users; For example as shown in https://paste.opendev.org/show/bpUROsl5E4dhYMo3KQWJ/

File: templates/glance/config/glance-api-config.json [hunk 1]
```diff
@@ -7,6 +7,12 @@
         "owner": "glance",
         "perm": "0600"
       },
+      {
+        "source": "/var/lib/config-data/merged/policy.yaml",
+        "dest": "/etc/glance/glance.conf.d/policy.yaml",
```
fmount, line 12: Thank @konan-abhi, now I can see the right policy file being used!
I'm just wondering if a change in the README can be added to this PR so we can make sure we document how to build projects and interact w/ glance

fmount:
> Simple way to change policies are working is, create 2 more projects, create 2 users for each project (member and reader) and assign member and reader roles to these projects and users; For example as shown in https://paste.opendev.org/show/bpUROsl5E4dhYMo3KQWJ/

Maybe the pastebin can be turned into a section in the operator `README` so that people know the steps required to interact with Glance (even for CI purposes it would be useful)

konan-abhi:
Once https://github.com/openstack-k8s-operators/glance-operator/pull/83 is merged I don't think this is required.
Earlier policies were not working as the request were not piped to keystone for authentication.

Will test this once again once PR 83 is merged and if required the will reopen it.

konan-abhi:
Nope,this is still required.

konan-abhi:
@fultonj Please review, thank you!

README.md
```diff
@@ -84,6 +84,74 @@ spec:
 When the service is up and running, it's possible to interact with the glance
 API and upload an image using the Ceph backend.
 
+## Example: Testing Glance policies
```
fmount, line 87: cool, nice we have an example to keep testing the functions of the deployed `Glance` service.
From the previous conversations, I see the agreement is not to use `install_yamls` project to build project/users that can be used by the Galnce operator, but from a dev perspective, I think it's useful having a way to deploy the operator, see the service up & running, and try to list or even upload an image.
We can maybe provide a dev/ directory into hack as a place where we can store scripts (e.g. bash scripts) to help w/ development, but definitely something we can follow up later.
Thanks for this patch Abhishek!

fmount (approved):

PR #74, opened 2022-11-04
Fix defaulting PasswordSelectors

stuggi:
When a CRD has an optional struct field the defaulting of that field needs special care. Both the field with the struct type need a full default value defined and each individual subfields in the struct needs default value defined. Otherwise in the first reconcile call the PasswordSelectors is empty and then in the second reconcile it is filled.

Now documented via https://github.com/openstack-k8s-operators/docs/pull/12

### DIFF
--- a/api/v1beta1/glanceapi_types.go
+++ b/api/v1beta1/glanceapi_types.go
@@ -69,7 +69,8 @@ type GlanceAPISpec struct {
 	Secret string `json:"secret"`
 
 	// +kubebuilder:validation:Optional
-	// PasswordSelectors - Selectors to identify the AdminUser password from the Secret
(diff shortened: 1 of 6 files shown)


api/v1beta1/glanceapi_types.go
```diff
@@ -69,7 +69,8 @@ type GlanceAPISpec struct {
 	Secret string `json:"secret"`
 
 	// +kubebuilder:validation:Optional
```
stuggi, line 71: do we support setup glanceapi without glance ? if not, we could also set this to be required on the glanceapi and set via the glance controller when create the glanceapi?

abays (approved):
/lgtm

PR #75, opened 2022-11-04
Introduce GlanceAPI Extra Volumes support

fmount:
The `Glance` operator can now get rid of the `CephBackend` specific implementation and rely on the general purpose approach provided by `lib-common/modules/storage`.
`ExtraVolumes` are now defined as part of the glance top level crd and can be propagated to the `GlanceAPI` pods according to the specified policy.

Depends-on: https://github.com/openstack-k8s-operators/lib-common/pull/88


fmount:
@abays Not sure here I can really add a dependency on lib-common [1] that can be included in the CI pipeline: I suspect we should keep this change as it is until [1] is available.

[1] https://github.com/openstack-k8s-operators/lib-common/pull/88

templates/glance/config/glance-api-config.json
```diff
@@ -24,18 +24,6 @@
         "dest": "/etc/glance/logging.conf",
         "owner": "glance",
         "perm": "0600"
-      },
```
fmount, line 27: @konan-abhi I see a problem to keep the old ceph mechanism coexisting with `extraVoumes`: in particular if I propagate an extra volume in `/etc/ceph` via something like the following:

```
 extraMounts:
      extraVol:
        - propagation:
          - Glance
          extraVolType: Ceph
          volumes:
          - name: ceph
            projected:
              sources:
              - secret:
                  name: ceph-client-conf
          mounts:
          - name: ceph
            mountPath: "/etc/ceph"
            readOnly: false
```

It won't take effect because when the pod starts `kolla` overwrites/replaces the content of the mounted path (even `merge: true` doesn't solve the problem) [1].

As I have done with cinder-operator,  I'm thinking about just removing the `old` ceph way, and rely entirely on this feature once it merges.

FYI see the POC that proves how it works [2]

[1] https://docs.openstack.org/kolla/ussuri/admin/kolla_api.html
[2] https://asciinema.org/a/533951

api/v1beta1/glance_types.go
```diff
@@ -186,3 +186,27 @@ func (instance Glance) IsReady() bool {
 	// - Both the internal and external API endpoints are ready (count >= 1)
 	return instance.Status.GlanceAPIInternalReadyCount >= 1 && instance.Status.GlanceAPIExternalReadyCount >= 1
 }
+
+// GlanceExtraVolMounts exposes additional parameters processed by the glance-operator
+// and defines the common VolMounts structure provided by the main storage module
+type GlanceExtraVolMounts struct {
```
fmount, line 192: @stuggi This part that was previously in `lib-common` is now versioned as part of the glance `v1beta1` API. Is this good enough?

templates/glance/config/glance-api-config.json
konan-abhi: Looks good to me, just wondering how or from where rbd configuration and store details for enabled_backends are popped in glance-api.conf (because I need to enable image_conversion plugin if ceph is deployed as glance backend)

fmount: @konan-abhi this is a good point, and it's where (the so far unused) `extraVolType: Ceph` comes into the picture.

We usually mount `ExtraVolumes` according to the data structure mentioned in my previous comment, and in `api/v1beta1/glance_types.go` we have an `extension` of what `lib-common` provides with extra fields (e.g.,  name, region, etc).
Technically, we don't have to tweak that data structure, and we know that `if extraVolType is Ceph`, we have `/etc/ceph` populated with the files required to configure `Glance` with `Ceph` (this field is supposed to help within the operator logic).
However, I'm not sure this is enough at the operator level: we agreed (in the operators ecosystem) that we would like to not have hardcoded or parametrized configuration, and the meta-operator should be able to inject the configuration via the created CR, ending up with something like this:

```
kind: Glance
metadata:
  name: glance
spec:
  serviceUser: glance
  containerImage: quay.io/tripleozedcentos9/openstack-glance-api:current-tripleo
  customServiceConfig: |
    [DEFAULT]
    enabled_backends=default_backend:rbd,ceph1:rbd
    [glance_store]
    [default_backend]
    rbd_store_ceph_conf=/etc/ceph/ceph.conf
    rbd_store_user=<user>
    rbd_store_pool=<pool>
    store_description=Ceph glance store backend.
    [ceph1]
    rbd_store_ceph_conf=/etc/ceph2/ceph.conf
    rbd_store_user=<user>
    rbd_store_pool=<pool>
    store_description=Ceph glance store backend.
...
...
...
```
So in theory we should know for sure that Ceph is configured and used by Glance if two conditions are met:

1. `ExtraVolType` is Ceph
2. `deployment.Spec.CustomServiceConfig` contains `rbd` in the `enabled_backend` section.

As a follow up of this change, we can talk more and make a PR that implements the described behavior

fmount:
> @abays Not sure here I can really add a dependency on lib-common [1] that can be included in the CI pipeline: I suspect we should keep this change as it is until [1] is available.
> 
> [1] [openstack-k8s-operators/lib-common#88](https://github.com/openstack-k8s-operators/lib-common/pull/88)

This is now solved: `lib-common` PR merged and we updated it in `go.mod`

templates/glance/config/glance-api-config.json
konan-abhi: Thank you, make sense.

api/v1beta1/glance_types.go
```diff
@@ -69,10 +69,6 @@ type GlanceSpec struct {
 	// actual action pod gets started with sleep infinity
 	Debug GlanceDebug `json:"debug,omitempty"`
 
-	// +kubebuilder:validation:Optional
-	// CephBackend - The ceph Backend structure with all the parameters
-	CephBackend ceph.Backend `json:"cephBackend,omitempty"`
```
fultonj, line 74: Yes, let's remove this instead of trying to make them co-exist. We'll switch to the new one.

fultonj:
Patch looks fine to me. I tested this in my environment and it works.

fultonj (approved):

PR #83, opened 2022-11-15
Enable keystone authentication for glance

konan-abhi:
As of now if you deploy glance-operator doesn't enable auth middleware causing glance to use no-auth, which means glance is not validating any API calls via keystone. This is also the reason why owner field on image is set to None if you create glance image. In order to enable it you need to set 'flavor' as 'keystone' under 'paste_deploy' section of glance-api.conf. Setting this leads to another issue that under 'keystone_authtoken' section 'services' project name is used which is not a valid keystone project.

This commit fixes all the issues related to keystone authentication for glance operator.


fmount (approved):
Thanks Abhishek, looks good!

PR #84, opened 2022-11-15
Fix CRD inconsistency for glanceAPIInternal and External

gibizer:
Both glanceAPIInternal and glanceAPIExternal fields was marked as Required but had an incomplete default value defined as well. This caused the CRD validation to fail when glance-operator is deployed.

```
  - lastTransitionTime: "2022-11-15T13:02:29Z" message: 'CustomResourceDefinition.apiextensions.k8s.io "glances.glance.openstack.org" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[glanceAPIExternal].default.containerImage: Required value, spec.validation.openAPIV3Schema.properties[spec].properties[glanceAPIInternal].default.containerImage: Required value]' reason: InstallComponentFailed status: "True" type: InstallPlanFailed
```

So this patch drops the incomplete default and updates the sample to have the image specified for both.

### DIFF
--- a/api/bases/glance.openstack.org_glances.yaml
+++ b/api/bases/glance.openstack.org_glances.yaml
@@ -120,8 +120,6 @@ spec:
                   in /etc/<service> . TODO: -> implement'
                 type: object
               glanceAPIExternal:
-                default:
-                  replicas: 1
                 description: GlanceAPIExternal - Spec definition for the external
                   API service of this Glance deployment
                 properties:
@@ -280,8 +278,6 @@ spec:
                 - containerImage
                 type: object
               glanceAPIInternal:
-                default:
-                  replicas: 1
                 description: GlanceAPIInternal - Spec definition for the internal
                   and admin API service of this Glance deployment
                 properties:
--- a/api/v1beta1/glance_types.go
+++ b/api/v1beta1/glance_types.go
@@ -99,12 +99,10 @@ type GlanceSpec struct {
 	StorageRequest string `json:"storageRequest"`
 
 	// +kubebuilder:validation:Required
-	// +kubebuilder:default:={replicas: 1}
 	// GlanceAPIInternal - Spec definition for the internal and admin API service of this Glance deployment
 	GlanceAPIInternal GlanceAPISpec `json:"glanceAPIInternal"`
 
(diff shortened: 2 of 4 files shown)


api/bases/glance.openstack.org_glances.yaml
```diff
@@ -120,8 +120,6 @@ spec:
                   in /etc/<service> . TODO: -> implement'
                 type: object
               glanceAPIExternal:
-                default:
-                  replicas: 1
```
Akrog, line 124: If I understand correctly what you are saying we cannot set the default replicas to 1 unless we set a default for all the other required fields (eg. containerImage), right?

api/v1beta1/glance_types.go
```diff
@@ -99,12 +99,10 @@ type GlanceSpec struct {
 	StorageRequest string `json:"storageRequest"`
 
 	// +kubebuilder:validation:Required
-	// +kubebuilder:default:={replicas: 1}
 	// GlanceAPIInternal - Spec definition for the internal and admin API service of this Glance deployment
```
gibizer, line 102: This was basically a contradiction. You cannot say GlanceAPIInternal is a required field and also at the same time provide some default values for it. As far as I understand this combination of kubebuilder markers made the GlanceAPIInternal field still optional with the default (i.e. Required was ignored).

At the same time GlanceAPISpec.ContainerImage is marked as Required[1]. Which means creating a GlanceAPISpec with the value `{replicas: 1}` is not possible as `containerImage` is not specified in that value but the field is Required. This caused the actual failure.

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/e8a8dd567b252af6e89b6d9bd000c36db846c76c/api/v1beta1/glanceapi_types.go#L41-L43

api/bases/glance.openstack.org_glances.yaml
gibizer: The default value of a struct needs to fill every fields of that struct that are marked as Required otherwise the default value cannot be used to construct a struct instance. See my other comment in glance_types.go below

Akrog (approved):
/lgtm

dprince (approved):

PR #90, opened 2022-11-23
Update internal/external service configuration

ASBishop:
The show_multiple_locations setting, as well as
show_image_direct_url, needs to be configured for the internal and external services to work correctly. Both settings are True for the internal service in order for it to provide image location data. Conversely, both are set False for the external service, so that image location data is excluded from public view.

### DIFF
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -542,11 +542,14 @@ func (r *GlanceAPIReconciler) generateServiceConfigMaps(
 		templateParameters["OsdCaps"] = ceph.GetOsdCaps(instance.Spec.CephBackend.Pools)
 	}
 
-	// We set "show_image_direct_url" to true if this is an internal GlanceAPI,
-	// otherwise we default to false
-	templateParameters["ShowImageDirectUrl"] = false
+	// Configure the internal GlanceAPI to provide image location data, and the
+	// external version to *not* provide it.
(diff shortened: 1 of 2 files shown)


fultonj (approved):
This LGTM. I think we should merge it. Awaiting input from others.

konan-abhi (commented):
Looks good to me!

controllers/glanceapi_controller.go
```diff
@@ -542,11 +542,14 @@ func (r *GlanceAPIReconciler) generateServiceConfigMaps(
...
 	}
 
-	// We set "show_image_direct_url" to true if this is an internal GlanceAPI,
-	// otherwise we default to false
-	templateParameters["ShowImageDirectUrl"] = false
+	// Configure the internal GlanceAPI to provide image location data, and the
+	// external version to *not* provide it.
 	if instance.Spec.APIType == glancev1.APIInternal {
 		templateParameters["ShowImageDirectUrl"] = true
+		templateParameters["ShowMultipleLocations"] = true
```
Akrog, line 549: Since we want both to have the same value we could just name it `"IsInternalService"` or something like that and use a single template variable.

And we could then just simplify L547 to L552 to:

```
templateParameters["IsInteralService"] = instance.Spec.APIType == glancev1.APIInternal
```

Akrog (approved):
/lgtm

PR #91, opened 2022-11-28
use stable/zed container images

SeanMooney:
This change makes the container image optional and defaults
it to use stable/zed.


fmount:
I think this change is ok as long as we merge #93 right after: I made a quick test and it's good to go [1]

[1] https://paste.opendev.org/show/bjAhsg7LjCELHMwyiORo/

fmount (approved):

PR #94, opened 2022-12-01
Override maxDescLen for install and deploy

fmount:
Installing `CRDs` currently fails due to the length of the `CRD`. This patch aligns the glance operator to the cinder PR [1], where the descriptions on the `CRDs` are removed when `install` and `deploy` targets are run.
The patch [2], which introduces `extraVolumes` to `glance-operator` requires this fix, otherwise we'll fail with the following:

"invalid: metadata.annotations: Too long: must have at most 262144 bytes"

[1] https://github.com/openstack-k8s-operators/cinder-operator/pull/81
[2] https://github.com/openstack-k8s-operators/glance-operator/pull/75

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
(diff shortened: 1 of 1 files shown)


stuggi (approved):
/lgtm

PR #95, opened 2022-12-04
add service db finalizer logic

pinikomarov:
add_service_db_finalizer_logic

Ever since  : https://github.com/openstack-k8s-operators/lib-common/pull/87

Podified operators which use service dbs and the db.CreateOrPatchDB method ,
now will get the finalizer set on the db automatically when using the newest lib-common, 
but they won't have the code remove it on cleanup i.e reconcileDelete,
which might make the delete operator action hang.

This adds the service Db removal code in the operator delete (reconcileDelete)
function to work with the above.

Also updated lib-common to latest as to not break lib-common's exported functions

### DIFF
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -197,6 +197,18 @@ func (r *GlanceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 func (r *GlanceReconciler) reconcileDelete(ctx context.Context, instance *glancev1.Glance, helper *helper.Helper) (ctrl.Result, error) {
 	r.Log.Info(fmt.Sprintf("Reconciling Service '%s' delete", instance.Name))
 
+	// remove db finalizer first
+	db, err := database.GetDatabaseByName(ctx, helper, instance.Name)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return ctrl.Result{}, err
+	}
+
+	if !k8s_errors.IsNotFound(err) {
+		if err := db.DeleteFinalizer(ctx, helper); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
 	// Remove the finalizer from our KeystoneService CR
 	keystoneService, err := keystonev1.GetKeystoneServiceWithName(ctx, helper, glance.ServiceName, instance.Namespace)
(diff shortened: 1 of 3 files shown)


pinikomarov:
@fmount Hi can you please review ? 
thanks

controllers/glance_controller.go
```diff
@@ -197,6 +197,18 @@ func (r *GlanceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 func (r *GlanceReconciler) reconcileDelete(ctx context.Context, instance *glancev1.Glance, helper *helper.Helper) (ctrl.Result, error) {
 	r.Log.Info(fmt.Sprintf("Reconciling Service '%s' delete", instance.Name))
 
+	// remove db finalizer first
+	db, err := database.GetDatabaseByName(ctx, helper, instance.Name)
```
fmount, line 201: @pinikomarov I see that this comes from [1], so we're basically removing the finalizer checking the db entry.
LGTM, but I guess this change needs rebase first.

[1] https://github.com/openstack-k8s-operators/lib-common/pull/87

fmount (changes requested):

fmount (approved):

fmount (changes requested):
CI is not happy, I suspect we need to bump lib-common if the exported function isn't available

pinikomarov:
sure I'll take another shot at the rebase

pinikomarov:
updated lib-common/modules/database to a more recent version  in go.mod , now the import thing should be fine

pinikomarov:
uploaded a cleaner version with only new code in the controller and just the updated libs (as in the keystone operator's go.mod)

pinikomarov:
@fmount does this patch need any more permissions for tests from you ? 
all good here ?

stuggi:
have you run `go mod tidy` ? seems the go.sum changes are missing

pinikomarov:
> have you run `go mod tidy` ? seems the go.sum changes are missing

I have and updated the patch with the changes it made

stuggi (approved):
/lgtm

pinikomarov:
cool, thanks

PR #98, opened 2022-12-08
Add missing dependencies

fmount:
`make docker-build` target currently fails because `lib-common/modules/storage` is missing from the `api/` (where the `storage.VolMounts` is defined). This patch updates the mentioned dependencies.


fmount:
@fultonj I have to make the same change in the current cinder PR so we can avoid this issue when it merges!

abays (approved):
/lgtm

PR #110, opened 2022-12-10
Optimize the Propagation struct

fmount:
As done for `cinder operator`, this patch optimize the code to and set both `ServiceTypes` and `Groups` as `constants` in the glance code.
By doing this we can avoid redefining the same variable many times in the code and we can just call what's defined in a single place.


konan-abhi (approved):
Looks good, thank you!

PR #117, opened 2023-01-09
Use new lib-common 'PatchInstance' pattern

abays:
Introduces the use of `lib-common`'s `PatchIntstance()` func and attempts to remove instance-update race condition concerns

### DIFF
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -94,7 +94,7 @@ type GlanceReconciler struct {
 // +kubebuilder:rbac:groups=keystone.openstack.org,resources=keystoneservices,verbs=get;list;watch;create;update;patch;delete;
 
 // Reconcile reconcile Glance requests
-func (r *GlanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *GlanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
 	_ = log.FromContext(ctx)
 
 	// Fetch the Glance instance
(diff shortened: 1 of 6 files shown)


controllers/glance_controller.go
```diff
@@ -111,6 +111,36 @@ func (r *GlanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		err := helper.PatchInstance(ctx, instance)
+		if err != nil {
+			_err = err
+			return
+		}
+	}()
+
+	// If we're not deleting this and the service object doesn't have our finalizer, add it.
+	if instance.DeletionTimestamp.IsZero() && controllerutil.AddFinalizer(instance, helper.GetFinalizer()) {
+		return ctrl.Result{}, err
```
stuggi, line 141: same as `return ctrl.Result{}, nil`, so could also return this explicitly.

controllers/glance_controller.go
```diff
@@ -129,9 +159,7 @@ func (r *GlanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		instance.Status.Conditions.Init(&cl)
 
 		// Register overall status immediately to have an early feedback e.g. in the cli
-		if err := r.Status().Update(ctx, instance); err != nil {
-			return ctrl.Result{}, err
-		}
+		return ctrl.Result{}, err
```
stuggi, line 162: same as `return ctrl.Result{}, nil`, so could also return this explicitly.

controllers/glanceapi_controller.go
```diff
@@ -107,6 +107,36 @@ func (r *GlanceAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		err := helper.PatchInstance(ctx, instance)
+		if err != nil {
+			_err = err
+			return
+		}
+	}()
+
+	// If we're not deleting this and the service object doesn't have our finalizer, add it.
+	if instance.DeletionTimestamp.IsZero() && controllerutil.AddFinalizer(instance, helper.GetFinalizer()) {
+		return ctrl.Result{}, err
```
stuggi, line 137: same as `return ctrl.Result{}, nil`, so could also return this explicitly.

controllers/glanceapi_controller.go
```diff
@@ -125,9 +155,7 @@ func (r *GlanceAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		instance.Status.Conditions.Init(&cl)
 
 		// Register overall status immediately to have an early feedback e.g. in the cli
-		if err := r.Status().Update(ctx, instance); err != nil {
-			return ctrl.Result{}, err
-		}
+		return ctrl.Result{}, err
```
stuggi, line 158: same as `return ctrl.Result{}, nil`, so could also return this explicitly.

stuggi (commented):
looks good to me, could just use `return ctrl.Result{}, nil` as err is nil at the places where we return to update the object.

abays:
> looks good to me, could just use `return ctrl.Result{}, nil` as err is nil at the places where we return to update the object.

Whoops, that's an oversight on my part.  I will change that.

controllers/glance_controller.go
abays: Changed

controllers/glance_controller.go
abays: Changed

controllers/glanceapi_controller.go
abays: Changed

controllers/glanceapi_controller.go
abays: Changed

fmount (approved):

PR #118, opened 2023-01-10
Clean up some removed options

kajinamit:
The base glance-api.conf template file includes some options which were already removed. This removes these options to make the file clean.


konan-abhi (approved):
Looks good to me, thank you!

PR #120, opened 2023-01-13
Use internal endpoint to interact with Keystone

kajinamit:
Internal endpoint is meant for interaction between services. This ensures glance uses internal endpoints when it sends requests to keystone.


konan-abhi (approved):
Looks good, thank you for addressing it!

PR #121, opened 2023-01-16
Set KeystoneAuthURL according to the instance type

fmount:
Previous changes split the `GlanceAPI` between internal and external. This means that we should connect to the `Keystone` endpoint according to the nature of the `GlanceAPI` instance.
This change fixes the way `glance-api.conf` resolves the endpoint, keeping it generic as it's injected by the operator when the pods are created.

### DIFF
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -640,10 +640,8 @@ func (r *GlanceReconciler) apiDeploymentCreateOrUpdate(instance *glancev1.Glance
 	return deployment, op, err
 }
 
-//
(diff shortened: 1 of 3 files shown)


konan-abhi (approved):
Looks perfect, thank you!

fultonj (approved):
/lgtm thanks Francesco

kajinamit:
(I know I'm too late but ...)

May I know why we need this ? We should be able to use internal endpoint for all service interactions. both public/internal glance should be able to interact with the other services thus I expect both have access to internal endpoints.

PR #123, opened 2023-01-18
Allow attaching service pods to additional networks

stuggi:
For network isolation outgoing traffic, service pods need to be attached additional networks. To achive this, the the `k8s.v1.cni.cncf.io/networks` annotation needs to be set on the pods. This adds a networkAttachmentDefinitions parameter to the CRD where it can be specified to which additional networks the service pods should be attached to.

Depends-On: openstack-k8s-operators/lib-common#133


api/bases/glance.openstack.org_glances.yaml
```diff
@@ -5950,6 +5966,14 @@ spec:
                       to register in glance
                     type: string
                 type: object
+              networkAttachmentDefinitions:
+                description: NetworkAttachmentDefinitions list of network attachment
+                  definitions the service pod of extenal and internal API instances
+                  get attached to. Define them on the instance level will have no
+                  affect to overwrite the global ones.
```
stuggi, line 5973: or would it make sense that the 2 api instances could have different network attachments? right now I don't think so.

fultonj: Could each instance receive a different list of NADs? So the patch only adds one NAD list parameter but the list itself differs for internal vs external glance instances?

External:
```
k8s.v1.cni.cncf.io/networks: '[{"name": "publicapi", "namespace": "openstack"},{"name": "storage", "namespace": "openstack"}]'
```
Internal:
```
k8s.v1.cni.cncf.io/networks: '[{"name": "internalapi", "namespace": "openstack"},{"name": "storage", "namespace": "openstack"}]'
```

ASBishop: I don't know how this will all shake out in the end, but it definitely feels right that the internal and external glance services will need access to different networks.

fmount: I basically agree with the fact that we should have two separate `NADs` for internal/external: eventually, if at some point we decide to just run a single instance (e.g. when `file` is used as backend), we can merge the two structures but it would be cleaner having `NetworkAttachmentDefinitionInternal` and `NetworkAttachmentDefinitionExternal` in the top-level CR and assign the right one to the underlying service when the `Deployment` is created.

@stuggi not sure about your thoughts here: as long as the networking picture is not well defined I guess it's ok having this patch merged and follow up on the internal/external split.
With this change the two `GlanceAPI` instances will receive all the networks defined, but given you're doing the change it would be great setting up split mode for the network as well.

stuggi: For the initial version I guess we could then just have the two in the service spec, remove it from the top level glance and the db-init job from glance controller use the one from the internal glanceapi spec for its job?

ASBishop: That sounds good.

api/v1beta1/glanceapi_types.go
```diff
@@ -107,6 +107,10 @@ type GlanceAPISpec struct {
 	// +kubebuilder:validation:Optional
 	// ExtraMounts containing conf files and credentials
 	ExtraMounts []GlanceExtraVolMounts `json:"extraMounts"`
+
+	// +kubebuilder:validation:Optional
+	// NetworkAttachmentDefinitions list of network attachment definitions the service pod gets attached to
+	NetworkAttachmentDefinitions []string `json:"networkAttachmentDefinitions"`
```
fmount, line 113: So, not in the main `Glance` type, but in the `GlanceAPI` CR, if we add something like:

```
spec:
  ...
  ...
  networkAttachmentDefinitions:
      - internalapi
      - storage
  ...
  ...
```

we should be able to include the pre-existing NADs to the resulting POD.
@stuggi am I understanding it right?

fmount: @stuggi Maybe we can add a section in the `README` documenting how to set `NADs` and provide an example in the `samples/` directory.

stuggi: right, with the current  version you specify it per api instance, like in your example. works nice for me to attaching glance to a ceph deployed via https://github.com/openstack-k8s-operators/install_yamls/pull/54 . If you want to test it, I have a customized openstack-operator at `OPENSTACK_IMG=quay.io/mschuppe/openstack-operator-index:e2b9953ad06cdc2fc937aa9e5ac57e2bfef4141f` and used to following modifications:

```
     template:
       databaseInstance: openstack
       containerImage: quay.io/tripleozedcentos9/openstack-glance-api:current-tripleo
-      storageClass: ""
-      storageRequest: 10G
+      customServiceConfig: |
+        [DEFAULT]
+        enabled_backends = default_backend:rbd
+        debug = true
+        [glance_store]
+        default_backend = default_backend
+        [default_backend]
+        rbd_store_ceph_conf = /etc/ceph/ceph.conf
+        store_description = "RBD backend"
+        rbd_store_pool = images
+        rbd_store_user = admin
+      extraMounts:
+      - extraVol:
+        - extraVolType: Ceph
+          mounts:
+          - mountPath: /etc/ceph
+            name: ceph
+            readOnly: true
+          propagation:
+          - Glance
+          volumes:
+          - name: ceph
+            projected:
+              sources:
+              - secret:
+                  name: ceph-conf-files
       glanceAPIInternal:
         containerImage: quay.io/tripleozedcentos9/openstack-glance-api:current-tripleo
+        networkAttachmentDefinitions:
+        - storage
       glanceAPIExternal:
         containerImage: quay.io/tripleozedcentos9/openstack-glance-api:current-tripleo
+        networkAttachmentDefinitions:
+        - storage
```

created an image as
```
$ openstack image list
$ curl -L -o /tmp/cirros-0.5.2-x86_64-disk.img http://download.cirros-cloud.net/0.5.2/cirros-0.5.2-x86_64-disk.img
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   273  100   273    0     0    475      0 --:--:-- --:--:-- --:--:--   474
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 15.5M  100 15.5M    0     0  3595k      0  0:00:04  0:00:04 --:--:-- 4928k

$ qemu-img convert -O raw /tmp/cirros-0.5.2-x86_64-disk.img /tmp/cirros-0.5.2-x86_64-disk.img.raw
$ openstack image create --container-format bare --disk-format raw --file /tmp/cirros-0.5.2-x86_64-disk.img.raw cirros

$ openstack image list
+--------------------------------------+--------+--------+
| ID                                   | Name   | Status |
+--------------------------------------+--------+--------+
| 1c876c6a-046e-4f1b-9e66-e8895a58d142 | cirros | active |
+--------------------------------------+--------+--------+

$ oc rsh ceph
sh-4.4# rbd -p images list
1c876c6a-046e-4f1b-9e66-e8895a58d142
```

I plan to update the README of  this PR to show this example.

fmount: ack and thanks, I guess (but that could be a follow up) that we can introduce the `networkAttachmentDefinition` at the top level and make the split (internal vs external) for the sub CRs.
FWIW I tried your PR and I can reach the `GlanceAPI` pod from the `MetalLB` endpoint.

stuggi:
closing in favor of https://github.com/openstack-k8s-operators/glance-operator/pull/144

PR #124, opened 2023-01-19
Enable image import plugins

konan-abhi:
This patch will add glance-image-import.conf with default internal plugin 'no_op'. In feature we need to make sure to remove 'no_op' from plugin list if we enabled any valid plugin.

### DIFF
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -522,6 +522,9 @@ func (r *GlanceAPIReconciler) generateServiceConfigMaps(
 		templateParameters["ShowMultipleLocations"] = false
 	}
 
+    // Configure glance image import internal plugins, default to 'no_op'
+    templateParameters["GlanceImageImportPlugins"] = ['no_op']
+
 	cms := []util.Template{
 		// ScriptsConfigMap
 		{
--- a/templates/glance/config/glance-api-config.json
+++ b/templates/glance/config/glance-api-config.json
@@ -8,6 +8,12 @@
         "perm": "0600"
       },
       {
+        "source": "/var/lib/config-data/merged/glance-image-import.conf",
+        "dest": "/etc/glance/glance.conf.d/glance-image-import.conf",
+        "owner": "glance",
+        "perm": "0600"
+      },
+      {
         "source": "/var/lib/config-data/merged/policy.yaml",
         "dest": "/etc/glance/glance.conf.d/policy.yaml",
         "owner": "glance",
--- a/templates/glance/config/glance-image-import.conf
+++ b/templates/glance/config/glance-image-import.conf
@@ -1,0 +1,4 @@
+[DEFAULT]
+
+[image_import_opts]
+image_import_plugins = {{ .GlanceImageImportPlugins }}


controllers/glanceapi_controller.go
```diff
@@ -522,6 +522,9 @@ func (r *GlanceAPIReconciler) generateServiceConfigMaps(
 		templateParameters["ShowMultipleLocations"] = false
 	}
 
+    // Configure glance image import internal plugins, default to 'no_op'
+    templateParameters["GlanceImageImportPlugins"] = ['no_op']
```
fmount, line 526: I think this is a problem for golang, and this should actually be changed into something different.
The goal here is to render a string like "['no_op']" that, eventually, can be improved as long as plugins are enabled.
If you would like to give this default, a way would be having a function (in `pkg/glanceapi/`) like this:

```
func render_plugins() {

	var res string;
	var plugins = []string {
		"no_op",
		"p1",
		"p2",
	}

	res = "["
	for _, p := range plugins {
		res += fmt.Sprintf("'%s',", p)
	}

	strings.TrimSuffix(res, ",")
	res += "]"
	fmt.Println(res)
}
```

and assign to `templateParameters["GlanceImageImportPlugins"]` the result of this function.
However, if this is the default parameter we'd like to enforce, another and maybe better approach would just be removing this line and add the following to `glance-image-import.conf`:
```
[image_import_opts]
image_import_plugins = ['no_op']
```
and see if an override using `customServiceConfig` in the main `glance-api.conf` will work.
The point is: do we want the `glance-operator` being smart enough to `auto-enable` the right plugins?
If we have a set of values we always want to set we can just follow up on this change and change the `image_import_plugins` accordingly in the future, but for now enforcing ['no_op'] in the file should work.

fmount (changes requested):

controllers/glanceapi_controller.go
konan-abhi: Yes, we want glance-operator to auto-enabled 'image_conversion' plugin if we have configured ceph as glance backend. The motive of this patch is to introduce glance-image-import.conf with default no_op plugin and then modify it to set image_conversion related config options in the glance-image-import.conf

We also have other plugins such as;
inject_imgae_metadata, image_conversion but these plugins are totally deployment/customer specific. (i.e. to be enabled only if customer want to use it)

If you still think just setting ['no_op'] directly in conf file initially then I will modify the pull request.

fmount: > Yes, we want glance-operator to auto-enabled 'image_conversion' plugin if we have configured ceph as glance backend.

Yeah, I had something like that in mind: however, we first need to be able to "detect" that Ceph is enabled, and this can be done by inspecting the `extraMounts` (the `extraVolType` parameter) as well as parsing the `enabled_backends` (which is not part of this PR).

>The motive of this patch is to introduce glance-image-import.conf with default no_op plugin and then modify it to set image_conversion related config options in the glance-image-import.conf

We can set/enable `image_conversion` if Ceph is enabled, but this is something we can't easily do atm
> 
> We also have other plugins such as; inject_image_metadata, image_conversion but these plugins are totally deployment/customer specific. (i.e. to be enabled only if customer want to use it)

So I suspect we want the customer (human operator) being able to customize/override this option: am I understanding it correctly? 
In this case, we can `no_op` by default and rely on overrides, which is easier to handle.
If we go with this approach, we don't have to "detect" that Ceph is the backend, because it's the human operator, in the first place, that inject the Ceph bits (so he's supposed to be aware) and we should document how the following:

```
[image_import_opts]
image_import_plugins = ['no_op']
```
should be customized.

> 
> If you still think just setting ['no_op'] directly in conf file initially then I will modify the pull request.
Yeah, I think it's cleaner for now, and we can focus on finding a way to use `customServiceConfig` to override that section.
If the user provides something like:

```
kind: Glance
metadata:
  name: glance
spec:
  serviceUser: glance
  ...
  customServiceConfig: |
    [DEFAULT]
    enabled_backends=default_backend:rbd,ceph1:rbd
    enabled_backends=default_backend:rbd
    [glance_store]
    [default_backend]
    rbd_store_ceph_conf=/etc/ceph/ceph.conf
    rbd_store_user=admin
    rbd_store_pool=images
    store_description=Ceph glance store backend.
  databaseInstance: openstack
 ...
 ...
  extraMounts:
    - name: v1
      region: r1
      extraVol:
          extraVolType: Ceph
          volumes:
          - name: ceph
            projected:
              sources:
              - secret:
                  name: ceph-conf-files
          mounts:
          - name: ceph
            mountPath: "/etc/ceph"
```

we should try to inject the customized parameter with the Ceph bits.
Let me know your thoughts about this!

fultonj: FWIW I'd add the following directly to glance-image-import.conf:
```
[image_import_opts]
image_import_plugins = ['no_op']
```
and then have an expectation that the user would override the `image_import_plugins` within the `customServiceConfig`. The same way we expect the user to override `customServiceConfig` if they want to configure a backend like Ceph. Otherwise we go down a long road and write a lot of code to have some layer of abstraction on top of standard Glance config files which seems to simplify in the beginning of the project but adds complexity over time.

fultonj (changes requested):
Since the linter is rejecting `controllers/glanceapi_controller.go` it will need an update. 

Do we want to make GlanceImageImportPlugins a feature of the operator and do detection or simply require the user to know about glance configuration and for them to provide their own override if they want to use a plugin?

I'm leaning towards requiring the user to provide their own override for the same reasons we require the user to provide their own override to configure a backend like Ceph.

fmount:
> Since the linter is rejecting `controllers/glanceapi_controller.go` it will need an update.
Removing L526 is fine I guess.
> 
> Do we want to make GlanceImageImportPlugins a feature of the operator and do detection or simply require the user to know about glance configuration and for them to provide their own override if they want to use a plugin?
> 
> I'm leaning towards requiring the user to provide their own override for the same reasons we require the user to provide their own override to configure a backend like Ceph.

I agree on requiring the users to provide their own overrides (they should already inject the Ceph bits anyway, hence a detection feature doesn't make a lot of sense in this context)

konan-abhi:
> > Since the linter is rejecting `controllers/glanceapi_controller.go` it will need an update.
> 
> Removing L526 is fine I guess.
> 
> > Do we want to make GlanceImageImportPlugins a feature of the operator and do detection or simply require the user to know about glance configuration and for them to provide their own override if they want to use a plugin?
> > I'm leaning towards requiring the user to provide their own override for the same reasons we require the user to provide their own override to configure a backend like Ceph.
> 
> I agree on requiring the users to provide their own overrides (they should already inject the Ceph bits anyway, hence a detection feature doesn't make a lot of sense in this context)

Your Suggestions sounds good to me. However I just have one question, if we provide image_conversion related options as suggested by you and @fultonj in 'customServiceConfig' then will those get reflected in glance-image-import.conf or glance-api.conf?

It does not matter if it gets reflected in glance-api.conf but we do have specific conf file for import related configurations so ideally they should get reflected in glance-image-import.conf.

fultonj:
> Your Suggestions sounds good to me. However I just have one question, if we provide image_conversion related options as suggested by you and @fultonj in 'customServiceConfig' then will those get reflected in glance-image-import.conf or glance-api.conf?

It will end up in the `glance-api.conf`.

If you prefer it end up in `glance-image-import.conf`, then I think we'd need a new template here:

https://github.com/openstack-k8s-operators/glance-operator/tree/master/templates/glance/config

and some kind of new `customServiceConfigImport` which works like `customServiceConfig` but parallel to it. I wonder what other operators are doing?

@Akrog any thoughts?

konan-abhi:
> > Since the linter is rejecting `controllers/glanceapi_controller.go` it will need an update.
> 
> Removing L526 is fine I guess.
> 
> > Do we want to make GlanceImageImportPlugins a feature of the operator and do detection or simply require the user to know about glance configuration and for them to provide their own override if they want to use a plugin?
> > I'm leaning towards requiring the user to provide their own override for the same reasons we require the user to provide their own override to configure a backend like Ceph.
> 
> I agree on requiring the users to provide their own overrides (they should already inject the Ceph bits anyway, hence a detection feature doesn't make a lot of sense in this context)

Your Suggestions sounds good to me. However I just have one question, if we provide image_conversion related options as suggested by you and @fultonj in 'customServiceConfig' then will those get reflected in glance-image-import.conf or glance-api.conf?

It does not matter if it gets reflected in glance-api.conf but we do have specific conf file for import related configurations so ideally they should get reflected in glance-image-import.conf.

> > Your Suggestions sounds good to me. However I just have one question, if we provide image_conversion related options as suggested by you and @fultonj in 'customServiceConfig' then will those get reflected in glance-image-import.conf or glance-api.conf?
> 
> It will end up in the `glance-api.conf`.
> 
> If you prefer it end up in `glance-image-import.conf`, then I think we'd need a new template here:
> 
> https://github.com/openstack-k8s-operators/glance-operator/tree/master/templates/glance/config
> 
> and some kind of new `customServiceConfigImport` which works like `customServiceConfig` but parallel to it. I wonder what other operators are doing?
> 
> @Akrog any thoughts?

This sounds good to me, lets see what other operators are doing. The main reason I want to keep options in glance-image-import.conf file is to avoid changes in our deployment documentation where we already referring to the above file.

fmount:
> > > Since the linter is rejecting `controllers/glanceapi_controller.go` it will need an update.
> > 
> > 
> > Removing L526 is fine I guess.
> > > Do we want to make GlanceImageImportPlugins a feature of the operator and do detection or simply require the user to know about glance configuration and for them to provide their own override if they want to use a plugin?
> > > I'm leaning towards requiring the user to provide their own override for the same reasons we require the user to provide their own override to configure a backend like Ceph.
> > 
> > 
> > I agree on requiring the users to provide their own overrides (they should already inject the Ceph bits anyway, hence a detection feature doesn't make a lot of sense in this context)
> 
> Your Suggestions sounds good to me. However I just have one question, if we provide image_conversion related options as suggested by you and @fultonj in 'customServiceConfig' then will those get reflected in glance-image-import.conf or glance-api.conf?
> 
> It does not matter if it gets reflected in glance-api.conf but we do have specific conf file for import related configurations so ideally they should get reflected in glance-image-import.conf.
> 
> > > Your Suggestions sounds good to me. However I just have one question, if we provide image_conversion related options as suggested by you and @fultonj in 'customServiceConfig' then will those get reflected in glance-image-import.conf or glance-api.conf?
> > 
> > 
> > It will end up in the `glance-api.conf`.
> > If you prefer it end up in `glance-image-import.conf`, then I think we'd need a new template here:
> > https://github.com/openstack-k8s-operators/glance-operator/tree/master/templates/glance/config
> > and some kind of new `customServiceConfigImport` which works like `customServiceConfig` but parallel to it. I wonder what other operators are doing?
> > @Akrog any thoughts?
> 
> This sounds good to me, lets see what other operators are doing. The main reason I want to keep options in glance-image-import.conf file is to avoid changes in our deployment documentation where we already referring to the above file.

I think we shouldn't expose something like `customServiceConfigImport` which works like `customServiceConfig` because
I'm worried about the fact we start a pattern which is not followed by other operators and doesn't improve the user experience when we talk about injecting/overriding config bits.
AFAIK other operators (e.g. `cinder`) don't have a custom flow from different files, and everything is customized from the same `templates/` config.
I'm not sure I'm overlooking something in the way the config works, but,  from what I read in the code, a `customServiceConfigFile` (which is a const [1]) is created, and the service is started pointing to the `glance.conf.d` directory, where `01-custom.conf` is supposed to collect all the overrides:

`/usr/bin/python3 /usr/bin/glance-api --config-dir /etc/glance/glance.conf.d`

so ... IMO having the default value to `['no_op']` and overriding this block the same way we do with the main conf file
should be good enough, and it doesn't affect the user experience (we don't have to tell users to learn all the CR fields and how this pattern works and put the right overrides in the right places).

Not sure the path you want to take from here but I would keep if simple when possible.

[1]  https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/const.go#L24

konan-abhi:
> > > > Since the linter is rejecting `controllers/glanceapi_controller.go` it will need an update.
> > > 
> > > 
> > > Removing L526 is fine I guess.
> > > > Do we want to make GlanceImageImportPlugins a feature of the operator and do detection or simply require the user to know about glance configuration and for them to provide their own override if they want to use a plugin?
> > > > I'm leaning towards requiring the user to provide their own override for the same reasons we require the user to provide their own override to configure a backend like Ceph.
> > > 
> > > 
> > > I agree on requiring the users to provide their own overrides (they should already inject the Ceph bits anyway, hence a detection feature doesn't make a lot of sense in this context)
> > 
> > 
> > Your Suggestions sounds good to me. However I just have one question, if we provide image_conversion related options as suggested by you and @fultonj in 'customServiceConfig' then will those get reflected in glance-image-import.conf or glance-api.conf?
> > It does not matter if it gets reflected in glance-api.conf but we do have specific conf file for import related configurations so ideally they should get reflected in glance-image-import.conf.
> > > > Your Suggestions sounds good to me. However I just have one question, if we provide image_conversion related options as suggested by you and @fultonj in 'customServiceConfig' then will those get reflected in glance-image-import.conf or glance-api.conf?
> > > 
> > > 
> > > It will end up in the `glance-api.conf`.
> > > If you prefer it end up in `glance-image-import.conf`, then I think we'd need a new template here:
> > > https://github.com/openstack-k8s-operators/glance-operator/tree/master/templates/glance/config
> > > and some kind of new `customServiceConfigImport` which works like `customServiceConfig` but parallel to it. I wonder what other operators are doing?
> > > @Akrog any thoughts?
> > 
> > 
> > This sounds good to me, lets see what other operators are doing. The main reason I want to keep options in glance-image-import.conf file is to avoid changes in our deployment documentation where we already referring to the above file.
> 
> I think we shouldn't expose something like `customServiceConfigImport` which works like `customServiceConfig` because I'm worried about the fact we start a pattern which is not followed by other operators and doesn't improve the user experience when we talk about injecting/overriding config bits. AFAIK other operators (e.g. `cinder`) don't have a custom flow from different files, and everything is customized from the same `templates/` config. I'm not sure I'm overlooking something in the way the config works, but, from what I read in the code, a `customServiceConfigFile` (which is a const [1]) is created, and the service is started pointing to the `glance.conf.d` directory, where `01-custom.conf` is supposed to collect all the overrides:
> 
> `/usr/bin/python3 /usr/bin/glance-api --config-dir /etc/glance/glance.conf.d`
> 
> so ... IMO having the default value to `['no_op']` and overriding this block the same way we do with the main conf file should be good enough, and it doesn't affect the user experience (we don't have to tell users to learn all the CR fields and how this pattern works and put the right overrides in the right places).
> 
> Not sure the path you want to take from here but I would keep if simple when possible.
> 
> [1] https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/const.go#L24

I don't have major opinion here, my only concern is we need to be more specific in our documentation about these procedures (enabling the import plugins) as with prior deployments some config options were under specific config file but from now on it will be glance-api.conf (or a custom.conf from the glance.conf.d directory).

2nd Point;
if we have one customeServiceConfig (from one template) which is enabling the image_conversion plugin and then we also want to enable inject_image_metadata plugin then it should be taken care to not override existing image_conversion plugin but both the plugins should be listed there (Here I am assuming the templates to enable ceph and inject_metadata plugins will be different).

fmount:
> 



> > > > > Since the linter is rejecting `controllers/glanceapi_controller.go` it will need an update.
> > > > 
> > > > 
> > > > Removing L526 is fine I guess.
> > > > > Do we want to make GlanceImageImportPlugins a feature of the operator and do detection or simply require the user to know about glance configuration and for them to provide their own override if they want to use a plugin?
> > > > > I'm leaning towards requiring the user to provide their own override for the same reasons we require the user to provide their own override to configure a backend like Ceph.
> > > > 
> > > > 
> > > > I agree on requiring the users to provide their own overrides (they should already inject the Ceph bits anyway, hence a detection feature doesn't make a lot of sense in this context)
> > > 
> > > 
> > > Your Suggestions sounds good to me. However I just have one question, if we provide image_conversion related options as suggested by you and @fultonj in 'customServiceConfig' then will those get reflected in glance-image-import.conf or glance-api.conf?
> > > It does not matter if it gets reflected in glance-api.conf but we do have specific conf file for import related configurations so ideally they should get reflected in glance-image-import.conf.
> > > > > Your Suggestions sounds good to me. However I just have one question, if we provide image_conversion related options as suggested by you and @fultonj in 'customServiceConfig' then will those get reflected in glance-image-import.conf or glance-api.conf?
> > > > 
> > > > 
> > > > It will end up in the `glance-api.conf`.
> > > > If you prefer it end up in `glance-image-import.conf`, then I think we'd need a new template here:
> > > > https://github.com/openstack-k8s-operators/glance-operator/tree/master/templates/glance/config
> > > > and some kind of new `customServiceConfigImport` which works like `customServiceConfig` but parallel to it. I wonder what other operators are doing?
> > > > @Akrog any thoughts?
> > > 
> > > 
> > > This sounds good to me, lets see what other operators are doing. The main reason I want to keep options in glance-image-import.conf file is to avoid changes in our deployment documentation where we already referring to the above file.
> > 
> > 
> > I think we shouldn't expose something like `customServiceConfigImport` which works like `customServiceConfig` because I'm worried about the fact we start a pattern which is not followed by other operators and doesn't improve the user experience when we talk about injecting/overriding config bits. AFAIK other operators (e.g. `cinder`) don't have a custom flow from different files, and everything is customized from the same `templates/` config. I'm not sure I'm overlooking something in the way the config works, but, from what I read in the code, a `customServiceConfigFile` (which is a const [1]) is created, and the service is started pointing to the `glance.conf.d` directory, where `01-custom.conf` is supposed to collect all the overrides:
> > `/usr/bin/python3 /usr/bin/glance-api --config-dir /etc/glance/glance.conf.d`
> > so ... IMO having the default value to `['no_op']` and overriding this block the same way we do with the main conf file should be good enough, and it doesn't affect the user experience (we don't have to tell users to learn all the CR fields and how this pattern works and put the right overrides in the right places).
> > Not sure the path you want to take from here but I would keep if simple when possible.
> > [1] https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/const.go#L24
> 
> I don't have major opinion here, my only concern is we need to be more specific in our documentation about these procedures (enabling the import plugins) as with prior deployments some config options were under specific config file but from now on it will be glance-api.conf (or a custom.conf from the glance.conf.d directory).
> 

prior == `TripleO` ?

I guess it's just fine documenting the way things works in this context (we have to do it anyway and I created a
Jira for that :P)
> 2nd Point; if we have one customeServiceConfig (from one template) which is enabling the image_conversion plugin and then we also want to enable inject_image_metadata plugin then it should be taken care to not override existing image_conversion plugin but both the plugins should be listed there (Here I am assuming the templates to enable ceph and inject_metadata plugins will be different).

I guess here you want to make sure we append and not override to reach something like:

```
[image_import_opts]
image_import_plugins = ['inject_image_metadata', 'image_conversion']
```

Let's sync up on this, but I guess we can unblock the PR removing that line and defaulting to `['no_op']` in the template with the current, existing flow.

konan-abhi:
> > 
> 
> > > > > > Since the linter is rejecting `controllers/glanceapi_controller.go` it will need an update.
> > > > > 
> > > > > 
> > > > > Removing L526 is fine I guess.
> > > > > > Do we want to make GlanceImageImportPlugins a feature of the operator and do detection or simply require the user to know about glance configuration and for them to provide their own override if they want to use a plugin?
> > > > > > I'm leaning towards requiring the user to provide their own override for the same reasons we require the user to provide their own override to configure a backend like Ceph.
> > > > > 
> > > > > 
> > > > > I agree on requiring the users to provide their own overrides (they should already inject the Ceph bits anyway, hence a detection feature doesn't make a lot of sense in this context)
> > > > 
> > > > 
> > > > Your Suggestions sounds good to me. However I just have one question, if we provide image_conversion related options as suggested by you and @fultonj in 'customServiceConfig' then will those get reflected in glance-image-import.conf or glance-api.conf?
> > > > It does not matter if it gets reflected in glance-api.conf but we do have specific conf file for import related configurations so ideally they should get reflected in glance-image-import.conf.
> > > > > > Your Suggestions sounds good to me. However I just have one question, if we provide image_conversion related options as suggested by you and @fultonj in 'customServiceConfig' then will those get reflected in glance-image-import.conf or glance-api.conf?
> > > > > 
> > > > > 
> > > > > It will end up in the `glance-api.conf`.
> > > > > If you prefer it end up in `glance-image-import.conf`, then I think we'd need a new template here:
> > > > > https://github.com/openstack-k8s-operators/glance-operator/tree/master/templates/glance/config
> > > > > and some kind of new `customServiceConfigImport` which works like `customServiceConfig` but parallel to it. I wonder what other operators are doing?
> > > > > @Akrog any thoughts?
> > > > 
> > > > 
> > > > This sounds good to me, lets see what other operators are doing. The main reason I want to keep options in glance-image-import.conf file is to avoid changes in our deployment documentation where we already referring to the above file.
> > > 
> > > 
> > > I think we shouldn't expose something like `customServiceConfigImport` which works like `customServiceConfig` because I'm worried about the fact we start a pattern which is not followed by other operators and doesn't improve the user experience when we talk about injecting/overriding config bits. AFAIK other operators (e.g. `cinder`) don't have a custom flow from different files, and everything is customized from the same `templates/` config. I'm not sure I'm overlooking something in the way the config works, but, from what I read in the code, a `customServiceConfigFile` (which is a const [1]) is created, and the service is started pointing to the `glance.conf.d` directory, where `01-custom.conf` is supposed to collect all the overrides:
> > > `/usr/bin/python3 /usr/bin/glance-api --config-dir /etc/glance/glance.conf.d`
> > > so ... IMO having the default value to `['no_op']` and overriding this block the same way we do with the main conf file should be good enough, and it doesn't affect the user experience (we don't have to tell users to learn all the CR fields and how this pattern works and put the right overrides in the right places).
> > > Not sure the path you want to take from here but I would keep if simple when possible.
> > > [1] https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/const.go#L24
> > 
> > 
> > I don't have major opinion here, my only concern is we need to be more specific in our documentation about these procedures (enabling the import plugins) as with prior deployments some config options were under specific config file but from now on it will be glance-api.conf (or a custom.conf from the glance.conf.d directory).
> 
> prior == `TripleO` ?
> 
> I guess it's just fine documenting the way things works in this context (we have to do it anyway and I created a Jira for that :P)
> 
> > 2nd Point; if we have one customeServiceConfig (from one template) which is enabling the image_conversion plugin and then we also want to enable inject_image_metadata plugin then it should be taken care to not override existing image_conversion plugin but both the plugins should be listed there (Here I am assuming the templates to enable ceph and inject_metadata plugins will be different).
> 
> I guess here you want to make sure we append and not override to reach something like:
> 
> ```
> [image_import_opts]
> image_import_plugins = ['inject_image_metadata', 'image_conversion']
> ```
> 
> Let's sync up on this, but I guess we can unblock the PR removing that line and defaulting to `['no_op']` in the template with the current, existing flow.

So I guess I need to remove glance-image-import.conf and changes from glance-api-config.json file as well and add ['no_op'] as default in glance-api.conf

Correct?

fmount:
> 



> > > 
> > 
> > 
> > > > > > > Since the linter is rejecting `controllers/glanceapi_controller.go` it will need an update.
> > > > > > 
> > > > > > 
> > > > > > Removing L526 is fine I guess.
> > > > > > > Do we want to make GlanceImageImportPlugins a feature of the operator and do detection or simply require the user to know about glance configuration and for them to provide their own override if they want to use a plugin?
> > > > > > > I'm leaning towards requiring the user to provide their own override for the same reasons we require the user to provide their own override to configure a backend like Ceph.
> > > > > > 
> > > > > > 
> > > > > > I agree on requiring the users to provide their own overrides (they should already inject the Ceph bits anyway, hence a detection feature doesn't make a lot of sense in this context)
> > > > > 
> > > > > 
> > > > > Your Suggestions sounds good to me. However I just have one question, if we provide image_conversion related options as suggested by you and @fultonj in 'customServiceConfig' then will those get reflected in glance-image-import.conf or glance-api.conf?
> > > > > It does not matter if it gets reflected in glance-api.conf but we do have specific conf file for import related configurations so ideally they should get reflected in glance-image-import.conf.
> > > > > > > Your Suggestions sounds good to me. However I just have one question, if we provide image_conversion related options as suggested by you and @fultonj in 'customServiceConfig' then will those get reflected in glance-image-import.conf or glance-api.conf?
> > > > > > 
> > > > > > 
> > > > > > It will end up in the `glance-api.conf`.
> > > > > > If you prefer it end up in `glance-image-import.conf`, then I think we'd need a new template here:
> > > > > > https://github.com/openstack-k8s-operators/glance-operator/tree/master/templates/glance/config
> > > > > > and some kind of new `customServiceConfigImport` which works like `customServiceConfig` but parallel to it. I wonder what other operators are doing?
> > > > > > @Akrog any thoughts?
> > > > > 
> > > > > 
> > > > > This sounds good to me, lets see what other operators are doing. The main reason I want to keep options in glance-image-import.conf file is to avoid changes in our deployment documentation where we already referring to the above file.
> > > > 
> > > > 
> > > > I think we shouldn't expose something like `customServiceConfigImport` which works like `customServiceConfig` because I'm worried about the fact we start a pattern which is not followed by other operators and doesn't improve the user experience when we talk about injecting/overriding config bits. AFAIK other operators (e.g. `cinder`) don't have a custom flow from different files, and everything is customized from the same `templates/` config. I'm not sure I'm overlooking something in the way the config works, but, from what I read in the code, a `customServiceConfigFile` (which is a const [1]) is created, and the service is started pointing to the `glance.conf.d` directory, where `01-custom.conf` is supposed to collect all the overrides:
> > > > `/usr/bin/python3 /usr/bin/glance-api --config-dir /etc/glance/glance.conf.d`
> > > > so ... IMO having the default value to `['no_op']` and overriding this block the same way we do with the main conf file should be good enough, and it doesn't affect the user experience (we don't have to tell users to learn all the CR fields and how this pattern works and put the right overrides in the right places).
> > > > Not sure the path you want to take from here but I would keep if simple when possible.
> > > > [1] https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/const.go#L24
> > > 
> > > 
> > > I don't have major opinion here, my only concern is we need to be more specific in our documentation about these procedures (enabling the import plugins) as with prior deployments some config options were under specific config file but from now on it will be glance-api.conf (or a custom.conf from the glance.conf.d directory).
> > 
> > 
> > prior == `TripleO` ?
> > I guess it's just fine documenting the way things works in this context (we have to do it anyway and I created a Jira for that :P)
> > > 2nd Point; if we have one customeServiceConfig (from one template) which is enabling the image_conversion plugin and then we also want to enable inject_image_metadata plugin then it should be taken care to not override existing image_conversion plugin but both the plugins should be listed there (Here I am assuming the templates to enable ceph and inject_metadata plugins will be different).
> > 
> > 
> > I guess here you want to make sure we append and not override to reach something like:
> > ```
> > [image_import_opts]
> > image_import_plugins = ['inject_image_metadata', 'image_conversion']
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Let's sync up on this, but I guess we can unblock the PR removing that line and defaulting to `['no_op']` in the template with the current, existing flow.
> 
> So I guess I need to remove glance-image-import.conf and changes from glance-api-config.json file as well and add ['no_op'] as default in glance-api.conf
> 
> Correct?

Whatever works best for you. 
You can actually keep the `glance-image-import.conf` with the ['no_op'] default and leave it in `glance-api-config.json`.
In case people provide their own override via `customeServiceConfig` that section is supposed to take over the defaults, but we should test and double check on a local environment.
The other strategy is to remove glance-import.conf and just rely on the global glance-api.conf, where we know that the override will work.
Feel free to update the patch and I help you on the testing aspect!

fultonj:
I agree about not introducing `customServiceConfigImport`. Keeping `glance-image-import.conf` with the `['no_op']` default and then documenting that the user can override it with `customServiceConfig` sounds good to me.

In that case, I gather that `/etc/glance/glance.conf.d/` will contain `01-custom.conf` and it will override the default in `glance-image-import.conf`.

konan-abhi:
Moved to https://github.com/openstack-k8s-operators/glance-operator/pull/128

PR #128, opened 2023-01-20
Enable image import plugins

konan-abhi:
This patch will add glance-image-import.conf with default internal plugin 'no_op'. In feature we need to make sure to remove 'no_op' from plugin list if we enabled any valid plugin.


fultonj:
Looks good to me

fmount (approved):

PR #133, opened 2023-01-27
Webhook

rabi:
Adds validating webhook  scaffolding for glance service to add validations in subsequent PRs. Also makes changes to controller-manager labels to make them unique as it conflicts with other operators making the webhooks unusable.


Makefile
```diff
@@ -210,7 +210,7 @@ bundle: manifests kustomize ## Generate bundle manifests and metadata, then vali
 	operator-sdk bundle validate ./bundle
 
 .PHONY: bundle-build
-bundle-build: ## Build the bundle image.
+bundle-build: bundle ## Build the bundle image.
```
abays, line 213: Is this a change we're replicating across operators?  Operator-SDK doesn't scaffold it this way.  It seems we have a few operators with this extra change, but most currently do not have it.

rabi: Yeah I'm adding it across operator repos, as you can't run 'bundle-build' without running 'bundle' target (i.e there is no default bundle.Dockerfile).  Also, it's better to generate it rather than use a default. We can probably remove the default bundle.Dockerfile in repos where they exist.

abays: Ack

main.go
```diff
@@ -116,6 +116,10 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "Glance")
 		os.Exit(1)
 	}
+	if err = (&glancev1.Glance{}).SetupWebhookWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create webhook", "webhook", "Glance")
+		os.Exit(1)
+	}
```
abays, line 122: Should we wrap this in `if strings.ToLower(os.Getenv("ENABLE_WEBHOOKS")) != "false" {` like you had for the other PR?  Or did you perhaps figure out a way to get the webhooks to run locally?

Makefile
rabi: @dprince I thought I saw it in some repos. But can't find one other than the custom one used in openstack-operator. I think the issue I faced is it always used the one locally when running bundle-build rather than regenerating one.

main.go
rabi: I can add that, but not sure I understand what you mean by locally.  Do you mean webhook as a different local process? It works for me  when the operator is deployed on crc. I just looked at https://book.kubebuilder.io/cronjob-tutorial/running-webhook.html. Also, I don't know how you can set the env variable to enable disable the webhook.

abays: By locally I mean at the CLI via (in your `glance-operator` dir): `ENABLE_WEBHOOKS=false GOWORK= OPERATOR_TEMPLATES=./templates make run`.  This can be useful for developmental iterating and not having to redeploy the operator via OLM, etc, when you want to test changes.

rabi: I think it works fine if you generate the self-signed certs. We can add a target to generate self-signed certs to Makefile.

```
[stack@osp-df-2 glance-operator]$ openssl req -newkey rsa:4096 -days 3650 -nodes -x509   -keyout /tmp/k8s-webhook-server/serving-certs/tls.key   -out /tmp/k8s-webhook-server/serving-certs/tls.crt
....+...+.....+...+....+...+..+.+......+...........+.........................+..+.+.....+......+...+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++*.+...+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++*..+..................+.........+.+............+..............+.+.........+...+........+...+.........+...+...............+...+.+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
.+....................+......+.........+....+..............+...+.........+.........+.+......+..+...+.......+...+...+.....+.+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++*................+.....+..........+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++*....+......+......+...+.....+.+............+..+...+...+......................+.................+...+...+....+..+....+.........+....................+.+.........+.....+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
-----
You are about to be asked to enter information that will be incorporated
into your certificate request.
What you are about to enter is what is called a Distinguished Name or a DN.
There are quite a few fields but you can leave some blank
For some fields there will be a default value,
If you enter '.', the field will be left blank.
-----
Country Name (2 letter code) [XX]:IN
State or Province Name (full name) []:MH
Locality Name (eg, city) [Default City]:PNQ
Organization Name (eg, company) [Default Company Ltd]:Red Hat
Organizational Unit Name (eg, section) []:DF
Common Name (eg, your name or your server's hostname) []:osp-df2
Email Address []:
[stack@osp-df-2 glance-operator]$ GOWORK= OPERATOR_TEMPLATES=./templates make run
/home/stack/work/glance-operator/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases && \
rm -f api/bases/* && cp -a config/crd/bases api/
/home/stack/work/glance-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
test -f go.work || go work init
go work use .
go work use ./api
go vet ./...
go vet ./... ./api/...
go run ./main.go
I0131 07:35:39.771728 1634555 request.go:682] Waited for 1.038209601s due to client-side throttling, not priority and fairness, request: GET:https://api.crc.testing:6443/apis/security.openshift.io/v1?timeout=32s
1.6751685414748962e+09	INFO	controller-runtime.metrics	Metrics server is starting to listen	{"addr": ":8080"}
1.675168541478722e+09	INFO	controller-runtime.builder	skip registering a mutating webhook, object does not implement admission.Defaulter or WithDefaulter wasn't called	{"GVK": "glance.openstack.org/v1beta1, Kind=Glance"}
1.6751685414787903e+09	INFO	controller-runtime.builder	Registering a validating webhook	{"GVK": "glance.openstack.org/v1beta1, Kind=Glance", "path": "/validate-glance-openstack-org-v1beta1-glance"}
1.6751685414788954e+09	INFO	controller-runtime.webhook	Registering webhook	{"path": "/validate-glance-openstack-org-v1beta1-glance"}
1.6751685414790215e+09	INFO	setup	starting manager
1.675168541479233e+09	INFO	controller-runtime.webhook.webhooks	Starting webhook server
1.6751685414793482e+09	INFO	Starting server	{"kind": "health probe", "addr": "[::]:8081"}
1.675168541479361e+09	INFO	Starting server	{"path": "/metrics", "kind": "metrics", "addr": "[::]:8080"}
1.6751685414794888e+09	INFO	Starting EventSource	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "source": "kind source: *v1beta1.GlanceAPI"}
1.6751685414794989e+09	INFO	Starting EventSource	{"controller": "glance", "controllerGroup": "glance.openstack.org", "controllerKind": "Glance", "source": "kind source: *v1beta1.Glance"}
1.675168541479546e+09	INFO	Starting EventSource	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "source": "kind source: *v1beta1.KeystoneEndpoint"}
1.6751685414795575e+09	INFO	Starting EventSource	{"controller": "glance", "controllerGroup": "glance.openstack.org", "controllerKind": "Glance", "source": "kind source: *v1beta1.GlanceAPI"}
1.6751685414795642e+09	INFO	Starting EventSource	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "source": "kind source: *v1.Service"}
1.6751685414795775e+09	INFO	Starting EventSource	{"controller": "glance", "controllerGroup": "glance.openstack.org", "controllerKind": "Glance", "source": "kind source: *v1beta1.MariaDBDatabase"}
1.675168541479585e+09	INFO	Starting EventSource	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "source": "kind source: *v1.Secret"}
1.6751685414795997e+09	INFO	Starting EventSource	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "source": "kind source: *v1.ConfigMap"}
1.6751685414795897e+09	INFO	Starting EventSource	{"controller": "glance", "controllerGroup": "glance.openstack.org", "controllerKind": "Glance", "source": "kind source: *v1beta1.KeystoneService"}
1.6751685414796495e+09	INFO	Starting EventSource	{"controller": "glance", "controllerGroup": "glance.openstack.org", "controllerKind": "Glance", "source": "kind source: *v1.PersistentVolumeClaim"}
1.6751685414796588e+09	INFO	controller-runtime.certwatcher	Updated current TLS certificate
1.6751685414796715e+09	INFO	Starting EventSource	{"controller": "glance", "controllerGroup": "glance.openstack.org", "controllerKind": "Glance", "source": "kind source: *v1.Job"}
1.6751685414796872e+09	INFO	Starting EventSource	{"controller": "glance", "controllerGroup": "glance.openstack.org", "controllerKind": "Glance", "source": "kind source: *v1.ConfigMap"}
1.6751685414796984e+09	INFO	Starting EventSource	{"controller": "glance", "controllerGroup": "glance.openstack.org", "controllerKind": "Glance", "source": "kind source: *v1.Secret"}
1.6751685414796138e+09	INFO	Starting EventSource	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "source": "kind source: *v1.Deployment"}
1.6751685414797342e+09	INFO	Starting EventSource	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "source": "kind source: *v1.Route"}
1.675168541479744e+09	INFO	Starting Controller	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI"}
1.6751685414797142e+09	INFO	Starting Controller	{"controller": "glance", "controllerGroup": "glance.openstack.org", "controllerKind": "Glance"}
1.675168541479816e+09	INFO	controller-runtime.certwatcher	Starting certificate watcher
1.675168541479833e+09	INFO	controller-runtime.webhook	Serving webhook server	{"host": "", "port": 9443}

```

abays: That's awesome!  Let's get that Makefile target in place before we merge any of these PRs with webhooks, since without the `ENABLE_WEBHOOKS` check we would otherwise be breaking the local `make run` for them.

Makefile
dprince: @rabi so one use case to think about here is I often generate the bundle, mess with it, and then run a bundle build. I could do that manually (essentially manually build the Dockerfile). There isn't a whole lot of value in the bundle-build task as is...

rabi: The simple workflow for me after any PR to deploy with OLM involves, docker-build/docker-push, bundle-build/bundle-push, catalog-build/catalog-push. Do you think that's incorrect? Should not that Dockerfile be generated automatically every time without needing manual intervention(?)

rabi: @dprince I can remove that change from the PR if you think we should not change it. I can make it work locally for me with some hack.

main.go
rabi: Added target in the latest patchset.

Makefile
rabi: OK, removed it. I know it won't apply to everyone, but I guess there is a default workflow where folks don't tweak generated files.

config/certmanager/certificate.yaml
```diff
@@ -0,0 +1,39 @@
+# The following manifests contain a self-signed issuer CR and a certificate CR.
+# More document can be found at https://docs.cert-manager.io
+# WARNING: Targets CertManager v1.0. Check https://cert-manager.io/docs/installation/upgrading/ for breaking changes.
+apiVersion: cert-manager.io/v1
```
stuggi, line 4: do I get it correct that with this cert-manager is a requirement and needs to be installed?

abays: It just seems to work without needing any explicit install, as far as I have seen in my testing

fmount: > +1 from me. @dprince and @fmount, what do you think?

In general I'm ok with the patch,  but given we're in the development phase, something like:

`ENABLE_WEBHOOKS=false GOWORK= OPERATOR_TEMPLATES=./templates make run`

doesn't work anymore because we need to fix first the `create_self_signed_cert.sh`: I'll comment on the script.

`make run` works as expected, but I'd like to avoid breaking the good old way of running the operators locally :D

hack/create_self_signed_cert.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+set -ex
+
+TMPDIR={$TMPDIR:-"/tmp/k8s-webhook-server/serving-certs"}
```
fmount, line 4: TMPDIR=${TMPDIR:-"/tmp/k8s-webhook-server/serving-certs"}

hack/create_self_signed_cert.sh
```diff
@@ -0,0 +1,30 @@
...
+SERVICE={$SERVICE:-"glance"}
```
fmount, line 5: SERVICE=${SERVICE:-"glance"}

hack/create_self_signed_cert.sh
```diff
@@ -0,0 +1,30 @@
...
+NAMESPACE={$NAMESPACE:-"openstack"}
```
fmount, line 6: NAMESPACE=${NAMESPACE:-"openstack"}

hack/create_self_signed_cert.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
fmount, line 1: Can we change L{4,5,6} as per my comments? Otherwise if you just run `make generate-cert` the command will be:


```
+ openssl req -newkey rsa:4096 -days 3650 -nodes -x509 -keyout '{:-/tmp/k8s-webhook-server/serving-certs}/tls.key' -out '{:-/tmp/k8s-webhook-server/serving-certs}/tls.crt' -config '{:-/tmp/k8s-webhook-server/se
rving-certs}/tls.conf'
```

fmount (changes requested):

hack/create_self_signed_cert.sh
rabi: sorry that was typo... thanks for noticing. Fixed.

fmount: +1 Thanks!

fmount:
> +1 from me. @dprince and @fmount, what do you think?

It should be ready at this point. +1 on my side unless @dprince has more comments

config/certmanager/certificate.yaml
rabi: > do I get it correct that with this cert-manager is a requirement and needs to be installed?

Sorry, I missed to answer this question earlier. Though cert-manager scaffolding is created by sdk, cert-manager does not inject the certs atm. webhook uses the same OLM generated certificate as api service.

```
       volumeMounts:
        - mountPath: /tmp/k8s-webhook-server/serving-certs
          name: webhook-cert
          readOnly: true
        - mountPath: /apiserver.local.config/certificates
          name: apiservice-cert
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext:
        runAsNonRoot: true
      serviceAccount: keystone-operator-controller-manager
      serviceAccountName: keystone-operator-controller-manager
      terminationGracePeriodSeconds: 10
      volumes:
      - name: webhook-cert
        secret:
          defaultMode: 420
          items:
          - key: tls.crt
            path: tls.crt
          - key: tls.key
            path: tls.key
          secretName: keystone-operator-controller-manager-service-cert
      - name: apiservice-cert
        secret:
          defaultMode: 420
          items:
          - key: tls.crt
            path: apiserver.crt
          - key: tls.key
            path: apiserver.key
          secretName: keystone-operator-controller-manager-service-cert

```

dprince:
@rabi can we pick 1 operator (perhaps glance or keystone) and iterate on a proposal for what we want validations to look like before applying it to all the operators. I'd also like to tests how this plays with the openstack-operator's "umbrella" CRD for OpenStackControlplane in this regard.

rabi:
> iterate on a proposal for what we want validations to look like before applying it to all the operators

These are just scaffolds and certain other changes required to make the webhooks work when there are more than one operator deployed. Are you suggesting we should finish implementation of validation webhook for one operator before moving to another? I discussed with @abays and we agreed to make the basic changes to all operators to make the webhooks work. If we want to get it 'absolutely' right and implemented for one operator before moving to another I'm fine. But we're in iterative development as you mentioned, I guess:)

abays:
> > iterate on a proposal for what we want validations to look like before applying it to all the operators
> 
> These are just scaffolds and certain other changes required to make the webhooks work when there are more than one operator deployed. Are you suggesting we should finish implementation of validation webhook for one operator before moving to another? I discussed with @abays and we agreed to make the basic changes to all operators to make the webhooks work. If we want to get it 'absolutely' right and implemented for one operator before moving to another I'm fine. But we're in iterative development as you mentioned, I guess:)

We could consider this webhook PR the de facto standard for how we'll approach webhooks for other operators.  At this point we've iterated on reviews of it (and tested it) and I believe we're in agreement that this pattern will work going forward.

config/manager/manager.yaml
```diff
@@ -4,18 +4,18 @@ metadata:
   name: controller-manager
   namespace: system
   labels:
-    control-plane: controller-manager
+    control-plane: glance-controller-manager
```
abays, line 7: I just noticed one area where this particular change will cause us problems.  During CI we gather certain log/config data [1].  When we get the controller-manage logs, we select the pod via the `control-plane=controller-manager` label.  We might need to keep the old label but still add the new one.

https://github.com/gibizer/openshift-release/blob/7b87904de96e4b1f98d3b7a98cb4602c50500c09/ci-operator/step-registry/openstack-k8s-operators/gather/openstack-k8s-operators-gather-commands.sh#L16

rabi: Looks like regex is not supported for selectors[1], Unless we want to add a label for log collection, we change the command to 'oc get pods -n openstack --show-labels | awk '{if ($6 ~ "controller-manager") print $1}' | xargs -n1 oc logs'  in CI. I can propose a patch there if it's ok.

main.go
```diff
@@ -116,6 +117,12 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "Glance")
 		os.Exit(1)
 	}
+	if strings.ToLower(os.Getenv("ENABLE_WEBHOOKS")) != "false" {
```
gibizer, line 120: This can be made more user friendly with  https://pkg.go.dev/strconv#ParseBool

config/manager/manager.yaml
abays: Sorry, I missed your response here.  I think we should propose a patch for CI in this regard.

main.go
dprince: @gibizer we have prior art for this in other operators. Lets follow up on stuff like this later I think

config/manager/manager.yaml
dprince: agree with fixing in CI, it seems this is going to be required to make webhooks play nicely

dprince (approved):

config/manager/manager.yaml
rabi: Pushed https://github.com/openshift/release/pull/36394

PR #143, opened 2023-02-01
K8s.io updates

dprince:
This should resolve recent Renovate PR which was failing due to sigs.k8s.io/controller-runtime/pkg/envtest/printer no longer existing after the dependency bump.


abays (approved):
/lgtm

PR #144, opened 2023-02-03
Support attaching pods to networks using net-attach-defs and exposing api via metallb

stuggi:
* If needed, service pods can be attach to networks using the `networkAttachments` list in the CR. Its a list of pre-created net-attach-defs names. Per default the pods won't get attached.

```
spec:
...
      networkAttachments:
      - storage
```
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/133

* introduce externalEndpoints Parameters to configure metallb service using the pre-created metallb IPAddressPool and L2Advertisement . 

E.g. to expose the api on an internalapi network which
* should be registered as the internal endpoint
* its metallb ipaddresspool is configured in an osp-internalapi object
* and request a specific loadbalancerIP which is shared with other services (list of IPs for multi stack ipv4/ipv6)

```
...
spec:
...
      externalEndpoints:
      - endpoint: internal
        loadBalancerIPs:
        - 172.17.0.202
        ipAddressPool: osp-internalapi
        sharedIP: true
```

* removes registration of admin endpoint

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/160


README.md
```diff
@@ -175,6 +175,87 @@ spec:
...
+  ...
+  glanceAPIInternal:
+    ...
+    networkAttachents:
+    - storage
+  glanceAPIExternal:
+    ...
+    networkAttachents:
+    - storage
+...
```
fmount, line 229: Do we want to update the example adding the following:

```
 vip:
   - endpoint: storage
     ip: 172.18.0.202  
     ipAddressPool: storage
     sharedIP: true
```

or something that explain how the CR can be composed?

api/v1beta1/glanceapi_types.go
```diff
@@ -107,6 +107,34 @@ type GlanceAPISpec struct {
...
+	// NetworkAttachments list of network attachment definitions the service pod gets attached to
+	NetworkAttachments []string `json:"networkAttachments"`
+
+	// +kubebuilder:validation:Optional
+	// ExternalEndpoints, expose a VIP via MetalLB on the pre-created address pool
+	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints"`
+}
+
+// MetalLBConfig to configure the MetalLB loadbalancer service
+type MetalLBConfig struct {
```
fmount, line 121: @stuggi quick question: `ExternalEndpoints` is optional in the CR, which is ok, but as long as it's defined, I'm wondering if we should make sure that the `Endpoint` (`internal` or `public`) is always defined (otherwise it should fail). 
Should then we need to apply the same pattern at least to the `IP` address ?
I'm basically ok about keeping the struct optional, but when it's defined I think we need to make sure all the required parameters are definied.

api/go.sum
```diff
@@ -261,6 +259,8 @@ github.com/stretchr/testify v1.7.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/
 github.com/stretchr/testify v1.8.0/go.mod h1:yNjHg4UonilssWZ8iaSj1OCr/vHnekPRkoO+kdMU+MU=
 github.com/stretchr/testify v1.8.1 h1:w7B6lhMri9wdJUVmEZPGGhZzrYTPvgJArz7wNPgYKsk=
 github.com/stretchr/testify v1.8.1/go.mod h1:w2LPCIKwWwSfY2zedu0+kehJoqGctiVI29o6fzry7u4=
+github.com/stuggi/lib-common/modules/common v0.0.0-20230201125726-9ae3213ded0d h1:/jbMyhUV1RTlYhKDbfUhGR7jlaI99vnxJf56Jix25Qk=
```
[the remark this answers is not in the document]
stuggi: yes, need to be updated when the depending PR to lib-common merged.

README.md
stuggi: right, will do that

api/v1beta1/glanceapi_types.go
stuggi: right, its probably a good idea to make the Endpoint required, the `IPAddressPool` is already. Initially I was thinking we make the `Endpoint` optional to be able to expose the api via a service to any network, but only register the endpoint on the net where `Endpoint` is specified. would that use case make sense? I am not sure right now if it is needed

fmount: Oh I see your point: I'm not sure either atm, it would be nice having the ability to register the endpoint on the net where the `Endpoint` is specified, but we can take this as a follow up in case it will be a pattern followed by the other operators.

go.sum
```diff
@@ -294,6 +294,8 @@ github.com/stretchr/testify v1.7.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/
 github.com/stretchr/testify v1.8.0/go.mod h1:yNjHg4UonilssWZ8iaSj1OCr/vHnekPRkoO+kdMU+MU=
 github.com/stretchr/testify v1.8.1 h1:w7B6lhMri9wdJUVmEZPGGhZzrYTPvgJArz7wNPgYKsk=
 github.com/stretchr/testify v1.8.1/go.mod h1:w2LPCIKwWwSfY2zedu0+kehJoqGctiVI29o6fzry7u4=
+github.com/stuggi/lib-common/modules/common v0.0.0-20230213111010-7458e283a386 h1:ETecHuZuuw8JcV8ovdnygpi+RrW9UX7o2fymH9pUi4w=
```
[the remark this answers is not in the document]
stuggi: correct, I hope we get that landed today and I can remove the replace to my github

stuggi:
@fmount updated now and good to go

api/v1beta1/glanceapi_types.go
```diff
@@ -107,6 +107,43 @@ type GlanceAPISpec struct {
...
+	// NetworkAttachments is a list of NetworkAttachment resource names to expose the services to the given network
+	NetworkAttachments []string `json:"networkAttachments"`
+
+	// +kubebuilder:validation:Optional
+	// ExternalEndpoints, expose a VIP via MetalLB on the pre-created address pool
+	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints"`
+}
+
+// MetalLBConfig to configure the MetalLB loadbalancer service
+type MetalLBConfig struct {
```
fmount, line 121: Is this something we're going to replicate on all the operators that require the MetalLB info [1]? It's technically ok but I'm wondering if we want to have this struct as part of `lib-common` instead of re-implementing for each operator. Also, I think this struct is not going to change over the time (or diverge according to specific operator's needs), so I'm proposing to have a common definition.

[1] https://github.com/openstack-k8s-operators/cinder-operator/pull/136

stuggi: we plan to move this into the infra-operator when we add there more network related stuff, like ipam, dns and so to have it versioned there within the api. until then we keep it in the operators

fmount: ack and thanks for clarifying, I agree to get this information from `infra-operator` as more (common) network related components are going to be defined there.

fmount (approved):

controllers/glance_controller.go
```diff
@@ -396,6 +400,9 @@ func (r *GlanceReconciler) reconcileInit(
 	}
 	instance.Status.Conditions.MarkTrue(condition.DBSyncReadyCondition, condition.DBSyncReadyMessage)
 
+	// when job passed, mark NetworkAttachmentsReadyCondition ready
+	instance.Status.Conditions.MarkTrue(condition.NetworkAttachmentsReadyCondition, condition.NetworkAttachmentsReadyMessage)
+
```
abays, line 405: I'm confused as to why `NetworkAttachmentsReadyCondition` is set to `True` here and then is potentially set to something else between L495-515.  Should we maybe be doing something like you've proposed in the Keystone changes [1]?

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/2266fea2a2b862bf77f2aa139b19b94cd228007a/controllers/keystoneapi_controller.go#L634-L652

abays: Nevermind, GitHub code collapsing hid the fact that this is a function called later :)

abays (approved):
/lgtm

stuggi:
had to rebase

fmount (approved):

PR #153, opened 2023-02-14
Remove reconciler getters

rabi:
Getters and setters are non-idomatic in go for public fields, unless you're changing them. Anyway, we're accessing public fields everywhere.

### DIFF
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -53,26 +53,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// GetClient -
-func (r *GlanceReconciler) GetClient() client.Client {
-	return r.Client
-}
-
-// GetKClient -
-func (r *GlanceReconciler) GetKClient() kubernetes.Interface {
-	return r.Kclient
-}
-
-// GetLogger -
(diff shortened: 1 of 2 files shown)


api/bases/glance.openstack.org_glanceapis.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.10.0
+    controller-gen.kubebuilder.io/version: v0.9.2
```
stuggi, line 6: you remove your local bin/controller-gen to get the new version installed on next run

stuggi (commented):
looks good. those getters are left overs from the old ReconcilerCommon we had before the helper. just update your controller-gen to not flip this version

api/bases/glance.openstack.org_glanceapis.yaml
rabi: > you remove your local bin/controller-gen to get the new version installed on next run

Phew, I had proposed a change to cleanup old binaries elsewhere[1] , but you know I've to manually ensure that:/ 

[1] https://github.com/openstack-k8s-operators/keystone-operator/pull/179

PR #155, opened 2023-02-16
Update pre-commit hook with bashate

amolkahat:
This patch removes golint and by default
use golangci-lint.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,17 @@
 repos:
+- repo: https://github.com/dnephin/pre-commit-golang
+  rev: v0.5.1
+  hooks:
+    - id: go-fmt
+      exclude: ^vendor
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -51,3 +55,9 @@ repos:
     - id: no-commit-to-branch
     - id: trailing-whitespace
       exclude: ^vendor
+
+- repo: https://github.com/openstack-dev/bashate.git
```
abays, line 59: I noticed this was changed to `openstack` instead of `openstack-dev` in the similar Cinder PR.  Should it be changed here as well?

amolkahat: Yes. you are right.. I've missed it. Changing. Thanks!

abays (approved):
/lgtm

PR #157, opened 2023-02-16
Add build_root image configuration for prow

bshewale:


dprince:
Looks like we have 2 PRs doing the same thing. As https://github.com/openstack-k8s-operators/glance-operator/pull/156 was first I approved it.

PR #168, opened 2023-02-21
Add omitempty tag for slices

rabi:
Otherwise marshalling/unmarshalling CRDs with webhooks results in error as it serializes these struct fields as "null".

`The KeystoneAPI "keystone" is invalid: spec.networkAttachmentDefinitions:
Invalid value: "null": spec.networkAttachmentDefinitions in body must be of
type array: "null"`


abays (approved):
/lgtm

PR #170, opened 2023-02-21
Set default images via webhooks

abays:
Uses env vars set in CSV to provide defaults for container images in `Glance` and `GlanceAPI` CRs.  One possible approach using webhooks (an alternate non-webhook approach is proposed in #172).


api/v1beta1/glance_webhook.go
```diff
@@ -23,19 +23,56 @@ import (
...
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type
+func (r *Glance) Default() {
+	glancelog.Info("default", "name", r.Name)
+
+	r.Spec.Default(glanceDefaults)
+}
+
+// Default - set defaults for this Glance spec
+func (spec *GlanceSpec) Default(defaults GlanceDefaults) {
```
abays, line 57: I'm ambivalent on whether having the `defaults GlanceDefaults` arg here is worthwhile or not, but I just thought I'd include it initially to see what people thought

hjensas (approved):
This looks fine to me.

api/v1beta1/glance_webhook.go
```diff
@@ -23,19 +23,51 @@ import (
...
+}
+
+var glanceDefaults GlanceDefaults
+
 // log is for logging in this package.
 var glancelog = logf.Log.WithName("glance-resource")
 
 // SetupWebhookWithManager sets up the webhook with the Manager
-func (r *Glance) SetupWebhookWithManager(mgr ctrl.Manager) error {
+func (r *Glance) SetupWebhookWithManager(mgr ctrl.Manager, defaults GlanceDefaults) error {
```
gibizer, line 37: So when we move the defaulting to run from the openstack-operator we need to make sure that the defaults are provided to `func (spec *GlanceSpec) Default()` in some way. In that case the SetupWebhookWithManager call well have  OpenStackOperator receiver. So maybe we need to change the signature to `func (spec *GlanceSpec) Default(defaults)` to be able to pass in some defaults from openstack-operator to the glance defaulter.

abays: I actually had it that way initially.  But then I realized that the env vars are there in the Glance deployment within the OpenStack operator CSV, so the env vars will be acquired and passed to the webhooks when Glance's `main.go` is executed...unless we have the OpenStack operator set `ENABLE_WEBHOOKS=false` on the sub-deployments, which I know we are considering.

gibizer: Yeah I think you are correct. So the only "problematic" case is when the webhook is only enabled in the openstack-operator but not in the glance-operator

main.go
```diff
@@ -120,11 +120,25 @@ func main() {
 		os.Exit(1)
 	}
 	if strings.ToLower(os.Getenv("ENABLE_WEBHOOKS")) != "false" {
-		if err = (&glancev1.Glance{}).SetupWebhookWithManager(mgr); err != nil {
+		glanceDefaults := glancev1.GlanceDefaults{
```
fmount, line 123: @abays quick question here: regardless of `ENABLE_WEBHOOKS`, we're always running on L127 `SetupWebhookWithManager`, is this the expected behaviour? Not sure I'm misunderstanding the purpose of `ENABLE_WEBHOOKS` variable.

api/v1beta1/glance_webhook.go
```diff
@@ -23,19 +23,51 @@ import (
...
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type
+func (r *Glance) Default() {
+	glancelog.Info("default", "name", r.Name)
+
+	r.Spec.Default()
+}
+
+// Default - set defaults for this Glance spec
+func (spec *GlanceSpec) Default() {
```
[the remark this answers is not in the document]
fmount: @gibizer this is reusable because it's part of the `api/` and  can be imported in the `openstack-operator`, right? 
By doing this we can enforce defaults from  `meta-operator` (populating `GlanceDefaults`, which is passed to `SetupWebhookWithManager`), then, when the service operator is executed and the webhook is started, we can process those default values via this function (where the logic resides).
Is that the right understanding of the potential workflow?

fmount (commented):
I still have to try and get more familiar with webhooks (I left some comments and questions, mostly caused by my lack of knowledge), but in general the patch looks good to me. Still waiting a bit to see if there are additional feedback on the approach by other folks.

api/v1beta1/glance_webhook.go
rabi: Just wondering why we want to use env variables for default images and change the webhook setup? As I understand the only things that can change for images are [1] and those can be inputs (like storageClass[2]) and there can be defaults for those (ex. tag: latest and namespace: quay.io/tripleomaster or openshift internal registry namespace). The defaulting logic here or the reconciler code could use those to populate containerImage field.

[1]
namespace: quay.io/tripleozedcentos9 (or what is used by the container build pipeline)
name_prefix: openstack-
tag: current-tripleo

ex. quay.io/tripleozedcentos9/openstack-<mariadb/service_name>:current-tripleo

[2] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/apis/core/v1beta1/openstackcontrolplane_types.go#L47

hjensas: > Just wondering why we want to use env variables for default images and change the webhook setup? As I understand the only things that can change for images are [1] and those can be inputs (like storageClass[2]) and there can be defaults for those (ex. tag: latest and namespace: quay.io/tripleomaster or openshift internal registry namespace). The defaulting logic here or the reconciler code could use those to populate containerImage field.
> 
> [1] namespace: quay.io/tripleozedcentos9 (or what is used by the container build pipeline) name_prefix: openstack- tag: current-tripleo
> 
> ex. quay.io/tripleozedcentos9/openstack-<mariadb/service_name>:current-tripleo
> 
> [2] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/apis/core/v1beta1/openstackcontrolplane_types.go#L47

This seems like something to think about.
Is there more background somewhere on they why for this change?

abays: @dprince Could you provide the background here as to why we're repeating this pattern that we also used in OSPdO [1]?

https://github.com/openstack-k8s-operators/osp-director-operator/blob/8006828eaa2c4eff79975b4afc2ee3fdf7936471/main.go#L328-L355

api/v1beta1/glance_webhook.go
```diff
@@ -23,19 +23,51 @@ import (
...
+		spec.ContainerImage = glanceDefaults.ContainerImageURL
+	}
+
+	if spec.GlanceAPIExternal.ContainerImage == "" {
+		spec.GlanceAPIExternal.ContainerImage = glanceDefaults.ContainerImageURL
+	}
+
+	if spec.GlanceAPIInternal.ContainerImage == "" {
+		spec.GlanceAPIInternal.ContainerImage = glanceDefaults.ContainerImageURL
+	}
```
gibizer, line 68: Below @fmount, you ask about supporting different `ContainerImage` defaults for `GlanceAPIExternal` and `GlanceAPIInternal`. I think this code here needs to be changed to read different defaults from `glanceDefaults` to support that use case.

However if we only need to support manually defining different `ContainerImage` but the default is OK to be the same always. Then this code is OK as is.

api/v1beta1/glanceapi_webhook.go
```diff
@@ -0,0 +1,89 @@
...
+func (r *GlanceAPI) Default() {
+	glanceapilog.Info("default", "name", r.Name)
+
+	r.Spec.Default()
+}
+
+// Default - set defaults for this GlanceAPI spec
+func (spec *GlanceAPISpec) Default() {
+	if spec.ContainerImage == "" {
+		spec.ContainerImage = glanceAPIDefaults.ContainerImageURL
```
gibizer, line 59: This feels redundant now. The `func (spec *GlanceSpec) Default()` on Glance level already set a value to this field if it was empty so this code will never be triggered. 

I see couple of possible solutions alternatives:

1. `func (spec *GlanceSpec) Default()` is changed to call `Glance.Spec.GlanceAPIExternal.Default()` and `Glance.Spec.GlanceAPIInternal.Default()` respectively. In this case the webhook registration is not needed in GlanceAPI at all. This also means that openstack-operator only need to call `func (spec *GlanceSpec) Default()`  and the rest is automatic. However the downside of this is that if GlanceAPI is ever instantiated independently from Glance struct then the defaulting on GlanceAPI will not be triggered.

2. `func (spec *GlanceSpec) Default()` can simply not default anything on any sub Spec struct and let the webhook in GlanceAPI level do it.  This means that openstack-operator needs to call `func (spec *GlanceSpec) Default()`  and `func (spec *GlanceAPISpec) Default()` as well.

fmount: This is a key point of this implementation: I think we don't have to expose at `openstack-operator` level the subCRs specs, and from that layer we should call `func (spec *GlanceSpec) Default()`.
It's the service implementation that then triggers  `Glance.Spec.GlanceAPIExternal.Default()` and `Glance.Spec.GlanceAPIInternal.Default()` to enforce defaults (it's the same in case of `ContainerImage` but it could be different for other parameters).
So I think option 1. could be a good choice in a top-down view, but I see there's a problem on the `GlanceAPI` trigger.
@abays wdyt about the two options?

api/v1beta1/glance_webhook.go
fmount: In this case `ContainerImage` is the same parameter that should be enforced to both `Internal` and `External`, so no big concerns. We can try refactoring the defaults in case we need to diverge a bit in the future.

main.go
fmount: ack, I thought that not having `if err = (&glancev1.Glance{}).SetupWebhookWithManager(mgr); err != nil {` on L123 would have enabled the webhook anyway (because the `if strings.ToLower(os.Getenv("ENABLE_WEBHOOKS")) != "false" {` is closed on L125.

main.go
```diff
@@ -120,11 +120,25 @@ func main() {
...
+			ContainerImageURL: os.Getenv("GLANCE_API_IMAGE_URL_DEFAULT"),
+		}
+
+		if err = (&glancev1.Glance{}).SetupWebhookWithManager(mgr, glanceDefaults); err != nil {
 			setupLog.Error(err, "unable to create webhook", "webhook", "Glance")
 			os.Exit(1)
 		}
+
+		glanceAPIDefaults := glancev1.GlanceAPIDefaults{
```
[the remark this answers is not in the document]
fmount: Discussed on a diff thread w/ @gibizer

main.go
abays: I actually have a different version of this section that I forgot to push last Thursday.  Since we seem to be somewhat ambivalent on whether webhooks will only exist in OpenStack Operator or if they will also be enabled in the service operators, I have moved the setup on the defaults out of the webhook setup.  i.e. it looks something like this:

```
        // Acquire environmental defaults and initialize Glance and GlanceAPI defaults with them
	glanceDefaults := glancev1.GlanceDefaults{
		ContainerImageURL: os.Getenv("GLANCE_API_IMAGE_URL_DEFAULT"),
	}

	glanceAPIDefaults := glancev1.GlanceAPIDefaults{
		ContainerImageURL: os.Getenv("GLANCE_API_IMAGE_URL_DEFAULT"),
	}

	(&glancev1.Glance{}).Spec.SetupDefaults(glanceDefaults)
	(&glancev1.GlanceAPI{}).Spec.SetupDefaults(glanceAPIDefaults)

	// Setup webhooks if requested
	if strings.ToLower(os.Getenv("ENABLE_WEBHOOKS")) != "false" {
		if err = (&glancev1.Glance{}).SetupWebhookWithManager(mgr); err != nil {
	...
	etc
```

That way the defaults are initialized regardless of whether the service operator has webhooks enabled or not, and thus the OpenStack Operator could still properly call `Glance.Spec.Default()` in the "no service operator webhooks" context.  `SetupDefaults()` would just be placed somewhere in the `api` module like so:

```
// SetupDefaults - initialize Glance spec defaults for use with either internal or external webhooks
func (r *GlanceSpec) SetupDefaults(defaults GlanceDefaults) {
	glanceDefaults = defaults
}
```

api/v1beta1/glanceapi_webhook.go
abays: > However the downside of this is that if GlanceAPI is ever instantiated independently from Glance struct then the defaulting on GlanceAPI will not be triggered.

That is exactly why I had initially included the redundant `GlanceAPI` defaults.  I don't know where we are right now on whether we'll support individual, direct sub-resource creation as a valid user workflow.  I think we should come to a consensus in that regard and then that decision will inform the proper approach to these defaulting patterns.

main.go
fmount: +1 on having the defaults initialized out of the webhook setup, and your snippet makes sense to me as it's enabled according to he `ENABLE_WEBHOOKS` parameter.
I'm assuming your're going to push the last version of it (unless other conversations are still ongoing)

abays: Yes, I'll push the changes now

api/v1beta1/glanceapi_webhook.go
gibizer: Let's try to discuss this tomorrow on the sync meeting.

api/v1beta1/glance_webhook.go
abays: @hjensas So the context around why we would want to use env vars to provide defaults for service container images involves CSV, `relatedImages` and offline/air-gapped installations.  See the `IMPORTANT` block here: https://docs.openshift.com/container-platform/4.12/operators/admin/olm-restricted-networks.html

api/v1beta1/glanceapi_webhook.go
abays: We discussed the sub-resource CR concern today, and I believe the consensus is that we won't support creating/applying them.  That is, in the context of this particular operator, only `Glance` CRs should ever be created.  So I think we can remove the `GlanceAPI` defaulting logic.  @gibizer was that your understanding from today's conversation as well?

gibizer: Yeah, so we only add the Defaulting interface to the Glance level but never on GlanceAPI level. And the Glance level defaulter will do defaulting on the GlanceAPISpec stuct as well.

fmount: +1 on that

fmount:
@abays I see you removed the `GlanceAPI` logic in the last change. The patch looks good, I guess we can merge it as long as CI is green (I think the failure is unrelated)

abays:
> @abays I see you removed the `GlanceAPI` logic in the last change. The patch looks good, I guess we can merge it as long as CI is green (I think the failure is unrelated)

Yeah, the CI issue might be a hiccup.  Will retest now.

PR #172, opened 2023-02-22
Set default images during reconcile

abays:
Uses env vars set in CSV to provide defaults for container images in `Glance` and `GlanceAPI` CRs.  One possible approach using defaulting functions in the reconcile loop, instead of using webhooks (webhooks option is proposed in #170).

This method avoids the use of webhooks and instead follows the pattern of prior art for feeding env vars into the controller manager [1].

Note that this approach will require changes to `lib-common` [2] to patch instance `spec`.

[1] https://github.com/openstack-k8s-operators/openstack-operator/commit/d72cf7158a866153b66ec2125e3d1012a04ab135
[2] https://github.com/openstack-k8s-operators/lib-common/blob/4d837b4ac7d2620f4ff019e08777f4a57d6a02b4/modules/common/helper/helper.go#L203-L227

### DIFF
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -54,9 +54,10 @@ import (
 // GlanceAPIReconciler reconciles a GlanceAPI object
 type GlanceAPIReconciler struct {
 	client.Client
-	Kclient kubernetes.Interface
-	Log     logr.Logger
-	Scheme  *runtime.Scheme
+	Kclient           kubernetes.Interface
+	Log               logr.Logger
+	Scheme            *runtime.Scheme
+	ContainerImageURL string
 }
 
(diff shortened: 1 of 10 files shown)


controllers/glanceapi_controller.go
```diff
@@ -616,3 +622,12 @@ func (r *GlanceAPIReconciler) createHashOfInputHashes(
 	}
 	return hash, changed, nil
 }
+
+func (r *GlanceAPIReconciler) applyDefaults(instance *glancev1.GlanceAPI) bool {
```
fmount, line 626: Looks like we're going to call this function within the `Reconcile`, actively trying to enforce the `ContainerImageURL`, while using webhooks we just react on a change to the CR, which is more efficient.

abays:
We have at a plurality agreement against this form of defaulting and will be using webhooks in one way or another instead

PR #176, opened 2023-03-14
Remove description from Glance CRD

fmount:
This patch rebuilds the Glance `CRD` removing the description to avoid a potential issue when the Glance bundle is deployed via `OLM` by the meta operator and the following is seen:

```
openstack/install-7ws9s"} failed: failed to update installplan bundle lookups: etcdserver: request is too large
```

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -13,4 +13,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: quay.io/openstack-k8s-operators/glance-operator
-  newTag: v0.0.1
+  newTag: latest
--- a/Makefile
+++ b/Makefile
@@ -232,6 +232,7 @@ $(ENVTEST): $(LOCALBIN)
 	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
 .PHONY: bundle
+bundle: CRDDESC_OVERRIDE=:maxDescLen=0
 bundle: manifests kustomize ## Generate bundle manifests and metadata, then validate generated files.
 	operator-sdk generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
--- a/api/bases/glance.openstack.org_glanceapis.yaml
+++ b/api/bases/glance.openstack.org_glanceapis.yaml
@@ -31,22 +31,14 @@ spec:
     name: v1beta1
     schema:
       openAPIV3Schema:
-        description: GlanceAPI is the Schema for the glanceapis API
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation
-              of an object. Servers should convert recognized schemas to the latest
-              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
             type: string
           kind:
-            description: 'Kind is a string value representing the REST resource this
-              object represents. Servers may infer this from the endpoint the client
-              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
             type: string
           metadata:
             type: object
           spec:
-            description: GlanceAPISpec defines the desired state of GlanceAPI
             properties:
               apiType:
                 default: external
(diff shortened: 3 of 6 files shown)


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openstack-k8s-operators/glance-operator
-  newTag: v0.0.1
+  newName: quay.io/openstack-k8s-operators/glance-operator-index
```
abays, line 15: This should actually point to the operator image itself, not the index.  But the tag change is correct.

fmount: sorry my bad on that one, removing the `-index` while bumping the tag

fmount:
@abays @stuggi I'm wondering if  we want to enfore `CRDDESC=0` even in `make manifests`

abays:
> @abays @stuggi I'm wondering if we want to enforce `CRDDESC=0` even in `make manifests`

I'm guessing that removing these descriptions from the bundle and potentially even the CRD manifests will mostly only be noticeable when trying to create the CRs through the OpenShift web console.  The fields there would no longer have the descriptions, but I don't think we're supporting GUI creation of resources anyhow.  Of course, the `oc get crd <glance CRD>` would not have them either.  But I think we need to do this to avoid the problem mentioned in the PR description.

fmount:
> > @abays @stuggi I'm wondering if we want to enforce `CRDDESC=0` even in `make manifests`
> 
> I'm guessing that removing these descriptions from the bundle and potentially even the CRD manifests will mostly only be noticeable when trying to create the CRs through the OpenShift web console. The fields there would no longer have the descriptions, but I don't think we're supporting GUI creation of resources anyhow. Of course, the `oc get crd <glance CRD>` would not have them either. But I think we need to do this to avoid the problem mentioned in the PR description.

Yeah, I agree. I would reduce the scope to `oc get crd <glance CRD>` because in theory creating it with the meta-operator we're not supposed to expose the sub-resources.

stuggi:
> > > @abays @stuggi I'm wondering if we want to enforce `CRDDESC=0` even in `make manifests`
> > 
> > 
> > I'm guessing that removing these descriptions from the bundle and potentially even the CRD manifests will mostly only be noticeable when trying to create the CRs through the OpenShift web console. The fields there would no longer have the descriptions, but I don't think we're supporting GUI creation of resources anyhow. Of course, the `oc get crd <glance CRD>` would not have them either. But I think we need to do this to avoid the problem mentioned in the PR description.
> 
> Yeah, I agree. I would reduce the scope to `oc get crd <glance CRD>` because in theory creating it with the meta-operator we're not supposed to expose the sub-resources.

If possible we should  keep them in the CRD the user is interacting with (ctlplane) that it is not required to check in a doc and not being able to check the meaning of a param from the system.
Also the question is still if the total size ~4MB uncompressed or a single manifest exceeds ~1MB, I guess its the total.

Another point to keep in mind, when we have newer api version in any of the component/s, at least for an intermediate time where we have multiple versions, it will double the size for that component/s.

fmount:
> > > > @abays @stuggi I'm wondering if we want to enforce `CRDDESC=0` even in `make manifests`
> > > 
> > > 
> > > I'm guessing that removing these descriptions from the bundle and potentially even the CRD manifests will mostly only be noticeable when trying to create the CRs through the OpenShift web console. The fields there would no longer have the descriptions, but I don't think we're supporting GUI creation of resources anyhow. Of course, the `oc get crd <glance CRD>` would not have them either. But I think we need to do this to avoid the problem mentioned in the PR description.
> > 
> > 
> > Yeah, I agree. I would reduce the scope to `oc get crd <glance CRD>` because in theory creating it with the meta-operator we're not supposed to expose the sub-resources.
> 
> If possible we should keep them in the CRD the user is interacting with (ctlplane) that it is not required to check in a doc and not being able to check the meaning of a param from the system. Also the question is still if the total size ~4MB uncompressed or a single manifest exceeds ~1MB, I guess its the total.
> 
> Another point to keep in mind, when we have newer api version in any of the component/s, at least for an intermediate time where we have multiple versions, it will double the size for that component/s.

Ack, makes sense. The `request too large` problem I guess is seen only when everything is deployed by the meta-operator [1],
if I deploy the single operator (e.g. through `install_yamls`) I don't see this kind of problem: the question is if we want to address it with a distributed series of patches across the operators or we should find a different strategy at the meta-operator level.

[1] https://paste.opendev.org/show/b9UFDOkEkucoHWDLTKcn/

Akrog (changes requested):
This looks ok since it's in `install` and `deploy` targets as well.

Akrog (approved):

PR #177, opened 2023-03-14
Makefile: Fix renovate PRs

fmount:
We previously added the "CRDDESC_OVERRIDE=:maxDescLen=0" environmental variable to "Makefile" targets "install", "deploy", and "bundle" to remove the description from our CRDs and avoid error:

'''
openstack/install-7ws9s"} failed: failed to update installplan bundle lookups: etcdserver: request is too large
'''

These previous patches fixed a number of cases, but not the PRs generated by the Renovate Bot, which are using command "make manifests generate", as can be seen in the repository's "renovate.json".

This means that when the bot proposes a new PR to update a dependency it will include the description, which is something we don't want.

To prevent this from happening this patch changes the approach to setting the "maxDescLen" parameter as a global variable instead of setting it in the above mentioned targets.

This value can be overriden by any caller by setting the environmental variable on the call:

  CRDDESC_OVERRIDE= make manifests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,8 @@ ifneq ($(origin CHANNELS), undefined)
 BUNDLE_CHANNELS := --channels=$(CHANNELS)
 endif
 
+CRDDESC_OVERRIDE ?= :maxDescLen=0
+
 # DEFAULT_CHANNEL defines the default channel used in the bundle.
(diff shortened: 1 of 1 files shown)


abays (approved):
/lgtm

PR #179, opened 2023-03-16
Trim GlanceCRD and optimize extraMounts struct usage

fmount:
Due to the bundle size limit we recently hit in CI we had to take some decisions in the way the CRD is build and the API are exposed in the sub CR.
With this patch we only allow the ExtraMounts definition in the top-level CRD: it will be assigned in the controller to the underlying resources that are still able to properly process their volumes thanks to the internal propagation feature.


api/v1beta1/glanceapi_types.go
```diff
@@ -182,6 +178,7 @@ type GlanceAPI struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	Spec   GlanceAPISpec   `json:"spec,omitempty"`
+	ExtraMounts []GlanceExtraVolMounts `json:"extraMounts,omitempty"`
```
fmount, line 181: This is actually something we can avoid here: I'm not sure there's a way to pass this param to the `GlanceAPI` deployment, but if we are able to pass it down from the top-level crd, we could reduce a lot the CRD size.

dprince: I'm game to try this. It is certainly more user friendly this way in that all the data is collocated together

fmount: @abays  @stuggi  in my tests I tried changing this line to:

```
ExtraMounts []GlanceExtraVolMounts `json:"-"`
``` 
which results in an "hidden" field because it's not generated in the CRD, and it works fine in the `glance_controller`.
However, the problem is that even though I set the variable at some point in the glance-controller code:

`deployment.ExtraMounts = instance.Spec.ExtraMounts`

the underlying resource (glanceapi instance) which is fetched in [1] doesn't have the parameter, because I think when the request is sent through the API there's no field (due to `json: "-"`): this basically means that the parameter cant' survive this assignment, it's valid in the scope of a single controller.
I'm wondering if you know if a way (or a strategy) of passing parameter from the `top-level` instance to the sub-cr exists w/o exposing the parameter in the API.

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/master/controllers/glanceapi_controller.go#L80

gibizer: You are correct about the "-" behavior it basically tells the json serializer to ignore the field.  https://pkg.go.dev/encoding/json#Marshal

fmount:
@abays @dprince the branch [1] is interesting because we can reach the max optimization on the glance CR.

we started with:

```
> du -sh ./manifests/glance*
48K     ./manifests/glance.openstack.org_glanceapis.yaml
136K    ./manifests/glance.openstack.org_glances.yaml
``` 

Then with this patch we reach:

```
> du -sh config/crd/bases/glance.openstack.org_glance*
44K     config/crd/bases/glance.openstack.org_glanceapis.yaml
52K     config/crd/bases/glance.openstack.org_glances.yaml
```

and with the branch [1] (where I still need to review the code and make it more solid), I can basically do the `json: "-"` I mentioned in the previous comment and fetch the `extraVolumes` from the top-level `CR` instead of injecting (I can't actually because I don't serialize the param which at this point is not exposed through the API of the underlying CR):

```
> du -sh config/crd/bases/glance.openstack.org_glance*
8.0K    config/crd/bases/glance.openstack.org_glanceapis.yaml
52K     config/crd/bases/glance.openstack.org_glances.yaml
```

[1] https://github.com/openstack-k8s-operators/glance-operator/compare/master...fmount:glance-operator:more_crd_optimization

api/v1beta1/glanceapi_types.go
abays: @fmount One option is to get the parent `Glance` resource within the `GlanceAPI` reconcile loop and then directly copy the `ExtraMounts` field from that object.  We have a similar concept in the Cinder operator where any of the sub-resource CRs (`CinderAPI`, for instance) look for the parent `Cinder` CR [1].  In that particular case we're trying to ultimately get a `ConfigMap` created by the `Cinder` controller for that parent `Cinder` CR [2], but the "spirit" of the approach is the same -- and, in the case of Glance, it's even simpler since we only need to get the parent `Glance` CR and don't need to look further.

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/cf28b9aa5fd300715ec9b6128ff36a7dfe257040/controllers/cinderapi_controller.go#L503-L529

fmount: @abays cool, thanks for confirming it: I have something similar here [1]  and I wanted to check with you if I was doing something crazy; however, it looks like I'm adopting the approach you just described (which it's easy enough for Glance).


[1] https://github.com/openstack-k8s-operators/glance-operator/compare/master...fmount:glance-operator:more_crd_optimization

fmount: @gibizer good point: not sure if strategies exist to get notified is something different from the "Spec" changes. After the greenfield deployment, if I `oc edit glance glance` and I change anything in the `ExtraMounts`, I'm able to trigger the reconcile loop in the top-level CR, but I'm not able to trigger a `reconcileLoop` for the `GlanceAPI` resources that should update both `internal` and `external` deployments.
@abays not sure a way exist to handle this case.

fmount: Closing this PR as much of the conversation has been moved in [1] where the mechanism described by @gibizer has been implemented.

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/184

PR #181, opened 2023-03-20
Trim GlanceCRD and optimize extraMounts struct usage

fmount:
Due to the bundle size limit we recently hit in CI we had to take some decisions in the way the CRD is build and the API are exposed in the sub CR.
With this patch we only allow the ExtraMounts definition in the top-level CRD: it will be assigned in the controller to the underlying resources that are still able to properly process their volumes thanks to the internal propagation feature.


controllers/glanceapi_controller.go
```diff
@@ -120,6 +120,27 @@ func (r *GlanceAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	if err := r.Client.List(context.Background(), glance, listOpts...); err != nil {
+		r.Log.Error(err, "Unable to retrieve Glance CR %v")
+		return ctrl.Result{}, nil
+	}
+	for _, cr := range glance.Items {
+		// Filter the Glance CR by kind and name
+		if cr.Kind == "Glance" && cr.Name == parentGlance {
+			r.Log.Info(fmt.Sprintf("Glance Top Level CR found %s", cr.Name))
+			r.Log.Info(fmt.Sprintf("ExtraVolumes Defined in the Top Level CR %v", cr.Spec.ExtraMounts))
+			instance.ExtraMounts = cr.Spec.ExtraMounts
```
fmount, line 140: @abays so basically here I'm doing something similar to what `Cinder` does: I watch/get the top-level (parent) CR and I fetch the `ExtraMounts` (if not defined, `instance.ExtraMounts` is not assigned).

abays: I'm curious as to why we would get a list here as opposed to just getting the `Glance` CR directly, since we have its name and namespace.  i.e. why not do something like this [1] instead (old example from OSPdO but demonstrates the approach nonetheless)? 

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/3d651137b62cfe7ff937c8ed5d04638f3194282a/api/v1beta1/common_openstacknet.go#L213-L227

fmount: I was thinking more at this use case, you can technically have more than one `Glance` CR deployed, but the relationship with the sub-resources is established in the controller, so I don't have to list all the `Glance` related `CRs` and watch for the right one to fetch `data`: I basically created that part starting from a wrong assumption.
Thanks actually for pointing me to [1], it's helpful and I'm pushing the same approach described there.

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/3d651137b62cfe7ff937c8ed5d04638f3194282a/api/v1beta1/common_openstacknet.go#L213-L227

controllers/glanceapi_controller.go
```diff
@@ -120,6 +121,27 @@ func (r *GlanceAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+
+	// we don't need to fail here (or return an err) because it might happen
+	// that the Top Level resource is deleted but the underlying resources are
+	// not yet cleaned up: we need to wait for r.reconcileDelete being executed
+	// if the top-level resource (parentGlance) is empty, an empty array is
+	// assigned to instance.ExtraMounts
+	if err != nil {
+		r.Log.Info("Reconciling Service: Glance CR not present")
+	}
+	instance.ExtraMounts = parentGlance.Spec.ExtraMounts
```
abays, line 143: Should we check if the `err` is something other than `Not Found` and return immediately, just to be consistent with how we handle errors across the various operators?

fmount: Done

dprince (approved):

PR #183, opened 2023-03-21
Revert "Trim GlanceCRD and optimize extraMounts struct usage"

fmount:
Reverts openstack-k8s-operators/glance-operator#181

We still don't have a way to notify the update to the sub-resources if something changes in the top-level CR (we can't trigger the reconcile on the sub-resource if something different than "Spec" changes.
This change will be re-proposed if a clean way will be found.


abays:
I think one idea might be to store a hash of the `ExtraMount` in the `Glance.Status`.  Then the `GlanceAPI` watches `Glance` and its reconcile loop will be trigger in the event that `Glance.ExtraMount` changes (because the `Glance` controller will update the `Glance` status with the new hash).  At least...I think when a controller watches a resource, it will be triggered by status updates and not just for spec updates.

abays:
> I think one idea might be to store a hash of the `ExtraMount` in the `Glance.Status`. Then the `GlanceAPI` watches `Glance` and its reconcile loop will be triggered in the event that `Glance.ExtraMount` changes (because the `Glance` controller will update the `Glance` status with the new hash). At least...I think when a controller watches a resource, it will be triggered by status updates and not just for spec updates.

Or is the problem that the `Glance` controller doesn't wake when the `Glance` resource's `ExtraMounts` is updated?

fmount:
> > I think one idea might be to store a hash of the `ExtraMount` in the `Glance.Status`. Then the `GlanceAPI` watches `Glance` and its reconcile loop will be triggered in the event that `Glance.ExtraMount` changes (because the `Glance` controller will update the `Glance` status with the new hash). At least...I think when a controller watches a resource, it will be triggered by status updates and not just for spec updates.
> 
> Or is the problem that the `Glance` controller doesn't wake when the `Glance` resource's `ExtraMounts` is updated?

So the `Glance controller` reconcile loop is triggered when `ExtraMount(s)` change, the problem I see here is being notified in the `glanceAPI controller` and trigger the reconcile loop due to this event.
I already added the hash with something like:

```
+       var ev []storage.VolMounts
+       for _, vm := range instance.Spec.ExtraMounts {
+               ev = append(ev, vm.VolMounts...)
+       }
+       h, err := storage.Hash(ev)
+       configMapVars["ExtraMounts"] = env.SetValue(h)
```
and I see this `hash` being update if I stop my local operator and `make run` it again: actually stopping and starting triggers the GlanceAPI controller and I'm able to update both `internal` and `external` deployments.
I'm trying to solve the problem of being notified by playing a bit w/ `SetupWithManager` in the `glanceapi_controller`:

```
+       extraVols := func(o client.Object) []reconcile.Request {
+               result := []reconcile.Request{}
+               // Get Glance Top level CR and fetch ExtraVolumes if defined
+               glances := &glancev1.GlanceList{}
+               listOpts := []client.ListOption{
+                       client.InNamespace(o.GetNamespace()),
+               }
+               if err := r.Client.List(context.Background(), glances, listOpts...); err != nil {
+                       r.Log.Error(err, "Unable to retrieve API CRs %v")
+                       return nil
+               }
+
+               for _, cr := range glances.Items {
+                       if cr.Kind == "Glance" {
+                               name := client.ObjectKey{
+                                       Namespace: o.GetNamespace(),
+                                       Name:      cr.Name,
+                               }
+                               result = append(result, reconcile.Request{NamespacedName: name})
+                       }
+               }
+               if len(result) > 0 {
+                       r.Log.Info("Reconciling Request!")
+                       return result
+               }
+               return nil
+       }
        return ctrl.NewControllerManagedBy(mgr).
                For(&glancev1.GlanceAPI{}).
                Owns(&keystonev1.KeystoneEndpoint{}).
@@ -192,6 +210,9 @@ func (r *GlanceAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
                Owns(&corev1.ConfigMap{}).
                Owns(&appsv1.Deployment{}).
                Owns(&routev1.Route{}).
+               // Watch for Glance{} ExtraVolumes
+               Watches(&source.Kind{Type: &glancev1.Glance{}},
+                       handler.EnqueueRequestsFromMapFunc(extraVols)).
                Complete(r)
 }
```
with something like the above, when glance CR is modified, I should always trigger a reconcile loop (then the idea is to reduce the scope to the input hash that changes), but so far I only get:

```
2023-03-21T14:25:39+01:00       INFO    controllers.GlanceAPI   CR KIND Glance
2023-03-21T14:25:39+01:00       INFO    controllers.GlanceAPI   FOUND glance
2023-03-21T14:25:39+01:00       INFO    controllers.GlanceAPI   Reconciling Request
2023-03-21T14:25:39+01:00       INFO    controllers.GlanceAPI   CR KIND Glance
2023-03-21T14:25:39+01:00       INFO    controllers.GlanceAPI   FOUND glance
2023-03-21T14:25:39+01:00       INFO    controllers.GlanceAPI   Reconciling Request
2023-03-21T14:25:39+01:00       INFO    controllers.Glance      Reconciling Service 'glance'
2023-03-21T14:25:39+01:00       INFO    controllers.Glance      (TOP_LEVEL) EXTRAMOUNTS (CONFIGMAP): 0x16be600
2023-03-21T14:25:39+01:00       INFO    controllers.Glance      Reconciling Service 'glance' init
2023-03-21T14:25:39+01:00       INFO    controllers.Glance      Reconciled Service 'glance' init successfully
2023-03-21T14:25:39+01:00       INFO    controllers.Glance      Reconciling Service 'glance' update
2023-03-21T14:25:39+01:00       INFO    controllers.Glance      Reconciled Service 'glance' update successfully
2023-03-21T14:25:39+01:00       INFO    controllers.Glance      Reconciling Service 'glance' upgrade
2023-03-21T14:25:39+01:00       INFO    controllers.Glance      Reconciled Service 'glance' upgrade successfully
2023-03-21T14:25:39+01:00       INFO    controllers.Glance      Deployment glance successfully reconciled - operation: updated
2023-03-21T14:25:39+01:00       INFO    controllers.Glance      Deployment glance successfully reconciled - operation: updated
2023-03-21T14:25:39+01:00       INFO    controllers.Glance      Reconciled Service 'glance' successfully
```

I'm able to see the top-level CR being changed, I think I'm able to enqueue a reconciliation request, but I don't see it triggered in the underlying controller!

PR #184, opened 2023-03-21
Get notified in the glanceAPI controller

fmount:
This patch is a follow up of PR#181.
If the `Glance` top level `CR` changes (e.g. an `extraMount` is updated), we should watch it and `trigger the reconcile loop (which means we should be notified as long as this event occur).

### DIFF
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -27,7 +27,10 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/go-logr/logr"
 	routev1 "github.com/openshift/api/route/v1"
@@ -121,27 +124,6 @@ func (r *GlanceAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, nil
 	}
 
-	// Get Glance Top level CR and fetch ExtraVolumes if defined
-	parentGlance := &glancev1.Glance{}
-
-	err = r.Client.Get(
-		context.TODO(),
-		types.NamespacedName{
-			Name:      glance.GetOwningGlanceName(instance),
-			Namespace: req.Namespace,
-		},
-		parentGlance)
-
-	if err != nil {
-		if !k8s_errors.IsNotFound(err) {
-			return ctrl.Result{}, err
-		}
-		r.Log.Info("Reconciling Service: Glance CR not present")
-	}
-	// if the top-level resource (parentGlance) is empty, an empty array is
-	// assigned to instance.ExtraMounts
-	instance.ExtraMounts = parentGlance.Spec.ExtraMounts
-
 	//
 	// initialize status
 	//
@@ -184,6 +166,68 @@ func (r *GlanceAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *GlanceAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	glance := func(o client.Object) []reconcile.Request {
+
+		result := []reconcile.Request{}
+
+		// Get GlanceAPIList that should be updated
+		glanceapis := &glancev1.GlanceAPIList{}
+
+		listOpts := []client.ListOption{
+			client.InNamespace(o.GetNamespace()),
+		}
+		if err := r.Client.List(context.Background(), glanceapis, listOpts...); err != nil {
+			r.Log.Error(err, "Unable to retrieve GlanceAPI CRs %v")
+			return nil
+		}
+
+		// For each GlanceAPI instance (typically internal and external), we
+		// create a reconcile request if ExtraMounts changed in the parent CR
+		for _, api := range glanceapis.Items {
+
+			// Get parent Glance CR checking the GetOwnerReferences in the
+			// glance pkg
+			parentGlance := &glancev1.Glance{}
+			err := r.Client.Get(
+				context.TODO(),
+				types.NamespacedName{
+					Name:      glance.GetOwningGlanceName(&api),
+					Namespace: o.GetNamespace(),
+				},
+				parentGlance)
+
+			// No parent in the OwnerReferences available, we should return
+			// as no reconcile loop should be triggered
+			if err != nil {
+				return nil
+			}
+
+			// The new reconcile request should be enqueued as long as the
+			// ExtraMounts struct has changed in the top-level CR and the
+			// ownerRef matches with the client.Object
+			hashChanged, hash, err := r.extraMountsHashChanged(&api, parentGlance.Spec.ExtraMounts)
+			if err != nil {
+				r.Log.Error(err, "Failing to get instance Hash")
+				return nil
+			}
+
+			util.SetHash(api.Status.Hash, "extraMounts", hash)
+			if parentGlance.Name == o.GetName() && hashChanged {
+				name := client.ObjectKey{
+					Namespace: o.GetNamespace(),
+					Name:      api.Name,
+				}
+				result = append(result, reconcile.Request{NamespacedName: name})
+				r.Log.Info(fmt.Sprintf("REQUESTS: %v", result))
+			}
+		}
+		if len(result) > 0 {
+			return result
+		}
+		return nil
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&glancev1.GlanceAPI{}).
 		Owns(&keystonev1.KeystoneEndpoint{}).
@@ -192,6 +236,9 @@ func (r *GlanceAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.ConfigMap{}).
 		Owns(&appsv1.Deployment{}).
 		Owns(&routev1.Route{}).
+		// Watch for Glance{} ExtraVolumes
+		Watches(&source.Kind{Type: &glancev1.Glance{}},
+			handler.EnqueueRequestsFromMapFunc(glance)).
 		Complete(r)
 }
 
@@ -375,6 +422,37 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
 	instance.Status.Conditions.MarkTrue(condition.InputReadyCondition, condition.InputReadyMessage)
 	// run check OpenStack secret - end
 
+	// Get Glance Top level CR and fetch ExtraVolumes if defined
+	parentGlance := &glancev1.Glance{}
+
+	err = r.Client.Get(
+		context.TODO(),
+		types.NamespacedName{
+			Name:      glance.GetOwningGlanceName(instance),
+			Namespace: instance.Namespace,
+		},
+		parentGlance)
+
+	if err != nil {
+		if !k8s_errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		r.Log.Error(err, "Reconciling Service: Glance CR not present")
+	}
+
+	// if the top-level resource (parentGlance) is empty, an empty array is
+	// assigned to instance.ExtraMounts
+	instance.ExtraMounts = parentGlance.Spec.ExtraMounts
+
+	if len(instance.ExtraMounts) > 0 {
+		h, err := util.ObjectHash(instance.ExtraMounts)
+		if err != nil {
+			r.Log.Error(err, "Failed to get ExtraMountsHash")
+			return ctrl.Result{}, err
+		}
+		util.SetHash(instance.Status.Hash, "extraMounts", h)
+	}
+
 	//
 	// Create ConfigMaps and Secrets required as input for the Service and calculate an overall hash of hashes
 	//
@@ -488,7 +566,6 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
 		deplDef,
 		time.Duration(5)*time.Second,
 	)
-
 	ctrlResult, err = depl.CreateOrPatch(ctx, helper)
 	if err != nil {
 		instance.Status.Conditions.Set(condition.FalseCondition(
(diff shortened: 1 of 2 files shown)


controllers/glanceapi_controller.go
```diff
@@ -184,6 +166,35 @@ func (r *GlanceAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			return nil
+		}
+
+		for _, apis := range glanceapis.Items {
+			name := client.ObjectKey{
+				Namespace: o.GetNamespace(),
+				Name:      apis.Name,
+			}
+			result = append(result, reconcile.Request{NamespacedName: name})
+		}
```
abays, line 191: While I would say that we should check `ownerReferences` here, I suppose that we will only ever support one `Glance` CR per namespace anyhow.

fmount: I tried that: I thought that `client.Object` was supposed to get the current resources, but adding:

```
r.Log.Info(fmt.Sprintf("OWNER_REF: %v", o.GetOwnerReferences()))
r.Log.Info(fmt.Sprintf("Name: %s", o.GetName()))
```
I get:

```
2023-03-21T20:59:51+01:00       INFO    controllers.GlanceAPI   OWNER_REF: []
2023-03-21T20:59:51+01:00       INFO    controllers.GlanceAPI   Name: glance
```

There must be a wrong assumption on my side, but in general I agree we're going to support one `Glance` CR per namespace.

fmount: I made a few more tests and the change works and triggers a reconcile loop as long as I update the `ExtraVolumes`.
The assumption here is that we have one `Glance` "umbrella" per `Namespace`: in this case we should be good here. We can follow up in case we might want to support multiple `Glances` in the same `Namespace`, but I'd say this would be part of a bigger problem which I'm not sure is going to be supported.

abays: The `ownerReferences` would be on the objects in the `GlanceAPIList`.  I don't think it matters at the moment, however, because realistically we should only have one `Glance` top-level CR per namespace.  As you noted, we would have a bigger problem if two or more of them were present.

abays: One improvement we might want to make is to only have this `Watch` function enqueue a `GlanceAPI` if the `Glance.Spec.ExtraMount` has changed.  As it is right now, if anything changes on the `Glance` CR, this will trigger a reconciling of the `GlanceAPI`s.  That could lead to a double-enqueuing of `GlanceAPI`, if the changes made to the `Glance` CR causes the `Glance` reconcile loop to update the children `GlanceAPI`s.  In that case, those propagated changes would cause one enqueue of the `GlanceAPI`s, and then this new watch func would also get triggered for the initial changes to the `Glance` CR and cause a second enqueuing of the `GlanceAPI`s.

The way to check for differences in the `Glance.Spec.ExtraMounts` would be to store a hash of it in the `GlanceAPI.Status`.  Then, when the watch function is triggered, we calculate the hash of the `Glance.Spec.ExtraMounts` and compare each to each `GlanceAPI.Status.ExtraMountHash` and decide whether to enqueue or not.  We could also store the hash in the `Glance.Status` as well if we don't want to calculate it during the watch function (and in that case it would just be a comparison of `Glance.Status.ExtraMountHash` to `GlanceAPI.Status.ExtraMountHash`).

fmount: Right, so it might be safe checking that the `top-level CR` name (which comes with `client.Object`) matches the `ownerRef` name.
I'm going to update the patch adding this additional check, so we can eventually filter and avoid an unwanted reconcile loop in case the owner CR is not the one which has been changed.

fmount: +1 on watching the `ExtraMounts` only looking for the hash, as we can reduce (a lot) the scope of the entity we're watching.

fmount: @abays so I think I basically did what's described in your previous comment: we have a watcher over the top-level CR, and the request is created and enqueued only when the `ExtraMountsHash` (set in the `GlanceAPI.Status`) changes.
This is done by a function that is able to compare the `GlanceAPI.Status.ExtraMountsHash` (if defined) with the `util.HashObject(GlanceExtraMounts)` <- coming from the top-level CR (which is where we change the struct).
I'm not sure we can still improve things here, I'll do a few more tests and experiments, but feel free to give me your feedback on the approach I've taken so far.

abays: I think this latest iteration looks good.  As long as you're not seeing any reconcile loop looping, I think it will suit our purpose here.

controllers/glanceapi_controller.go
```diff
@@ -184,6 +166,65 @@ func (r *GlanceAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+
+		// Get GlanceAPIList that should be updated
+		glanceapis := &glancev1.GlanceAPIList{}
+
+		listOpts := []client.ListOption{
+			client.InNamespace(o.GetNamespace()),
+		}
+		if err := r.Client.List(context.Background(), glanceapis, listOpts...); err != nil {
+			r.Log.Error(err, "Unable to retrieve GlanceAPI CRs %v")
+			return nil
```
abays, line 182: Should we return `err` here instead?

controllers/glanceapi_controller.go
```diff
@@ -184,6 +166,65 @@ func (r *GlanceAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+				types.NamespacedName{
+					Name:      glance.GetOwningGlanceName(&api),
+					Namespace: o.GetNamespace(),
+				},
+				parentGlance)
+
+			// No parent in the OwnerReferences available, we should return
+			// as no reconcile loop should be triggered
+			if err != nil {
+				return nil
```
abays, line 203: Should we return `err` here instead?

controllers/glanceapi_controller.go
```diff
@@ -184,6 +166,65 @@ func (r *GlanceAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			}
+
+			// The new reconcile request should be enqueued as long as the
+			// ExtraMounts struct has changed in the top-level CR and the
+			// ownerRef matches with the client.Object
+			hashChanged, err := r.extraMountsHashChanged(&api, parentGlance.Spec.ExtraMounts)
+			if err != nil {
+				r.Log.Error(err, "Failing to get instance Hash")
+				return nil
```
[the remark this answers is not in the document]
fmount: same reply here (see previous comment)

controllers/glanceapi_controller.go
abays: Ah right, I forgot about the wrapping function.  All good then.

controllers/glanceapi_controller.go
fmount: same comment applies here

abays (commented):
+1 from me

controllers/glanceapi_controller.go
```diff
@@ -375,6 +419,37 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
...
+	if err != nil {
+		if !k8s_errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		r.Log.Error(err, "Reconciling Service: Glance CR not present")
+	}
+
+	// if the top-level resource (parentGlance) is empty, an empty array is
+	// assigned to instance.ExtraMounts
+	instance.ExtraMounts = parentGlance.Spec.ExtraMounts
```
gibizer, line 442: I consider this as hack. I think if data needs to flow from Glance to GlanceAPI that should flow through the GlanceAPI.Spec fields. I know that we try to reduce the CRD size but I think this step went too far in that regard. This change makes the connection between Glance and GlanceAPI unnecessarily bidirectional and complicates the design. This breaks the rule that the Spec defines the expected state of the resource, as it is now the Spec and the Parent's Spec defines it together.

controllers/glanceapi_controller.go
```diff
@@ -375,6 +422,37 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
...
+
+	if len(instance.ExtraMounts) > 0 {
+		h, err := util.ObjectHash(instance.ExtraMounts)
+		if err != nil {
+			r.Log.Error(err, "Failed to get ExtraMountsHash")
+			return ctrl.Result{}, err
+		}
+		util.SetHash(instance.Status.Hash, "extraMounts", h)
```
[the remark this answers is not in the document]
fmount: The status will have something like:

```
  hash:
    extraMounts: n5c6hf9h56dh57dh655h66fh647hd5h5d5h5b6h549hch58bh649h59ch664h66ch687h598h59ch578h564h7fh689h67bh5d6h95h86h679hb4h578h9dq
    input: n66ch549h585h595h9ch57fh696h588h5f8h97h65ch5c5h5d9h686h67hc9h656h5c9h578h698h8bh64dh64ch585h54bh584h547hf9h5f9hfh687h579q
```

controllers/glanceapi_controller.go
fmount: In theory as long as the pattern is well documented and the picture is clear it's not an hack. We store the hash of the data structure (as it would be an `input` for the sub-resource) and we react here in case it changes: the difference here is that instead being an external resource (e.g. the `TransportURL`), it's the parent.
This could probably be a pattern (unless a different strategy is known) to inject data from the top-level to the sub-resources, without exposing any parameter that could be changed in the CR.
However, the main point here is to figure out if I'm really breaking the rules or creating an anti-pattern: in that case I guess we won't be able to trim the CRD size and find a strategy at `meta-operator` level.
@dprince @abays @stuggi @Akrog thoughts on this?

gibizer: By using this pattern the GlanceAPI.Spec can be empty. Every field populated there is populated from the parent Glance instance. So instead of populating it, the GlanceAPI controller could read it out from it's parent and live with an empty Spec. I guess that would be super confusing.

TransportURL is different as GlanceAPI references a TransportURL instance by its name in a field in GlanceAPI.Spec. I think we discussed before the option to move ExtraMounts as a CRD and reference it by name in GlanceAPI.Spec. That would look a lot better to me as the GlanceAPI.Spec would still describe the whole expected state of GlanceAPI.

gibizer: Another option is to have ExtraMounts in GlanceAPI.Spec but does not include GlanceAPISpec directly into Glance.Spec as that triplicates the definition of ExtraMounts in Glance, once in Glance.ExtraMounts once in Glance.GlanceAPIExternal.ExtraMounts and once in Glance.GlanceAPIInternal.ExtraMounts. 

Instead create a GlanceAPITempalate that only has fields that are not directly populated from Glance.Spec. Then when the Glance controller creates GlanceAPI it creates it from GlanceAPITemplate + Glance.Spec.

As an example you can see how Nova.Spec does not include a full NovaAPISpec just a NovaAPITemplate.

gibizer: > I think I can't do what has been done for the `TransportURL` because of the nature of the data structure and the places where we want it to be injected and modified. I think also that the `GlanceAPITemplate` in the top level CR would be a clean approach to inject data as long as they're not duplicated in the sub-level resources. I see that basically a `NovaAPITemplate` is defined in [1], but I think the key point is having `NovaServiceBase` [2] that is included in `NovaAPISpec` [3] and in theory is not rendered in the CRD (the same way I did with `ExtraMounts`) because we have:
> 
> ```
> NovaServiceBase `json:",inline"`
> ```

The `inline` only spares a single named field layer in the CRD but all the content of NovaServiceBase is copied directly into NovaAPISpec: https://github.com/openstack-k8s-operators/nova-operator/blob/master/config/crd/bases/nova.openstack.org_nova.yaml#L54-L64 

> 
> Finally, you're building the whole spec in the `nova_controller` [4], and in the generated `novaAPI` the Spec is assigned:
> 
> ```
> 	op, err := controllerutil.CreateOrPatch(ctx, r.Client, api, func() error {
> 		api.Spec = apiSpec
> 
> 		err := controllerutil.SetControllerReference(instance, api, r.Scheme)
> 		if err != nil {
> 			return err
> 		}
> 
> 		return nil
> 	})
> ```
> 
> This should be enough to assign hidden fields to the sub-resource and avoid "watching" the parent resource and break the "top-down" approach, right?

There is no such thing as hidden field. If a CRD need a field filled then that field needs to show in in the CRD. What we can spare is how many times we mention the same field definition.

If the Glance.Spec.ExtraMounts is the only field that is written by the end user and the GlanceExternalAPI.Spec.ExtraMounts always holds a copy of the value of Glance.Spec.ExtraMounts then you don't need to add the ExtraMounts field in Glance.Spec.GlanceExternalAPI only in GlanceAPI.Spec. 

> 
> I'll create a separate patch where I try the approach suggested, so we can review and choose the way that works best. I'd like to stay consistent across the operators, so if that is the pattern, I'll try that. I'm also wondering if we can benefit of this approach to reduce the size of the `openstack-operator`, but that's another story I guess.
> 
> [1] https://github.com/openstack-k8s-operators/nova-operator/blob/master/api/v1beta1/novaapi_types.go#L34 [2] https://github.com/openstack-k8s-operators/nova-operator/blob/master/api/v1beta1/common_types.go#L26 [3] https://github.com/openstack-k8s-operators/nova-operator/blob/master/api/v1beta1/novaapi_types.go#L133 [4] https://github.com/openstack-k8s-operators/nova-operator/blob/master/controllers/nova_controller.go#L713

fmount: > I agree that we need to be careful to have bidirectional dependencies on the CRDs/controllers. I am wondering if it would be cleaner and consistent if we follow the same pattern as with networkAttachments or transporturl and have a dedicated CRD for the extraMounts a user can create. The extraMounts, which has the CRD kind as part of the propagation could then be passed in
> 
>     * either pass as a name list into the service operators by the openstack-operator
> 
>     * or the extraMounts controller could set an annotation on the service CRD to reference which one to use.
> 

I would not prefer breaking the existing pattern because it's already in all the operators: in addition, even though I know that everything is driven by the `openstack-operator`, I still want to keep the standalone use case for development reasons, hence having the ability to explicitly set `extraVolumes` in the top-level CR is a requirement.
I think that pattern of the shared CR is good, but not applies to this use case. 
> 
> Just two ideas, there might be more. With this I think the service CRDs could get rid of the big extraMount type def as part of their CRD.

As @gibizer suggested, I think an additional idea that keeps the current approach unchanged is implemented in #191.
That's enough to not repeat the same `struct` in the top-level CRD (where it appears only once, removing the references from the sub-resources).
It will also cut a lot on both `Cinder` and `Manila`, where the struct is repeated not only for the API but for other services (it's a x4 in case of cinder, a x3 in case of Manila).

abays:
@fmount Shall we close this one since we merged #191 ?

fmount:
> @fmount Shall we close this one since we merged #191 ?

Sure, thanks. Looks like #191 is the approach we're going to use in the other operators as well.

PR #190, opened 2023-04-03
Use consistent words for keywords

kajinamit:
This updates the keywords of cinder CRD so that the items are more consistent with the other operators.


fmount (approved):
Makes sense, thanks!

PR #191, opened 2023-04-03
Trim GlanceAPI CRD and optimize extraMounts usage

fmount:
The goal of this patch is to give an alternative approach to #184 where we tried to trim the `GlanceAPI` `CRD` as much as possible by injecting `extraMounts` from the top-level `Glance` `CR` and get a notification/event on a change coming from upper layer (that would trigger a reconciliation loop). However, the approach described in #184 "breaks" the top-down pattern where parameters are specified in the top level `Spec` and passed to the deployment: in a way that strategy "hides" data from the requests. The current patch represents an attempt to solve the same problem with a different strategy: it trims less than #184 (*) but keeps the pattern unchanged. In particular the `GlanceAPISpec` is built within the `Glance` controller, but it's not explicitly included in the `GlanceAPI`. A `GlanceAPITemplate`, instead, is included and used to define the top-level (user) input parameters provided and set to the sub-level resources. By doing this we can still keep the `extraMounts` definition in the `GlanceAPISpec`, but we don't replicate the same parameter in the top-level `CRD` (the `Spec` are not explicitly included).

(*)
```
> du -shc config/crd/bases/glance.openstack.org_glance*
48K     config/crd/bases/glance.openstack.org_glanceapis.yaml
52K     config/crd/bases/glance.openstack.org_glances.yaml
100K    total
```

VS

```
> du -shc config/crd/bases/glance.openstack.org_glance*
8.0K    config/crd/bases/glance.openstack.org_glanceapis.yaml
52K     config/crd/bases/glance.openstack.org_glances.yaml
60K     total
```

from #184.


api/v1beta1/common.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/openstack-k8s-operators/lib-common/modules/common/endpoint"
+)
+
+// GlanceAPITemplate defines the desired state of GlanceAPI
+type GlanceAPITemplate struct {
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=glance
+	// ServiceUser - optional username used for this service to register in glance
+	ServiceUser string `json:"serviceUser"`
```
gibizer, line 30: I'm wondering if we need both Glance.Spec.ServiceUser and Glance.Spec.GlanceAPI[Internal|External].ServiceUser as well in the top level. Or we can simply copy the Glance.Spec.ServiceUser to the GlanceAPI.Spec.ServiceUser directly?

api/v1beta1/common.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// +kubebuilder:validation:Required
+	// GlanceAPI Container Image URL
+	ContainerImage string `json:"containerImage"`
```
gibizer, line 34: It might worth to thing through if we need to set this differently per GlanceAPIInternalSpec / GlanceAPIExternalSpec. If we need to set it differently then yes, we need this in the Template, but if the two always have the same value then we can remove ContainerImage from the Template and copy the value from Glance.Spec.ContainerImage when creating the GlanceAPISpec struct.

api/v1beta1/common.go
```diff
@@ -0,0 +1,134 @@
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Maximum=32
+	// +kubebuilder:validation:Minimum=0
+	// Replicas of glance API to run
+	Replicas int32 `json:"replicas"`
+
+	// +kubebuilder:validation:Optional
+	// DatabaseHostname - Glance Database Hostname
+	DatabaseHostname string `json:"databaseHostname"`
```
gibizer, line 50: This is something that the end user don't know as it is depends on the value of Glance.Spec.DatabaseInstance. Isn't it?

api/v1beta1/common.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=glance
+	// DatabaseUser - optional username used for glance DB, defaults to glance
+	// TODO: -> implement needs work in mariadb-operator, right now only glance
+	DatabaseUser string `json:"databaseUser"`
```
gibizer, line 56: I guess both the internal and the external GlanceAPI uses the same DB schema with the same user / password. If so then this can be removed here and the value in GlanceAPI.Spec.DatabaseUser can be filled from Glance.Spec.DatabaseUser automatically.

api/v1beta1/common.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// +kubebuilder:validation:Optional
+	// Secret containing OpenStack password information for glance AdminPassword
+	Secret string `json:"secret"`
```
gibizer, line 60: I guess this can be filled from Glance.Spec.Secret as we don't expect a different Secret to be passed to GlanceAPIExternal and GlanceAPIInternal. Isn't so?

gibizer (commented):
I think there is additional fields in the Template now that can be removed.

api/v1beta1/common.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default={database: GlanceDatabasePassword, service: GlancePassword}
+	// PasswordSelectors - Selectors to identify the DB and ServiceUser password from the Secret
+	PasswordSelectors PasswordSelector `json:"passwordSelectors"`
```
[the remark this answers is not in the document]
fmount: +1

api/v1beta1/common.go
fmount: yeah, we get it from the `Glance` Status because it's set there.
We get the user from the upper layer, so I guess we can remove it from the template, right?

```
DatabaseHostname:       instance.Status.DatabaseHostname,
DatabaseUser:           instance.Spec.DatabaseUser,
```

api/v1beta1/common.go
fmount: Interesting, I think it's not supported having different images for External/Internal (and with webhooks we always enforce the same image), so I guess it makes sense removing this part from the template

api/v1beta1/common.go
fmount: Yes, we actually do in the controller, which means we don't need this in the template.

fmount:
@gibizer doing an addiitonal iteration to clean up the `GlanceAPITemplate` struct.

api/v1beta1/common.go
gibizer: yepp, these are then unused fields in the Template.

api/v1beta1/common.go
fmount: Actually, I'm thinking about keeping this parameter in the `Template` because I'm referencing the struct in [1] to inject the images via webhooks: this would keep the current logic unchanged.

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/master/api/v1beta1/glance_webhook.go#L66

gibizer: I think the code in [1] is just a consequence of having two separate levels where we can set the ContainerImage. I.e. if we have two levels to specify it then we need two levels to default it. But I'm OK to keep ContainerImage now. We can decide if we need two levels of specifying it later.

gibizer (commented):
Looks good to me, lets wait for the deploy job to be green then I'm OK to merge this.

gibizer:
> Looks good to me, lets wait for the deploy job to be green then I'm OK to merge this.

OK, I relaized that we still want to decide between the two trimming options. Sorry. So I vote for this solution instead of looking up the parent and read out fields from it.

gibizer:
/test glance-operator-build-deploy

Is it some repository problem? It seems the bundle is pushed successfully then later it cannot be pulled.

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_glance-operator/191/pull-ci-openstack-k8s-operators-glance-operator-master-glance-operator-build-deploy/1643173367794110464/artifacts/glance-operator-build-deploy/openstack-k8s-operators-build/build-log.txt
```
Successfully pushed quay.rdoproject.org/openstack-k8s-operators/openstack-operator-storage-bundle@sha256:75fe1c4deb2f299b7b93b994a78380fa101fb477d659719cb2ce8ec4022da5e4
Push successful
[...snip...]
 opm index add --bundles quay.rdoproject.org/openstack-k8s-operators/openstack-operator-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178,quay.rdoproject.org/openstack-k8s-operators/openstack-operator-storage-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178 --out-dockerfile index.Dockerfile --generate
time="2023-04-04T09:08:22Z" level=info msg="building the index" bundles="[quay.rdoproject.org/openstack-k8s-operators/openstack-operator-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178 quay.rdoproject.org/openstack-k8s-operators/openstack-operator-storage-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178]"
time="2023-04-04T09:08:22Z" level=error msg="permissive mode disabled" bundles="[quay.rdoproject.org/openstack-k8s-operators/openstack-operator-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178 quay.rdoproject.org/openstack-k8s-operators/openstack-operator-storage-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178]" error="[error resolving name : quay.rdoproject.org/openstack-k8s-operators/openstack-operator-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178: not found, image \"quay.rdoproject.org/openstack-k8s-operators/openstack-operator-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178\": not found]"
Error: [error resolving name : quay.rdoproject.org/openstack-k8s-operators/openstack-operator-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178: not found, image "quay.rdoproject.org/openstack-k8s-operators/openstack-operator-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178": not found]

```

fmount:
> /test glance-operator-build-deploy
> 
> Is it some repository problem? It seems the bundle is pushed successfully then later it cannot be pulled.
> 
> https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_glance-operator/191/pull-ci-openstack-k8s-operators-glance-operator-master-glance-operator-build-deploy/1643173367794110464/artifacts/glance-operator-build-deploy/openstack-k8s-operators-build/build-log.txt
> 
> ```
> Successfully pushed quay.rdoproject.org/openstack-k8s-operators/openstack-operator-storage-bundle@sha256:75fe1c4deb2f299b7b93b994a78380fa101fb477d659719cb2ce8ec4022da5e4
> Push successful
> [...snip...]
>  opm index add --bundles quay.rdoproject.org/openstack-k8s-operators/openstack-operator-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178,quay.rdoproject.org/openstack-k8s-operators/openstack-operator-storage-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178 --out-dockerfile index.Dockerfile --generate
> time="2023-04-04T09:08:22Z" level=info msg="building the index" bundles="[quay.rdoproject.org/openstack-k8s-operators/openstack-operator-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178 quay.rdoproject.org/openstack-k8s-operators/openstack-operator-storage-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178]"
> time="2023-04-04T09:08:22Z" level=error msg="permissive mode disabled" bundles="[quay.rdoproject.org/openstack-k8s-operators/openstack-operator-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178 quay.rdoproject.org/openstack-k8s-operators/openstack-operator-storage-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178]" error="[error resolving name : quay.rdoproject.org/openstack-k8s-operators/openstack-operator-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178: not found, image \"quay.rdoproject.org/openstack-k8s-operators/openstack-operator-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178\": not found]"
> Error: [error resolving name : quay.rdoproject.org/openstack-k8s-operators/openstack-operator-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178: not found, image "quay.rdoproject.org/openstack-k8s-operators/openstack-operator-bundle:5a38bd9e82681d456d1dcab00b7c9e7944db6178": not found]
> ```

yeah, it seems a flaky issue, let's see how the re-triggered job goes.

abays (approved):
/lgtm

PR #197, opened 2023-04-05
Fix make tidy not to depend on ./bin path

gibizer:
We should not assume that ./bin exists as it is not part of the git repository but created dynamically. Without this renovate might fail to generate a PR as ./bin does not exists in a fresh clone of the repo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,10 +121,11 @@ vet: gowork ## Run go vet against code.
 	go vet ./...
 	go vet ./... ./api/...
 
+APIPATH ?= $(shell pwd)/api
 .PHONY: tidy
 tidy: fmt
 	go mod tidy; \
-	pushd "$(LOCALBIN)/../api"; \
+	pushd $(APIPATH); \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -121,10 +121,11 @@ vet: gowork ## Run go vet against code.
 	go vet ./...
 	go vet ./... ./api/...
 
+APIPATH ?= $(shell pwd)/api
 .PHONY: tidy
 tidy: fmt
 	go mod tidy; \
-	pushd "$(LOCALBIN)/../api"; \
+	pushd $(APIPATH); \
```
fmount, line 128: I think this is covered in #196. I'm also wondering why we can't just have something like:

```
.PHONY: tidy
tidy: fmt
	go mod tidy
	cd ./api
	go mod tidy
```

given we're already in the root directory of the operator.

gibizer:
closing this in favor of https://github.com/openstack-k8s-operators/glance-operator/pull/196

PR #199, opened 2023-04-07
Ctrl manager labels for webhooks and logging

abays:
For initial webhook work, we found that we needed to remove the `control-plane: controller-manager` label and label-selector from the `controller-manager` pod and its associated peripherals in order to properly direct webhook traffic to the appropriate webhook server for our various operators' CRDs.  This had the unfortunate side effect of removing a label that could be used for bulk-selecting all our `controller-manager` pods.  This was noted here [1].  We therefore have opted to reintroduce the `control-plane: controller-manager` default label (originally added by `operator-sdk` scaffolding) and instead use another unique label (also added by `operator-sdk` scaffolding, at least in newer versionf of `operator-sdk`) for routing webhook traffic.  This unique label takes the form of `app.kubernetes.io/name: <operator name>` and get be used with label-selectors to ensure webhooks work correctly.

[1] https://github.com/openshift/release/pull/37084#discussion_r1138314478

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -4,18 +4,22 @@ metadata:
   name: controller-manager
   namespace: system
   labels:
-    control-plane: glance-controller-manager
+    control-plane: controller-manager
+    app.kubernetes.io/name: glance-operator
+    app.kubernetes.io/component: glance
 spec:
   selector:
     matchLabels:
-      control-plane: glance-controller-manager
+      app.kubernetes.io/name: glance-operator
   replicas: 1
   template:
     metadata:
       annotations:
         kubectl.kubernetes.io/default-container: manager
       labels:
-        control-plane: glance-controller-manager
+        control-plane: controller-manager
+        app.kubernetes.io/name: glance-operator
+        app.kubernetes.io/component: glance
     spec:
       securityContext:
         runAsNonRoot: true
--- a/config/prometheus/monitor.yaml
+++ b/config/prometheus/monitor.yaml
@@ -4,7 +4,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   labels:
-    control-plane: glance-controller-manager
+    control-plane: controller-manager
   name: controller-manager-metrics-monitor
   namespace: system
 spec:
@@ -17,4 +17,4 @@ spec:
         insecureSkipVerify: true
   selector:
     matchLabels:
-      control-plane: glance-controller-manager
+      app.kubernetes.io/name: glance-operator
--- a/config/rbac/auth_proxy_service.yaml
+++ b/config/rbac/auth_proxy_service.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: Service
 metadata:
   labels:
-    control-plane: glance-controller-manager
+    control-plane: controller-manager
   name: controller-manager-metrics-service
   namespace: system
 spec:
@@ -12,4 +12,4 @@ spec:
     protocol: TCP
     targetPort: https
   selector:
-    control-plane: glance-controller-manager
+    app.kubernetes.io/name: glance-operator
--- a/config/webhook/service.yaml
+++ b/config/webhook/service.yaml
@@ -17,4 +17,4 @@ spec:
       protocol: TCP
       targetPort: 9443
   selector:
-    control-plane: glance-controller-manager
+    app.kubernetes.io/name: glance-operator


config/webhook/service.yaml
```diff
@@ -17,4 +17,4 @@ spec:
       protocol: TCP
       targetPort: 9443
   selector:
-    control-plane: glance-controller-manager
+    app.kubernetes.io/name: glance-operator
```
abays, line 20: This alteration to the default `operator-sdk` scaffolding is what allows webhook traffic to be properly routed and not clash with other operators.

dprince:
I'm good with it. Letting it sit till Monday to get another ack

config/prometheus/monitor.yaml
```diff
@@ -17,4 +17,4 @@ spec:
         insecureSkipVerify: true
   selector:
     matchLabels:
-      control-plane: glance-controller-manager
+      control-plane: controller-manager
```
rabi, line 20: AFAIK you need to have unique newly added label here too for the endpoint..

config/rbac/auth_proxy_service.yaml
```diff
@@ -12,4 +12,4 @@ spec:
     protocol: TCP
     targetPort: https
   selector:
-    control-plane: glance-controller-manager
+    control-plane: controller-manager
```
rabi, line 15: same here.

config/prometheus/monitor.yaml
abays: I haven't seen any salient problem in my testing, but what you're saying makes sense.

abays: I think for now I would opt not to change this here or with `config/rbac/auth_proxy_service.yaml`, since they pre-existed webhooks (see Nova operator [1][2] as an example, which does not yet have webhooks) and seem to have been working fine so far.  Perhaps, however, they just haven't been properly tested.  Anyhow, I'd just like to deal with the webhooks in this patch and we can follow-up with other PRs for these other label selectors.

[1] https://github.com/openstack-k8s-operators/nova-operator/blob/master/config/prometheus/monitor.yaml#L18-L20
[2] https://github.com/openstack-k8s-operators/nova-operator/blob/aa6184d9b90492a975f56c1791994072c48ad824/config/rbac/auth_proxy_service.yaml#L14-L15

rabi: Not sure what "working fine means here" :smile:  The issue is not webhook specific as I mentioned in https://issues.redhat.com/browse/OSP-23156. AFAICT we don't test them and don't see the issue. Also, this change is reverting something which fixed it earlier.

abays: > Also, this change is reverting something which fixed it earlier.

In the context of webhooks, it is indeed reverting the `<operator name>-controller-manager` labeling, but it is not "unfixing" the functionality in that regard.  The webhook traffic will still be routed properly with the new `app.kubernetes.io/name: glance-operator` label and label selector.  I have tested both `Glance` and `Keystone` operator deployed at the same time with these new labels, and their respective webhooks properly received traffic (I repeated the test over and over so as to rule-out that I was "just getting lucky").

As I mentioned above regarding the Prometheus and RBAC concerns, "`Perhaps, however, they just haven't been properly tested`".  We should figure out what that testing would look like.

abays: I wonder if we could just use the new `app.kubernetes.io/name: <operator name>-operator` label here [1] and here [2].

[1] https://github.com/abays/glance-operator/blob/30ad3b2abf249cba2ec93235fc8323ccdea09d25/config/rbac/auth_proxy_service.yaml#L14-L15
[2] https://github.com/abays/glance-operator/blob/30ad3b2abf249cba2ec93235fc8323ccdea09d25/config/prometheus/monitor.yaml#L18-L20

rabi: >  but it is not "unfixing" the functionality in that regard

Well, we're assuming the earlier change was `only` in the context of webhooks, which is not correct.  I changed it for all services including webhook service.  Though I did not change in all operators.

rabi: > I wonder if we could just use the new app.kubernetes.io/name: <operator name>-operator label here [1] and here [2].

We could/should and that's what I was suggesting in my comment I think. They have to be unique labels for the operator deployments.

abays: > > but it is not "unfixing" the functionality in that regard
> 
> Well, we're assuming the earlier change was `only` in the context of webhooks, which is not correct. I changed it for all services including webhook service. Though I did not change in all operators.

That's not what I said.  `"In the context of webhooks, it is indeed..."`.  I was speaking of webhooks.  Then later I mentioned `"...regarding the Prometheus and RBAC concerns"`, which, yes, is another concern affected by these labels.  But that is something we need to qualify with a proper test so to understand how to determine what the proper approach is here.

abays: > > I wonder if we could just use the new app.kubernetes.io/name: -operator label here [1] and here [2].
> 
> We could/should and that's what I was suggesting in my comment I think. They have to be unique labels for the operator deployments.

Okay, I will change those and see what happens.  And we can follow-up with investigating how to properly test the Prometheus/RBAC stuff.

gibizer (approved):
looks good to me

config/manager/manager.yaml
```diff
@@ -4,18 +4,22 @@ metadata:
...
   labels:
-    control-plane: glance-controller-manager
+    control-plane: controller-manager
+    app.kubernetes.io/name: glance-operator
+    app.kubernetes.io/component: glance
 spec:
   selector:
     matchLabels:
-      control-plane: glance-controller-manager
+      app.kubernetes.io/name: glance-operator
```
fmount, line 13: Sorry @abays I didn't have enough time to properly review the patch here. 
Even though it's already merged, I'm +1 on the unique label to select the operator and properly route the requests.

PR #201, opened 2023-04-12
Propagate ServiceUser to GlanceAPI

gibizer:
Glance.Spec.ServiceUser allows defining the name of the service user glance services should use. However after a recent refactor in 5a38bd9e82681d456d1dcab00b7c9e7944db6178 the ServiceUser is not propagate from Glance.Spec to GlanceAPI.Spec. This patch fixes it.


gibizer:
I did a quick scan on GlanceAPISpec and I think now every field of it is set.

PR #206, opened 2023-04-19
Backend config samples

konan-abhi:
This patch adds a some backend samples for ceph and nfs.

The idea is for people to be able to just look at one of the files in `config/samples/backends/*/backend.yaml` and see just the relevant bits to configure the Glance backend.

### DIFF
--- a/config/samples/backends/nfs/backend.yaml
+++ b/config/samples/backends/nfs/backend.yaml
@@ -1,0 +1,41 @@
+# Sample using File/NFS as a glance backend
+# Requires a filepath ('/var/lib/glance/image/') is mounted on NFS share
+apiVersion: core.openstack.org/v1beta1
+kind: OpenStackControlPlane
+metadata:
+  name: openstack
+spec:
+  serviceUser: glance
+  containerImage: quay.io/podified-antelope-centos9/openstack-glance-api:current-podified
+  customServiceConfig: |
+    [DEFAULT]
+    enabled_backends = default_backend:file
+    [glance_store]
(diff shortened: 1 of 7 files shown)


config/samples/backends/nfs/backend.yaml
```diff
@@ -0,0 +1,28 @@
...
+spec:
+  serviceUser: glance
+  containerImage: quay.io/podified-antelope-centos9/openstack-glance-api:current-podified
+  customServiceConfig: |
+    [DEFAULT]
+    enabled_backends = default_backend:file
+    [glance_store]
+    default_backend = default_backend
+    [default_backend]
+    filesystem_store_datadir = /var/lib/glance/images/
```
fmount, line 16: I guess we need here `extraMounts` w/ the nfs export we would like glance to be pointed to.
I'm going to update this example accordingly as long as I test it!

konan-abhi: Are you sure extraVolType should be Ceph here?

fmount: no sorry, bad copy/paste, `extraVolType` is a label we can add to the extraMount to facilitate some logic in the operator. No relevant here, but definitely `s/ceph/nfs` is what we need

fmount: other than this comment you had, in general it looks good!

konan-abhi: Done

fmount (approved):

PR #207, opened 2023-04-19
Remove crudini dependency and unused scripts

fmount:
This change is part of a series that aim to remove the cruidini dependency and be consistent with what we've done in Cinder PR#108. In particular, secret snippets can be rendered in the config exposing customServiceConfigSecrets and processed by the init script.


fmount:
@konan-abhi  This is the first patch of a series: the goal is to remove initContainer and let the config being generated in the operator.

controllers/glance_controller.go
```diff
@@ -696,12 +701,11 @@ func (r *GlanceReconciler) generateServiceConfigMaps(
 	cms := []util.Template{
 		// ScriptsConfigMap
 		{
-			Name:               fmt.Sprintf("%s-scripts", instance.Name),
-			Namespace:          instance.Namespace,
-			Type:               util.TemplateTypeScripts,
-			InstanceType:       instance.Kind,
-			AdditionalTemplate: map[string]string{"common.sh": "/common/common.sh"},
```
konan-abhi, line 703: Is this related to this change?

controllers/glanceapi_controller.go
```diff
@@ -573,12 +612,11 @@ func (r *GlanceAPIReconciler) generateServiceConfigMaps(
 	cms := []util.Template{
 		// ScriptsConfigMap
 		{
-			Name:               fmt.Sprintf("%s-scripts", instance.Name),
-			Namespace:          instance.Namespace,
-			Type:               util.TemplateTypeScripts,
-			InstanceType:       instance.Kind,
-			AdditionalTemplate: map[string]string{"common.sh": "/common/common.sh"},
```
konan-abhi, line 580: ditto, is this related?

templates/glance/bin/init.sh
```diff
@@ -17,35 +17,61 @@ set -ex
...
-# Merge all templates from config-data
-for dir in /var/lib/config-data/default; do
-    merge_config_dir ${dir}
-done
-
-# set secrets
-# TODO: transportUrl (either set here or elsewhere)
-#crudini --set ${SVC_CFG_MERGED} DEFAULT transport_url $TRANSPORTURL
-crudini --set ${SVC_CFG_MERGED} database connection mysql+pymysql://${DBUSER}:${DBPASSWORD}@${DBHOST}/${DB}
-crudini --set ${SVC_CFG_MERGED} keystone_authtoken password $GLANCEPASSWORD
```
fultonj, line 49: so rather than use crudini to add these lines....

templates/glance/config/glance-api-config.json
```diff
@@ -2,10 +2,10 @@
     "command": "/usr/bin/glance-api --config-dir /etc/glance/glance.conf.d",
     "config_files": [
       {
-        "source": "/var/lib/config-data/merged/glance-api.conf",
-        "dest": "/etc/glance/glance.conf.d/00-glance-api.conf",
+        "source": "/var/lib/config-data/merged/glance.conf.d",
```
fultonj, line 5: Here we reference the new directory.

controllers/glanceapi_controller.go
```diff
@@ -162,6 +165,40 @@ func (r *GlanceAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			return nil
+		}
+		for _, cr := range apis.Items {
+			for _, v := range cr.Spec.CustomServiceConfigSecrets {
+				if v == secretName {
+					name := client.ObjectKey{
+						Namespace: namespace,
+						Name:      cr.Name,
+					}
+					r.Log.Info(fmt.Sprintf("Secret %s is used by Cinder CR %s", secretName, cr.Name))
```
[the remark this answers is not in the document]
fultonj: I agree perhaps change s/Cinder/Glance or just drop the name. I assume this is a copy/paste mistake... unless it's actually using Cinder

controllers/glance_controller.go
fmount: Yes, I'm basically removing `AdditionalTemplate: map[string]string{"common.sh": "/common/common.sh"},` as it's not used anymore.

controllers/glanceapi_controller.go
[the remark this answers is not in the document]
fmount: Yep, have to change that in my next iteration.

controllers/glanceapi_controller.go
fmount: Same as the previous comment

controllers/glance_controller.go
fultonj: It seems the diff is bigger than it has to be; a whitespace issue. I don't think that's a problem though. I think the real difference here is that `AdditionalTemplate: map[string]string{"common.sh": "/common/common.sh"},` is removed.

fultonj:
Look's OK to me.

templates/glance/bin/init.sh
```diff
@@ -17,35 +17,61 @@ set -ex
...
+EOF
+
+cp ${DEFAULT_DIR}/glance-api.conf ${SVC_CFG_MERGED_DIR}/00-default.conf
+
+# Generate 01-deployment-secrets.conf
+DEPLOYMENT_SECRETS=${SVC_CFG_MERGED_DIR}/01-deployment-secrets.conf
+
+cat <<EOF >> ${DEPLOYMENT_SECRETS}
+[database]
+connection = mysql+pymysql://${DBUSER}:${DBPASSWORD}@${DBHOST}/${DB}
```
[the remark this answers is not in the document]
fmount: Yes, the long term plan is to remove entirely the `init.sh` script and generate the config within the operator: this will result in simplifying the operator as we don't have to run the initContainer anymore.
However, this requires some experiments, so let's say this is the first step to reach that goal.

konan-abhi (approved):
Looks good to me, thank you!

PR #208, opened 2023-04-19
Remove non-existing pvc field from GlanceAPI spec

cescgina:
Some fields were removed from the GlanceAPI spec in a recent change, so
they need to be removed from the check in kuttl tests as well.


abays (approved):
/lgtm

fmount:
late here but +2

PR #210, opened 2023-04-20
Dynamic serviceaccount/role/rolebinding creation

dprince:
This PR dyamically creates the above resources within the namespace of the requested Glance instance.

The above resources are no longer created via OLM.

This allows the operator to deploy openstack service into any namespace within the OpenShift cluster.


dprince:
missing some logic to set the new rbac conditions to true

abays (approved):
/lgtm

PR #211, opened 2023-04-23
Propagate secretConf to the GlanceAPI pods

fmount:
The previous patch removed the crudini dependency and introduced a new parameter, `CustomServiceConfigSecrets`, that can be used to define a list of `Secrets` (by name) that should be processed by the operator and the `init.sh` script and propagated to the `Glance` service. The previous patch neglected to add the `pkg/glance` related change that is supposed to process/mount the resulting Volumes, and the purpose here is to fix that scenario.


api/v1beta1/common.go
```diff
@@ -17,8 +17,8 @@ limitations under the License.
 package v1beta1
 
 import (
-	corev1 "k8s.io/api/core/v1"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/endpoint"
+	corev1 "k8s.io/api/core/v1"
```
konan-abhi, line 21: Does this shift makes any impact? (me not aware about import positioning)

pkg/glance/volumes.go
```diff
@@ -70,11 +71,13 @@ func GetVolumes(name string, pvcName string, extraVol []glancev1.GlanceExtraVolM
 			vm = append(vm, vol.Volumes...)
 		}
 	}
+	secretConfig, _ := GetConfigSecretVolumes(secretNames)
+	vm = append(vm, secretConfig...)
```
konan-abhi, line 75: Minor comment: Should we extract this in separate small method which will return vm and call it here and at line 104 instead duplicating the code?

pkg/glanceapi/volumes.go
```diff
@@ -45,5 +46,7 @@ func getInitVolumeMounts(extraVol []glancev1.GlanceExtraVolMounts, svc []storage
 			vm = append(vm, vol.Mounts...)
 		}
 	}
+	_, secretConfig := glance.GetConfigSecretVolumes(secretNames)
+	vm = append(vm, secretConfig...)
```
konan-abhi, line 50: May be call that method here as well?

konan-abhi (commented):
Some minor comments inside, overall looks good.
Thank you!

pkg/glance/volumes.go
fmount: so I think this is not helping w/ de-duplication. In order to de-duplicate some code I created that function that returns you both `corev1.Volume` and `corev1.VolumeMount`: according to the place where it's called, I can choose the object I need and "append" to the existing slice.
Saving those two lines, imo, doesn't add value, and actually hides something that we might want to see to improve readability.
If can help, in the long term, where we want to generate the config secrets in the operator, that part might go away, so I would take the code as it is (also because I has been tested) and refactor in a follow up patch that takes care about config generation and init script removal.
@konan-abhi wdyt?

api/v1beta1/common.go
fmount: No it's not, the order is managed by the golang linter: I usually run `pre-commit` checks (defined in the repo) that takes care about this kind of things.

pkg/glanceapi/volumes.go
fmount: see my previous comment about the reason I'd like to keep it here.

pkg/glance/volumes.go
konan-abhi: Sounds good 👍🏻

konan-abhi (approved):
Thank you for answers!

PR #212, opened 2023-04-24
Remove APIType from GlanceAPITemplate

abays:
This PR removes the `APIType` from `GlanceAPITemplate` (and sets it appropriately in the `Glance` controller), and also changes the `Replicas` field in `GlanceAPITemplate` to be a pointer so that the defaulting webhook does not set it to the zero-value and subvert the `kubebuilder` defaulting for that field.


fmount (approved):

api/v1beta1/glance_webhook.go
```diff
@@ -75,6 +75,10 @@ func (spec *GlanceSpec) Default() {
 	if spec.GlanceAPIInternal.ContainerImage == "" {
 		spec.GlanceAPIInternal.ContainerImage = glanceDefaults.ContainerImageURL
 	}
+
+	// Force the external and internal GlanceAPI children to have the proper APIType
+	spec.GlanceAPIExternal.APIType = APIExternal
+	spec.GlanceAPIInternal.APIType = APIInternal
```
gibizer, line 81: This now ignores the value coming from the user. At least this should be documented in the description of the field.

gibizer (changes requested):

api/v1beta1/glance_webhook.go
gibizer: Or even better, if we agree to hard code the type then we can drop the APIType field from the GalnceAPITemplate and only have it in the GlanceAPISpec.

gibizer: Ohh the controller already hard codes it https://github.com/openstack-k8s-operators/glance-operator/blob/4e086b0f7de33e5e824ac3b5b3fa52597a4120af/controllers/glance_controller.go#L555 So I think we can simply drop the APIType field from the Template.

fmount: Long term I have to revisit how the APIType is used, I have it in my backlog, so I'm fine with either a comment in the webhook or removing the field, but I feel I have to reintroduce it in the future and change L555 in the controller to fit well the logic I have in mind when we're not going to have only Internal/External, but an []GlanceAPI (that can be a single instance, Internal/External, a list created by some criteria).

fmount: Definitely more, but we're still trying to design the appropriate solution.
If `backend` is `File` we don't have to split, and 1 API is enough, otherwise we need at least 2 (internal/external).
However there are use cases (that we'll see later in Q3/Q4) that are supposed to cover dedicated resource assignments or even Edge, which requires us to build a slice instead of just hardcoding internal and external.
As I mentioned, however, not a priority for the short term, so I'm good with the current solution.

gibizer: @fmount Thanks for the explanation. It explains why []GlanceAPI is needed and not just two optional fields.

abays: I switched this PR to remove the `APIType` from `GlanceAPITemplate` (and set it appropriately in the `Glance` controller), and to also change the `Replicas` field in `GlanceAPITemplate` to be a pointer so that the defaulting webhook does not set it to the zero-value and subvert the `kubebuilder` defaulting for that field.

PR #214, opened 2023-04-28
Remove initcontainer and process secrets passed as input

fmount:
There are many goals achieved by this patch:

1. we're able to remove (entirely) the initcontainer and the related scripts that are no longer required to start the service

2. kolla is still used to copy files from `src` to `dst`, as they are rendered and mounted accordingly with the right permissions
    in the destination directory

4. Controllers are able to parse many secrets referenced by the `CustomServiceConfigSecrets` parameter and merge them into a single secret (`04-secret.conf`) which is passed to the `deployment` and mounted to the target pod

Still WIP but good progress so far on the main goal.


pkg/glance/dbsync.go
```diff
@@ -27,19 +27,20 @@ import (
...
 	instance *glancev1.Glance,
 	labels map[string]string,
 	annotations map[string]string,
+	confSecrets []string,
 ) *batchv1.Job {
 	runAsUser := int64(0)
 
-	dbSyncExtraMounts := []glancev1.GlanceExtraVolMounts{}
-	secretNames := []string{}
+	//dbSyncExtraMounts := []glancev1.GlanceExtraVolMounts{}
```
konan-abhi, line 43: nit: We can remove line 43 if its not relevant anymore.

templates/glance/config/glance-api.conf
```diff
@@ -8,7 +8,7 @@ bind_host=0.0.0.0
 bind_port=9292
 workers=3
 image_cache_dir=/var/lib/glance/image-cache
-log_config_append=/etc/glance/logging.conf
+log_config_append=/etc/glance/glance.conf.default.d/logging.conf
 enabled_backends=default_backend:file
```
konan-abhi, line 12: I think similar path change is required at line 38 and 41

konan-abhi (commented):
Few comments are nit and can be addressed later but one should be addressed before merging this (last comment)

controllers/glance_controller.go
```diff
@@ -732,29 +736,92 @@ func (r *GlanceReconciler) generateServiceConfigMaps(
...
+}
+
+// GenDeploymentSecret - a function that produces a basic 01-deployment-secret
+func (r *GlanceReconciler) GenDeploymentSecret(
+	ctx context.Context,
+	helper *helper.Helper,
+	instance *glancev1.Glance,
+) error {
+
+	secretName := "01-deployment-secret"
```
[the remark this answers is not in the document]

pkg/glance/dbsync.go
fmount: It's actually needed, I was just experimenting with this draft on a couple of approaches.

templates/glance/config/glance-api.conf
fmount: Right, thanks!

fmount: did it for policy, I'll do it later for the other files

controllers/glance_controller.go
```diff
@@ -374,11 +374,19 @@ func (r *GlanceReconciler) reconcileInit(
 		instance.Status.Hash = map[string]string{}
 	}
 
+	// Define new deployment secret
+	err = r.GenDeploymentSecret(ctx, helper, instance)
+	if err != nil {
+		return ctrlResult, err
+	}
+
+	secrets := []string{"01-deployment-secret"}
```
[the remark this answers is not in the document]
fmount: done

konan-abhi (approved):
Looks better now, thank you!

pkg/glance/volumes.go
```diff
@@ -19,26 +19,19 @@ import (
...
-	var scriptsVolumeDefaultMode int32 = 0755
-	var config0640AccessMode int32 = 0640
+func GetVolumes(
+	name string,
+	pvcName string,
+	extraVol []glancev1.GlanceExtraVolMounts,
+	svc []storage.PropagationType,
+	confSecret []string,
+) []corev1.Volume {
+	var config0640AccessMode int32 = 0644
```
bogdando, line 32: there is a mismatch in the name and its value

pkg/glance/volumes.go
```diff
@@ -64,60 +51,46 @@ func GetVolumes(name string, pvcName string, secretNames []string, extraVol []gl
...
+) []corev1.VolumeMount {
 
 	vm := []corev1.VolumeMount{
 		{
-			Name:      "scripts",
-			MountPath: "/usr/local/bin/container-scripts",
-			ReadOnly:  true,
+			Name:      "config-data",
+			MountPath: "/var/lib/config-data/merged",
+			ReadOnly:  false,
```
bogdando, line 89: this could be also good time to make volume mounts immutable, or allowing changes in Debug mode only

pkg/glance/volumes.go
```diff
@@ -132,36 +105,37 @@ func GetVolumeMounts(secretNames []string, extraVol []glancev1.GlanceExtraVolMou
...
-	for idx, secretName := range secretNames {
-		secretVol := corev1.Volume{
-			Name: secretName,
-			VolumeSource: corev1.VolumeSource{
-				Secret: &corev1.SecretVolumeSource{
-					SecretName:  secretName,
-					DefaultMode: &config0640AccessMode,
-				},
+func GetConfigSecretVolumes(secretName string) (corev1.Volume, corev1.VolumeMount) {
+	var config0640AccessMode int32 = 0644
```
bogdando, line 121: mismatching name and value

fmount: +1 good catch, fixed.

PR #217, opened 2023-05-03
correct DEFAULT_IMG to match what is in kustomize

yazug:
currently when you run make bundle a file is left with changes This patch updates the Makefile default to match
what is checked in.


fmount (approved):
makes sense, thanks

PR #218, opened 2023-05-05
Makefile cleanup

yazug:
cleanup make vet target duplicates
enforce no leftovers from make bundle instead of just make manifests


.pre-commit-config.yaml
```diff
@@ -7,11 +7,11 @@ repos:
       entry: make
       args: ["tidy"]
       pass_filenames: false
-    - id: make-manifests
-      name: make-manifests
+    - id: make-bundle
```
fmount, line 10: It makes sense as long as `manifests` is called by `bundle`.
Is this a pattern we're going to spread across the board?

Makefile
```diff
@@ -119,7 +119,7 @@ fmt: ## Run go fmt against code.
 .PHONY: vet
 vet: gowork ## Run go vet against code.
 	go vet ./...
```
fmount, line 121: I'm wondering if running:

```
go vet ./... ./api/...
```

is equivalent to:

```
go vet ./...
go vet ./api/...
```

yazug: no issue just normalized to the pattern other repos had where they had one folder per call.  but don't care one way or the other.

yazug: was just simplifying ... either remove the redundant ./... from the double call or remove the call above it

.pre-commit-config.yaml
yazug: yes that would be the plan if this works to spread this same change to all repos for the default.

abays:
Somehow the `make bundle` in `pre-commit` is using a bad version for the operator version.  It looks like it's using the Golang version somehow:

```
/go/src/github.com/openstack-k8s-operators/operator/bin/kustomize build config/manifests | operator-sdk generate bundle -q --overwrite --version 1.19  
Error: invalid command options: 1.19 is not a valid semantic version: No Major.Minor.Patch elements found
```

PR #219, opened 2023-05-09
Webhook cleanup

abays:
Cleans up the webhook mechanism based on the latest improvements to the pattern used across our operators.


abays:
> +1 from me in general. I'm just wondering if it makes sense define the `GlanceAPIContainerImage` in `pkg/glance/const.go` to have a single place where we gather GlanceAPI defaults. It's the only question I have, otherwise looks good.

We can't do that, otherwise it would create a circular dependency between the `api` module and the main one.

fmount:
> > +1 from me in general. I'm just wondering if it makes sense define the `GlanceAPIContainerImage` in `pkg/glance/const.go` to have a single place where we gather GlanceAPI defaults. It's the only question I have, otherwise looks good.
> 
> We can't do that, otherwise it would create a circular dependency between the `api` module and the main one.

Right, makes sense. I think we're good then.

fmount (approved):

PR #221, opened 2023-05-15
WIP - Remove initcontainer and process secrets passed as input

fmount:
There are many goals achieved by this patch:

1. we're able to remove (entirely) the initcontainer and the related scripts that are no longer required to start the service

2. kolla is still used to copy additional files and templates from src to dst, as they rendered accordingly (secrets for sensitive information and configmap for regular overrides) and mounted to the destination directory

3. Controllers are able to parse many secrets referenced by the CustomServiceConfigSecrets parameter and merge them into a single secret (04-secret.conf) which is passed to the deployment and mounted to the target pod

Still WIP but good progress so far on the main goal.


fmount:
This is too old at this point and all this patch has been reworked in https://github.com/openstack-k8s-operators/glance-operator/pull/295. Closing in favor of https://github.com/openstack-k8s-operators/glance-operator/pull/295

PR #222, opened 2023-05-15
Enforce Per-Tenant Quotas configuration

fmount:
This patch exposes the proper structure to enforce per-tenant quotas in `GlanceAPI`.
If limits are provided,  the related glance-api section is enabled and `templateParameters` are populated accordingly.

Depends-On: openstack-k8s-operators/lib-common/pull/284


templates/glance/config/glance-api.conf
```diff
@@ -54,3 +55,22 @@ filesystem_store_datadir = /var/lib/glance/os_glance_staging_store/
...
+system_scope = all
+{{ if (index . "DomainID") }}
+user_domain_id = {{ .DomainID }}
+{{ end }}
+{{ if (index . "EndpointID") }}
+endpoint_id = {{ .EndpointID }}
+{{ end }}
+{{ if (index . "Region") }}
+region_name = {{ .Region }}
+{{ end }}
```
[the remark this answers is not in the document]
fmount: Yeah, that one is needed, adding in my next iteration!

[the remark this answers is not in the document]
fmount: Done

konan-abhi:
Is this still WIP, since we are not registering the limits in keystone yet?

konan-abhi:
Sorry I missed that this is still WIP.

fmount:
> Is this still WIP, since we are not registering the limits in keystone yet?

Yes, I'm planning to finish that part (it's the only missing piece at this point) by EOW!

config/samples/glance_v1beta1_glance_quota.yaml
```diff
@@ -0,0 +1,29 @@
...
+    debug:
+      service: false
+    preserveJobs: false
+    replicas: 1
+  secret: osp-secret
+  storageClass: ""
+  storageRequest: 10G
+  quotas:
+    imageSizeTotal: 100
+    imageStageTotal: 100
```
konan-abhi, line 27: Even though it is sample template, I think we should keep default for above two to 1000 (similar to upstream glance default)
https://docs.openstack.org/glance/latest/admin/quotas.html#configuring-glance-for-per-tenant-quotas

controllers/glance_controller.go
```diff
@@ -766,3 +781,48 @@ func (r *GlanceReconciler) generateServiceConfigMaps(
...
+	if err != nil {
+		return limit, err
+	}
+	for lName, lValue := range quota {
+		r.Log.Info(fmt.Sprintf("Registering limit %s", lName))
+		defaultRegion := o.GetRegion()
+		m := openstack.RegisteredLimit{
+			RegionID:     defaultRegion,
+			ServiceID:    instance.Status.ServiceID,
+			Description:  "default",
```
konan-abhi, line 818: I think we can skip this or generate value like, "default limit for " + lNmae

controllers/glance_controller.go
```diff
@@ -246,6 +247,9 @@ func (r *GlanceReconciler) reconcileDelete(ctx context.Context, instance *glance
 		}
 	}
 
+	//TODO (fpantano): if quota limits are registered, we should remove.
```
konan-abhi, line 250: We should remove or update the registered limits?
Didn't understood this todo here.

controllers/glance_controller.go
fmount: +1 on a description like the one you mentioned! I'll do that in the next PS

config/samples/glance_v1beta1_glance_quota.yaml
fmount: sure, let me update it to reflect that doc.

api/bases/glance.openstack.org_glances.yaml
```diff
@@ -1026,6 +1032,30 @@ spec:
               preserveJobs:
                 default: false
                 type: boolean
+              quotas:
+                properties:
+                  imageCountTotal:
+                    default: 0
```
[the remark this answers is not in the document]
fmount: Right now the `tempest` failure is unrelated (it's due to some networking issue on the `SNO` environment), but in general this default is set to `0` because we want the user to explicitly pass the values documented in [1]. 
By having them set to `0` by default allow us to **not** enable the feature if the `quotas` struct is not passed in the `Glance` CR (and this preserves backward compatibility with what we already have in place):

```golang
// IsQuotaEnabled - return true if one of the QuotaLimits values is set
func (instance Glance) IsQuotaEnabled() bool {
  return (instance.Spec.Quotas.ImageSizeTotal > 0 ||
      instance.Spec.Quotas.ImageCountTotal > 0 ||
      instance.Spec.Quotas.ImageStageTotal > 0 ||
      instance.Spec.Quotas.ImageCountUpload  > 0)
}
```

We can implement a more strictly check here (e.g., enable only if it matches w/ [1]), but I think we can document this part in the `docs/` section of the operator, and explain how users can enable the feature.
 
[1] https://docs.openstack.org/glance/latest/admin/quotas.html#configuring-glance-for-per-tenant-quotas

config/samples/glance_v1beta1_glance_quota.yaml
fmount: Done

controllers/glance_controller.go

controllers/glance_controller.go

api/bases/glance.openstack.org_glances.yaml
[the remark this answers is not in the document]
konan-abhi: +1

controllers/glance_controller.go
```diff
@@ -766,3 +785,48 @@ func (r *GlanceReconciler) generateServiceConfigMaps(
...
+			Description:  "default limit for  " + lName,
+			ResourceName: lName,
+			DefaultLimit: lValue,
+		}
+		limit, err = o.CreateRegisteredLimit(r.Log, m)
+		if err != nil {
+			return limit, err
+		}
+	}
+	return limit, err
```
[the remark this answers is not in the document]
fmount: Good point, we're not actually checking the result value (other than `err`), so I suspect we can avoid returning the limits.
Updating it.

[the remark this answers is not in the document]
fmount: Done.

konan-abhi (approved):
Looks great now, Thank you!

PR #228, opened 2023-05-19
Rename apiEndpoint key for consistency

kajinamit:
Currently some operators use apiEndpoints(plural) while the other use apiEndpoint(singular). This is quite confusing and requires users to check which key is presented by a operator, to obtain the same information.

This renames the key to the plural, because this status field is a map which can contain multiple endpoints.


kajinamit:
Hmm. I noticed operators exposing APIEndpoints contain service type as top-level key. We probably have to consider adding that as well ?

kajinamit:
We are removing this key because its useless.

PR #233, opened 2023-05-22
Set PVC access mode ReadWriteOnce

SeanMooney:
This changes the PVC access mode from ReadWriteMany
to ReadWriteOnce to allow the default lvmtopo driver
installed in SNO to provide stoage to glance.

When using the local file backend for glance
each pod uses its own pvc and there is no assumtion that
the glance instnace can share data between the pods via
the filesystem. Infact glance assumes that the backing
store used by the file backend is not shared.

### DIFF
--- a/pkg/glance/pvc.go
+++ b/pkg/glance/pvc.go
@@ -17,7 +17,7 @@ func Pvc(api *glancev1.Glance, labels map[string]string) *corev1.PersistentVolum
 		},
 		Spec: corev1.PersistentVolumeClaimSpec{
 			AccessModes: []corev1.PersistentVolumeAccessMode{
-				corev1.ReadWriteMany,
+				corev1.ReadWriteOnce,
 			},
 			Resources: corev1.ResourceRequirements{
 				Requests: corev1.ResourceList{


pkg/glance/pvc.go
```diff
@@ -17,7 +17,7 @@ func Pvc(api *glancev1.Glance, labels map[string]string) *corev1.PersistentVolum
 		},
 		Spec: corev1.PersistentVolumeClaimSpec{
 			AccessModes: []corev1.PersistentVolumeAccessMode{
-				corev1.ReadWriteMany,
+				corev1.ReadWriteOnce,
```
fultonj, line 20: I think this was initially introduced by @abays with this patch. @abays do you recall why you used Many over Once at that time?

https://github.com/openstack-k8s-operators/glance-operator/commit/9a1a2a92b7a8941e24438f0873707624bef69aaf

SeanMooney: no i found that with git blame too but it was there before that change 

https://github.com/openstack-k8s-operators/glance-operator/commit/9a1a2a92b7a8941e24438f0873707624bef69aaf#diff-a42669946878ee7849c63b94c2266a9631dce4edecfc4728d3811bdc8d3f93d0L23

it was set to ReadWriteMany by @dprince in https://github.com/openstack-k8s-operators/glance-operator/commit/d3b03ff0e3d779e5158c1c87289178cd2e127ac6


i assume when matt linked to the kolla docs https://github.com/openstack-k8s-operators/glance-operator/pull/4#pullrequestreview-480647352 

they meant this sectionhttps://docs.openstack.org/kolla-ansible/latest/reference/shared-services/glance-guide.html#file-backend

the issue with that approach is assuming that the PVC is backed by network storage.

it does not really make sense to use multiple replicas of glance API using the same PVC
with host path, lvm or local storage.

glance also has a concept of multiple stores in its API.
you are generally expected to use that to replicate image between glance instance if your using the file store.

I'm not sure its safe to have multiple glance API pods share a volume generally.
we may get data curption depending on exactly how its managing the files.
with our current options, the storage will be allocated from a single server so it will not provide any HA so the only reason to have multiple pods sharing the same volume is for scale.

it would be good to get some of the glance folks to review.

i would assume for performance/ha we will want to advise using ceph or swift or cinder instead of the file store so i wonder is there a need to enforce  ReadWriteMany

fmount: Hi Sean,
to be clear, the config that sees glance w/ the shared `PVC` is not something
we're going to support, but it's useful for development purposes because we
don't require people to mount an `NFS` as shared storage [1].
Having `NFS` means that glance thinks that `File` is used, so it has no clues of
the filesystem backend details.
Being that said, moving it back to `ReadWriteOnce` can compromise the way
we're deploying glance today, where multiple pods (e.g. `internal` and `external`)
insist to the same filesystem.
So, unless you are able to elaborate more around the issue that you hit in your
testing/development and you have strong opinions to make us changing the
development workflow, I'm -1 on this change.
Also, I'd like you to provide tests (the CI job just "deploy" and run kuttl, which is
not enough to test the glance functionalities),
For the record, I'm not saying I disagree with all you wrote (in theory each glance
internal/external API can get a diff `PVC`), but I suspect we should set some
context first.

fmount (changes requested):

pkg/glance/pvc.go
konan-abhi: Agree with @fmount here.

SeanMooney: ths issue is simple.

by default singel node openshift when installed with the assisted installer provisions the lvm topology storage operator.

That does not support read write many file access.

as a result unless you provision an additional storage system the default one provided will not allow glance to be used when you are using a single node openshift.

i can provide an additional storage system but SNO is a fully production-supported installation of Openshift and one of the target platforms we intent to support for OpenStack so i see value in making its default storage system work out of the box.

SeanMooney: ok that did work but i coudl just manually create the PV
and bind it to the PVC

kind: PersistentVolume
apiVersion: v1
metadata:
  name: glance
  uid: 6e698bb3-6481-481a-b8b7-a815bbe7ccf0
  resourceVersion: '46778272'
  creationTimestamp: '2023-05-24T08:44:12Z'
  finalizers:
    - kubernetes.io/pv-protection
  managedFields:
    - manager: Mozilla
      operation: Update
      apiVersion: v1
      time: '2023-05-24T08:49:50Z'
      fieldsType: FieldsV1
      fieldsV1:
        'f:spec':
          'f:accessModes': {}
          'f:capacity':
            .: {}
            'f:storage': {}
          'f:claimRef':
            .: {}
            'f:apiVersion': {}
            'f:kind': {}
            'f:name': {}
            'f:namespace': {}
            'f:resourceVersion': {}
            'f:uid': {}
          'f:csi':
            .: {}
            'f:driver': {}
            'f:fsType': {}
            'f:volumeHandle': {}
          'f:persistentVolumeReclaimPolicy': {}
          'f:storageClassName': {}
          'f:volumeMode': {}
    - manager: kube-controller-manager
      operation: Update
      apiVersion: v1
      time: '2023-05-24T08:49:50Z'
      fieldsType: FieldsV1
      fieldsV1:
        'f:status':
          'f:phase': {}
      subresource: status
spec:
  capacity:
    storage: 10Gi
  csi:
    driver: topolvm.io
    volumeHandle: 176bc3d8-aa17-4b0a-98ef-25b84887b542
    fsType: xfs
  accessModes:
    - ReadWriteMany
  claimRef:
    kind: PersistentVolumeClaim
    namespace: openstack
    name: glance
    uid: 02713dc5-c63b-48a2-bd59-93ce8dd2c9cd
    apiVersion: v1
    resourceVersion: '43749785'
  persistentVolumeReclaimPolicy: Recycle
  storageClassName: lvms-vg1
  volumeMode: Filesystem
status:
  phase: Bound

the lvm storage operator will not automatically provision a volume with ReadWriteMany but you can do it manually

SeanMooney: or not Warning  FailedMount       90s (x11 over 7m42s)   kubelet            MountVolume.SetUp failed for volume "glance" : rpc error: code = FailedPrecondition desc = unsupported access mode: MULTI_NODE_MULTI_WRITER (5)

ill just drop this for now and install a different storage backend

fmount:
Hey @SeanMooney thanks for all the input in the patch and I'll make sure to review it properly and
follow up.
Not that I'm against your argument on the `ReadWriteOnce`, and if you look at [1] it proves that I see
your point; I had similar issues in the past, especially when you're not deploying it w/ crc but SNO
is used.
I think this is the result of some technical debt we have in the operator that should be smart enough at
some point to provide the right configuration and it's something we're going to prioritize next quarter.

[1] https://github.com/openstack-k8s-operators/install_yamls/pull/23

PR #238, opened 2023-05-31
Rename status field name for ready count

kajinamit:
The current name does not explain well these are counts for glanceAPI(Internal|External) but can give a wrong expression that internal and external represent how these counts are used.

This fixes the naming to make it more explicit that these are associated with glanceAPI(Internal|External).


abays (approved):
/lgtm

PR #240, opened 2023-06-02
Fix CRD copy in Makefile

gibizer:
In prow I see:

```
 /go/src/github.com/openstack-k8s-operators/glance-operator/bin/controller-gen rbac:roleName=manager-role crd:maxDescLen=0 webhook paths="./..." output:crd:artifacts:config=config/crd/bases && \
 rm -f api/bases/* && cp -a config/crd/bases api/
 cp: preserving times for 'api/bases': Operation not permitted
```

So this patch makes sure that we can copy the whole directory.


fmount (approved):

fmount:
@gibizer Even though the fix makes sense to me, FWIW I've never seen this error w/ other operators (in manila we run functional tests for instance).

gibizer:
> @gibizer Even though the fix makes sense to me, FWIW I've never seen this error w/ other operators (in manila we run functional tests for instance).
We might have difference in the prow job definition as I only saw this is prow where the permission in the image used for the test is complicated. :)

PR #242, opened 2023-06-05
Remove endpoint url from status

kajinamit:
Endpoint urls in status are not really used Endpoints can be discovered using Keystone.
    
This keeps ServiceIDs dislike the other operators, because glance operator now needs that information to ensure the keystone resource limit is deleted.


kajinamit:
I've been removed from oko org so this should be proposed by someone else.

PR #244, opened 2023-06-07
Name resources according to instance.Name

kajinamit:
... instead of using static names. This allows us to make naming more consistent across operators. Although we don't support deploying multiple control planes in a single name space, we have to deploy multiple instances of some CR(eg GlanceAPI) in the single namespace to deploy a single control plane, so instance.Name is the preferred method to name resources. This does not care about the naming conflicts between CRs in case different CRs are deployed with the same name, because adding prefix causes redundant names. (for example if a user deploys a KeystoneAPI instance with name keystone, we prefer creating a deployment with name keystone instead of keystone-keystone). To avoid such collisions in sub CRs this ensures the top service CR adds proper suffixes when creating sub CR instances.

This also fixes inconsistency in naming related to GlanceAPI CR (mainly about glance-external-api vs glance-api-external).

Also, services and endpoints are still using static names because of the current implementations in lib-common.

### DIFF
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -529,7 +529,7 @@ func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glance
 	//
 
 	serviceLabels := map[string]string{
-		common.AppSelector: glance.ServiceName,
+		common.AppSelector: instance.Name,
 	}
 
 	// networks to attach to
@@ -674,7 +674,7 @@ func (r *GlanceReconciler) apiDeploymentCreateOrUpdate(instance *glancev1.Glance
 
 	deployment := &glancev1.GlanceAPI{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-%s", instance.Name, apiType),
+			Name:      fmt.Sprintf("%s-api-%s", instance.Name, apiType),
 			Namespace: instance.Namespace,
 		},
 	}
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -446,7 +446,7 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
 	//
 
 	serviceLabels := map[string]string{
-		common.AppSelector: fmt.Sprintf("%s-%s", glance.ServiceName, instance.Spec.APIType),
+		common.AppSelector: instance.Name,
 	}
 
 	// networks to attach to
--- a/pkg/glanceapi/deployment.go
+++ b/pkg/glanceapi/deployment.go
@@ -16,8 +16,6 @@ limitations under the License.
 package glanceapi
 
 import (
-	"fmt"
-
 	glancev1 "github.com/openstack-k8s-operators/glance-operator/api/v1beta1"
 	glance "github.com/openstack-k8s-operators/glance-operator/pkg/glance"
 	common "github.com/openstack-k8s-operators/lib-common/modules/common"
@@ -100,7 +98,7 @@ func Deployment(
 
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-api", instance.Name),
+			Name:      instance.Name,
 			Namespace: instance.Namespace,
 		},
 		Spec: appsv1.DeploymentSpec{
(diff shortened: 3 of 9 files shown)


controllers/glance_controller.go
```diff
@@ -529,7 +529,7 @@ func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glance
 	//
 
 	serviceLabels := map[string]string{
-		common.AppSelector: glance.ServiceName,
+		common.AppSelector: instance.Name,
```
fmount, line 532: We don't support multiple `Glances` (the top level resource) in the same namespace, hence -1 here. 
`glance.ServiceName` looks correct as it points to the associated const [1]

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/pkg/glance/const.go#L24

controllers/glanceapi_controller.go
```diff
@@ -446,7 +446,7 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
 	//
 
 	serviceLabels := map[string]string{
-		common.AppSelector: fmt.Sprintf("%s-%s", glance.ServiceName, instance.Spec.APIType),
+		common.AppSelector: instance.Name,
```
fmount, line 449: I think we have a pattern where we are creating across the operators something like:

```
serviceLabels := map[string]string{
		common.AppSelector: glance.ServiceName
		common.ComponentSelector: glance.APIComponent
```

where `APIComponent` should be a `const` like `ServiceName`. 


This significantly improves the ability to search for all resources
owned by an operator, or one of its components:

  oc get all -l service=glance
  oc get all -l component=glance-api

"oc get all" doesn't list Secrets or ConfigMaps, but these can also
be queried.

  oc get secret -l service=glance
  oc get cm -l component=glance-api
 
So I'd like to see something like that instead of using `instance.Name` that can be either internal or external (and it will fail if tomorrow we would like to run an arbitrary numbers of `GlanceAPI` instances, which is in the roadmap).

controllers/glance_controller.go
```diff
@@ -529,7 +529,7 @@ func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glance
 	//
 
 	serviceLabels := map[string]string{
```
fmount, line 531: Define it before and pass it to the `GenerateConfigMap` function

pkg/glanceapi/deployment.go
```diff
@@ -155,7 +153,7 @@ func Deployment(
 	deployment.Spec.Template.Spec.Affinity = affinity.DistributePods(
 		common.AppSelector,
 		[]string{
-			glance.ServiceName,
+			labels[common.AppSelector],
```
fmount, line 156: -1: decorate the `ObjectMeta` passing:

`Labels: labels` which is already available in the deployment.

pkg/glanceapi/deployment.go
```diff
@@ -100,7 +98,7 @@ func Deployment(
 
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-api", instance.Name),
+			Name:      instance.Name,
```
fmount, line 101: I'm assuming this is already `%s-api` because you moved the logic in the controller.

fmount (changes requested):

controllers/glance_controller.go
kajinamit: The main concern behind this is that we are naming resources inconsistently. I agree we do not support multiple `Glance`s in the same name space. Then why do we name some resources such as database or config map according to `instance.Name` ?

I've spent more time looking into the other operators and I later noticed there are some logics not accepting multiple keystoneapi resources in a single namespace... then probably a better solution here is to implement the validation to avoid deployment of multiple CRs in a single namespace(unless that is required), and use the static names consistently across the all resources.

controllers/glanceapi_controller.go
kajinamit: I was using the single label because we have to pass this to `affinity.DistributePods` which only accepts a single label. Probably we have to fix that first to also accept ComponentSelector globally ?

kajinamit: Submitted https://github.com/openstack-k8s-operators/lib-common/pull/283

pkg/glanceapi/deployment.go
kajinamit: Sorry but I did not get the suggestion. Do you mind providing the snippet so that I can understand how the code should be fixed ?

PR #248, opened 2023-06-12
Get System wide admin OSClient when registering quota limits

fmount:
When the `quotas` struct is passed to the main `Glance CR`, the operator needs a system scoped token to make sure the request is run successfully. This change allows the glance-operator to not rely on the keystone operator and build the required `OSClient` based on the new `AuthOps` `System: true` parameter.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/284


konan-abhi (approved):
Looks Good!

PR #249, opened 2023-06-13
Add rootwrap files for glance cinder store

konan-abhi:
This PR add default rootwrap.conf and glance_cinder_store.filters file which are required if we need to configure cinder as a glance backend.

### DIFF
--- a/templates/glance/config/glance-api-config.json
+++ b/templates/glance/config/glance-api-config.json
@@ -24,6 +24,18 @@
         "dest": "/etc/glance/logging.conf",
         "owner": "glance",
         "perm": "0600"
+      },
+      {
+        "source": "/var/lib/config-data/merged/rootwrap.conf",
+        "dest": "/etc/glance/rootwrap.conf",
+        "owner": "glance",
+        "perm": "0600"
+      },
+      {
+        "source": "/var/lib/config-data/merged/rootwrap.d/glance_cinder_store.filters",
+        "dest": "/etc/glance/rootwrap.d/glance_cinder_store.filters",
+        "owner": "root",
+        "perm": "0600"
       }
     ]
 }
--- a/templates/glance/config/rootwrap.conf
+++ b/templates/glance/config/rootwrap.conf
@@ -1,0 +1,27 @@
+# Configuration for glance-rootwrap
+# This file should be owned by (and only-writable by) the root user
+
+[DEFAULT]
+# List of directories to load filter definitions from (separated by ',').
+# These directories MUST all be only writeable by root !
+filters_path=/etc/glance/rootwrap.d,/usr/share/glance/rootwrap
+
+# List of directories to search executables in, in case filters do not
+# explicitely specify a full path (separated by ',')
+# If not specified, defaults to system PATH environment variable.
+# These directories MUST all be only writeable by root !
+exec_dirs=/sbin,/usr/sbin,/bin,/usr/bin,/usr/local/bin,/usr/local/sbin
+
+# Enable logging to syslog
+# Default value is False
+use_syslog=False
+
+# Which syslog facility to use.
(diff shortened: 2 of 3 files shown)


kajinamit:
I might be wrong but aren't these files installed by the glance/glance_store package ?

https://github.com/rdo-packages/glance-distgit/blob/rpm-master/openstack-glance.spec#L290
https://github.com/rdo-packages/glance_store-distgit/blob/rpm-master/python-glance-store.spec#L76

fmount:
> I might be wrong but aren't these files installed by the glance/glance_store package ?
> 
> https://github.com/rdo-packages/glance-distgit/blob/rpm-master/openstack-glance.spec#L290 https://github.com/rdo-packages/glance_store-distgit/blob/rpm-master/python-glance-store.spec#L76

Thanks Takashi for your comment.
I can confirm that content is already present in the container image:

sh-5.1# find / -type d -iname "*rootwrap*"
```
/etc/glance/rootwrap.d
/usr/lib/python3.9/site-packages/oslo.rootwrap-6.3.1-py3.9.egg-info
/usr/lib/python3.9/site-packages/oslo_rootwrap
/usr/share/os-brick/rootwrap
```

sh-5.1# find / -type f -iname "*rootwrap*"
```
/etc/glance/rootwrap.conf
/usr/bin/glance-rootwrap
/usr/bin/oslo-rootwrap
/usr/bin/oslo-rootwrap-daemon
/usr/lib/python3.9/site-packages/os_brick/privileged/__pycache__/rootwrap.cpython-39.opt-1.pyc
/usr/lib/python3.9/site-packages/os_brick/privileged/__pycache__/rootwrap.cpython-39.pyc
/usr/lib/python3.9/site-packages/os_brick/privileged/rootwrap.py
/usr/share/openstack-tripleo-common-containers/container-images/kolla/swift-base/swift-rootwrap
```
sh-5.1# cat /etc/glance/rootwrap.conf
```
# Configuration for glance-rootwrap
# This file should be owned by (and only-writable by) the root user

[DEFAULT]
# List of directories to load filter definitions from (separated by ',').
# These directories MUST all be only writeable by root !
filters_path=/etc/glance/rootwrap.d,/usr/share/glance/rootwrap

# List of directories to search executables in, in case filters do not
# explicitely specify a full path (separated by ',')
# If not specified, defaults to system PATH environment variable.
# These directories MUST all be only writeable by root !
exec_dirs=/sbin,/usr/sbin,/bin,/usr/bin

# Enable logging to syslog
# Default value is False
use_syslog=False

# Which syslog facility to use.
# Valid values include auth, authpriv, syslog, local0, local1...
# Default value is 'syslog'
syslog_log_facility=syslog

# Which messages to log.
# INFO means log all usage
# ERROR means only log unsuccessful attempts
syslog_log_level=ERROR
```

sh-5.1# cat /etc/glance/rootwrap.d/glance_cinder_store.filters
```
# glance-rootwrap command filters for glance cinder store
# This file should be owned by (and only-writable by) the root user

[Filters]
# cinder store driver
disk_chown: RegExpFilter, chown, root, chown, \d+, /dev/(?!.*/\.\.).*

# os-brick library commands
# os_brick.privileged.run_as_root oslo.privsep context
# This line ties the superuser privs with the config files, context name,
# and (implicitly) the actual python code invoked.
privsep-rootwrap: RegExpFilter, privsep-helper, root, privsep-helper, --config-file, /etc/(?!\.\.).*, --privsep_context, os_brick.privileged.default, --privsep_sock_path, /tmp/.*

chown: CommandFilter, chown, root
mount: CommandFilter, mount, root
umount: CommandFilter, umount, rootsh-5.1#
```

konan-abhi:
Closing the PR as it is not required any more.

Thank you @kajinamit and @fmount for confirming the same.

PR #251, opened 2023-06-16
Update ginkgo in the operator api and fix precommit-check

fmount:
The previous patch missed this update in the api, and that let the pre-commit check fail on other PRs.


konan-abhi (approved):
Looks goo!

PR #256, opened 2023-06-20
Cinder backend sample

konan-abhi:
This change provides the Cinder backend sample file to guide people when cinder is
configured as Glance backend.


fmount:
will merge as long as the CI is green

fmount (approved):

PR #257, opened 2023-06-22
Update and Delete RegisteredLimits when Quota is defined

fmount:
This patch relies on the lib-common functions to properly createOrUpdate limits when glance CR is applied, as well as cleaning them up during the delete phase.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/287


controllers/glance_controller.go
```diff
@@ -822,7 +824,38 @@ func (r *GlanceReconciler) ensureRegisteredLimits(
 			ResourceName: lName,
 			DefaultLimit: lValue,
 		}
-		_, err = o.CreateRegisteredLimit(r.Log, m)
+		_, err = o.CreateOrUpdateRegisteredLimit(r.Log, m)
```
fmount, line 827: I tested this call and it works properly in my local environment: I'm able to update existing limits in keystone.

controllers/glance_controller.go
```diff
@@ -822,7 +824,38 @@ func (r *GlanceReconciler) ensureRegisteredLimits(
...
+	keystoneAPI, err := keystonev1.GetKeystoneAPI(ctx, h, instance.Namespace, map[string]string{})
+
+	if err != nil {
+		return err
+	}
+	o, _, err := glancev1.GetAdminServiceClient(ctx, h, keystoneAPI)
+	if err != nil {
+		return err
+	}
+	//TODO: This doesn't work properly atm as we have to wrap it further because
```
fmount, line 854: @konan-abhi this is a WIP here: I still have to do some work to properly wrap the deletion into something that, starting from the `resourceName` makes the mapping to the `resourceID` and call the Delete function as per L858.

konan-abhi: I think here you can list all the resources based on serviceid (ListRegisteredLimitsByServiceID) or get particular resource using (ListRegisteredLimitsByResourceName) method and then call the further deletion method over it.

controllers/glance_controller.go
```diff
@@ -822,7 +823,39 @@ func (r *GlanceReconciler) ensureRegisteredLimits(
...
+	keystoneAPI, err := keystonev1.GetKeystoneAPI(ctx, h, instance.Namespace, map[string]string{})
+
+	if err != nil {
+		return err
+	}
+	o, _, err := glancev1.GetAdminServiceClient(ctx, h, keystoneAPI)
+	if err != nil {
+		return err
+	}
+	fetchRegLimits, err := o.ListRegisteredLimitsByServiceID(r.Log, instance.Status.ServiceID)
```
fmount, line 853: @konan-abhi I'm able to fetch the limits as you suggested, get the ID (see L858) and call the delete.

PR #263, opened 2023-06-28
Use sidecar container to log

fmount:
According to [1] looks like it's a good practice introducing a sidecar container for logging purposes. By doing this we can rotate, process and even forward logs elsewhere without affecting the original service pod.

[1] https://kubernetes.io/docs/concepts/cluster-administration/logging/


fmount:
we might need to adjust kuttl tests before moving forward with this patch.

konan-abhi (approved):
Looks good, verified the results in tempest!

PR #266, opened 2023-07-05
Do not update mergo module

fmount:
imdario/mergo is an indirect dependency that let our renovate patches fail. The module has been recently moved to a vanity url, and this caused a lot of issues w/ renovate.
Removing it here as an explicit dependency as it will be updated as part of other module updates.


konan-abhi (approved):
Looks good!

PR #269, opened 2023-07-10
Watch NADs

abays:
Because we're using them in the reconcile logic for `GlanceAPI`, we should be watching `NetworkAttachmentDefinition` resources


fmount (approved):
Looks good and makes sense triggering a reconcile if any change occurs in the `NAD`.
Thanks @abays

PR #270, opened 2023-07-11
Update test helper references after reorg

gibizer:
Depends-On: openstack-k8s-operators/lib-common#301


stuggi (approved):
/lgtm

stuggi:
I think it also needs a change here https://github.com/openstack-k8s-operators/glance-operator/blob/main/tests/functional/suite_test.go#L33

gibizer:
> I think it also needs a change here https://github.com/openstack-k8s-operators/glance-operator/blob/main/tests/functional/suite_test.go#L33

you are right. It is strange that neither go mod tidy nor the build detected that we import a nonexistent module. Fixed now.

stuggi (approved):
/lgtm

PR #282, opened 2023-07-27
Remove duplicate jobs from github workflow

kajinamit:
We now run pre-commit job in prow which covers most of go-lang tasks.

This also adds a github action that checks for go.mod replace, which was added to several operator repos.


abays (approved):
/lgtm

PR #286, opened 2023-08-01
Bump k8s version in envtest

kajinamit:
... to the one consistent with the version in dependencies.


fmount (approved):

kajinamit:
I've been removed from oko org so this should be proposed by someone else.

PR #292, opened 2023-08-10
Sample files validation through envTest

fmount:
This patch improves the test area of the `Glance` operator. Sample files are loaded from the main directory and a `Glance` `CR` is created. The goal is to validate the syntax provided by the samples stored in the operator project.

### DIFF
--- a/config/samples/glance_v1beta1_glanceapi.yaml
+++ b/config/samples/glance_v1beta1_glanceapi.yaml
@@ -4,6 +4,7 @@ metadata:
   name: glanceapi
 spec:
   serviceUser: glance
+  serviceAccount: glance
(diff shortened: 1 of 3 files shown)


konan-abhi (approved):
Looks good to me, can you add somewhere in commit that atm it is not testing backend and import plugin samples from config/samples directory?

fmount:
> Looks good to me, can you add somewhere in commit that atm it is not testing backend and import plugin samples from config/samples directory?

We could, but basically we're not going to validate those files because they are supposed to deploy  an `OpenStack` ctlplane instead of a `Glance CR`. Before extending this test to the `backends/` directory, we should add Glance specific sample CRs w/ each backend. Something we can follow up in the future (less priority I'd say).

PR #295, opened 2023-08-20
Rework glance config process

fmount:
This patch represents an implementation of the proposal [1] and aligns the `glance-operator` with the work already done in `Cinder` and the other operators.
There are a few relevant changes in the bootstrap process of `Glance`, in particular:

- It stops using an `InitContainer` to generate the snippet files that configure each `GlanceAPI` service. The logic that was previously implemented in the `InitContainer` has been moved to the `Controllers`, where config files are generated and stored in k8s `Secrets`. `InitContainers` are fully removed from the bootstrap process;
- It uses `Kolla` to copy the generated content to the target directory `/etc/glance/glance.conf.d`, which is used by each `API` service to run;
- The relevant content, including scripts, previously stored in a `ConfigMap`, are now stored in a corresponding k8s `Secret`, which is mounted to the `GlanceAPI` deployment `Pods` (hence no additional calls to `GetVolume`/`GetVolumeMounts` are required);
- A total of 4 config snippet files are generated:
  - `00-config.conf` contains global settings that are common to every `GlanceAPI` Pod, including ones that are derived from deployment secrets (e.g. database password, etc.)
  - `01-config.conf` contains the global `customServiceConfig` settings that apply to every GlanceAPI service.
  - `02-config.conf` contains the `customServiceConfig` settings that are specific to each service.
  - `03-config.conf` contains secrets specified by each service's `customServiceConfigSecrets`.
- `glance-image-import.conf` has been removed, and the default config has been added to `00-config.conf`, which is the common file generated by the umbrella `Glance` controller;
- `logging.conf` has been removed as it's no longer required in the switch to a side container approach for logging purposes;
- functional tests are aligned to the use of k8s `Secrets` instead of the old pattern based on `ConfigMaps`;
- `kuttl` tests are updated and the `initContainer` has been removed.
- `DBsync` now mounts only the required files (a minimal `00-config.conf`) and a `db-sync-config.json` containing the command run through `kolla`.

[1] https://github.com/openstack-k8s-operators/docs/pull/31


templates/glance/config/00-config.conf
```diff
@@ -76,3 +91,6 @@ endpoint_id = {{ .EndpointID }}
 {{ if (index . "Region") }}
 region_name = {{ .Region }}
 {{ end }}
+
+[image_import_opts]
```
fmount, line 95: @konan-abhi is this correct or do we need a mandatory `glance-image-import.conf`? Keep in mind this config can be overridden by `customConfigService`

konan-abhi: There is no mandatory need of image-import.conf file. But while documenting import workflow we need to point to correct config files in the documentation.

templates/glance/config/00-config.conf
```diff
@@ -12,12 +12,18 @@ bind_host=0.0.0.0
 bind_port=9292
 workers=3
 image_cache_dir=/var/lib/glance/image-cache
-log_config_append=/etc/glance/logging.conf
```
fmount, line 15: @konan-abhi removing this line as we're going to go with https://github.com/openstack-k8s-operators/glance-operator/pull/263

templates/glance/config/00-config.conf
fmount: we already documented the way to change this option is via `customConfigService`, I checked [1] and there's no mention of the default `no_op` we added in `glance-image-import.conf`, so it should still be consistent with what we documented.
We can improve that part in a follow up change if you think we need more changes in terms of doc.
Thanks for confirming that section can stay in `00-config.conf`.

[1] https://github.com/openstack-k8s-operators/glance-operator/tree/main/config/samples/import_plugins

templates/glance/config/00-config.conf
konan-abhi: Sounds good

templates/glance/config/glance-api-config.json
```diff
@@ -2,28 +2,31 @@
...
+        "source": "/var/lib/config-data/default/00-config.conf",
+        "dest": "/etc/glance/glance.conf.d/00-config.conf",
         "owner": "glance",
-        "perm": "0750"
+        "perm": "0600"
       },
       {
-        "source": "/var/lib/config-data/merged/glance-image-import.conf",
-        "dest": "/etc/glance/glance.conf.d/glance-image-import.conf",
+        "source": "/var/lib/config-data/default/02-config.conf",
```
konan-abhi, line 11: have we missed 01-config.conf?

fmount: it's basically 01 in the umbrella glance, but 02 in the glance-api service, because the same config is passed to the instances created by the top-level CR [1].
We can improve that in a follow up change, or rename 02-config to 01-config: the goal is keeping consistency with the naming across the board, and don't expected a change in their meaning.
I noted this as a potential follow up! How does that sound?

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/controllers/glance_controller.go#L711

fmount:
@konan-abhi worked as expected: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_glance-operator/295/pull-ci-openstack-k8s-operators-glance-operator-main-glance-operator-build-deploy-kuttl/1693866122480193536/artifacts/glance-operator-build-deploy-kuttl/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-4bc3ccfa15317dc06b06988da2227028fb4c03aee8d52f05f0dfe4d2640d72aa/namespaces/openstack/secrets/glance/

konan-abhi (approved):
Thank you, verified and satisfied with the change.

PR #296, opened 2023-08-22
Do not ship policy.yaml by default

fmount:
`policy.yaml` is stored in a secret at the moment: it can't be changed by any human operator because `glance-operator` will restore it and it grows a the size of the `GlanceAPI` secret.
This patch removes it from the `Secret`, `kolla` won't sync it by default but it can still be provided via `extraVolumes`, and the `oslo_policy` section can be injected via `customConfigService` overrides.


config/samples/policy/README.md
```diff
@@ -0,0 +1,56 @@
...
+[example](https://github.com/openstack-k8s-operators/glance-operator/tree/main/config/samples/glance_v1beta_glance_apply_policy.yaml),
+when the `ConfigMap` is available, a `GlanceAPI` CR can be applied injecting
+the configuration that will point to the new `policy.yaml` file.
+
+```
+spec:
+  ...
+  customServiceConfig: |
+    [oslo_policy]
+    policy_file=/etc/glance/policy.d/policy.yaml
```
konan-abhi, line 31: Shall we also mention about 'enforce_scope' and 'enforce_new_defaults' parameters?
Those are set to true if SRBAC needs to be enabled.

https://github.com/openstack/oslo.policy/blob/master/oslo_policy/opts.py#L27

konan-abhi (commented):
One suggestion, otherwise looks perfect

config/samples/policy/README.md
fmount: of course, good catch! updating the patch in a bit!

konan-abhi (approved):
Looks perfect, thank you!

PR #301, opened 2023-09-01
Rename the tests directory

Akrog:
As per the standard golang proyect layout [1] the directory with the tests should be called `test` and not `tests`, so we rename it here.

We should be consistent in all openstack operators repositories and follow the standard so the documentation has no discrepancies [2].

[1]: https://github.com/golang-standards/project-layout
[2]: https://github.com/openstack-k8s-operators/docs/pull/58/files#r1311828677


fmount:
> As per the standard golang proyect layout [1](https://github.com/golang-standards/project-layout) the directory with the tests should be called `test` and not `tests`, so we rename it here.
> 
> We should be consistent in all openstack operators repositories and follow the standard so the documentation has no discrepancies [2](https://github.com/openstack-k8s-operators/docs/pull/58/files#r1311828677).

Ok, despite most of the other projects call this directory `tests`, I agree we should follow the standard project layout you mentioned, hence +1 on this change.

fmount (approved):

PR #307, opened 2023-09-06
Create -scripts as top-level resource

fmount:
`GlanceAPIs` have no reasons to regenerate the `-scripts` mounted to the `Pods`. The scripts represent a resource that can be inherited from the top level `CR`.
This patch let the `glance_controller` generate this resource: it will be mounted (if `Cinder` is used as a backend) by all the `GlanceAPI` pods.

### DIFF
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -762,8 +762,8 @@ func (r *GlanceReconciler) generateServiceConfig(
 	for key, data := range instance.Spec.DefaultConfigOverwrite {
 		customData[key] = data
 	}
-
-	return GenerateConfigsGeneric(ctx, h, instance, envVars, templateParameters, customData, labels, false)
+	// Generate both default 00-config.conf and -scripts
+	return GenerateConfigsGeneric(ctx, h, instance, envVars, templateParameters, customData, labels, true)
 
 }
 
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -693,8 +693,8 @@ func (r *GlanceAPIReconciler) generateServiceConfig(
 
 	// 00-default.conf will be regenerated as we have a ln -s of the
 	// templates/glance/config directory
-	// TODO: Make the scripts be a top level secret since they are shared for all the glanceapis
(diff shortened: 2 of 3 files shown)


fmount:
This is the result of my local test where `Cinder` is set as `Glance` backend.

```
[stack@osp-storage-05 glance-operator]$ oc rsh -c glance-api glance-external-api-77865cf879-x4fr5
sh-5.1# ls /usr/local/bin/
container-scripts/  kolla_extend_start  kolla_httpd_setup   kolla_set_configs   kolla_start         uid_gid_manage
sh-5.1# ls /usr/local/bin/container-scripts/
run-on-host
```

`oc describe pod glance-external-api-77865cf879-x4fr5`  shows:

```
    Mounts:
      /dev from dev (rw)
      /etc/iscsi from etc-iscsi (ro)
      /etc/nvme from etc-nvme (rw)
      /lib/modules from lib-modules (ro)
      /run from run (rw)
      /sys from sys (rw)
      /usr/local/bin/container-scripts from scripts (ro)
      /var/lib/config-data/default from config-data (ro)
      /var/lib/config-data/secret-0 from osp-glance-cinder-secret (ro)
      /var/lib/glance from lib-data (rw)
      /var/lib/kolla/config_files/config.json from config-data (ro,path="glance-api-config.json")
      /var/locks/openstack/os-brick from var-locks-brick (rw)
      /var/log/glance from logs (rw)
```

and

```
  scripts:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  glance-scripts
    Optional:    false
```

including the `/usr/local/bin/container-scripts from scripts (ro)` (same story for the `-internal` pod).

`oc get secrets | grep glance` returns:

```
glance-config-data
glance-external-config-data
glance-internal-config-data
glance-scripts
osp-glance-cinder-secret
```
where we can see `glance-scripts` only once (and it's inherited by the other Pods).

Akrog (approved):
/lgtm

fmount:
I suspect we have a flaky test when Glance is loaded from sample file. I wasn't able to reproduce but I'll try to work on it in a separated patch.

PR #308, opened 2023-09-06
Move Quota samples in a dedicated section

fmount:
This patch adds the quota samples in a dedicated samples section; a README is provided and a Glance CR describes where the fields are supposed to be added in the spec.

### DIFF
--- a/test/functional/sample_test.go
+++ b/test/functional/sample_test.go
@@ -56,7 +56,6 @@ func CreateGlanceAPIFromSample(sampleFileName string, name types.NamespacedName)
 // file has all the required field with proper types. But it does not
 // validate that using a sample file will result in a working deployment.
 var _ = Describe("Samples", func() {
-
 	When("glance_v1beta1_glance.yaml sample is applied", func() {
 		It("Glance is created", func() {
 			Eventually(func(g Gomega) {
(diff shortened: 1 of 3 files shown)


test/functional/sample_test.go
```diff
@@ -70,12 +69,6 @@ var _ = Describe("Samples", func() {
 			}, timeout, interval).Should(Succeed())
 		})
 	})
-	When("glance_v1beta1_glance_quota.yaml sample is applied", func() {
```
fmount, line 73: @abays this is causing your patch to fail envTest (it's an intermittent issue).

fmount:
let's hold this patch as we're investigating how CI does with this flaky test

konan-abhi (approved):
Looks good!

test/functional/sample_test.go
```diff
@@ -72,7 +71,7 @@ var _ = Describe("Samples", func() {
 	})
 	When("glance_v1beta1_glance_quota.yaml sample is applied", func() {
 		It("Glance is created", func() {
-			name := CreateGlanceFromSample("glance_v1beta1_glance_quota.yaml", glanceTest.Instance)
+			name := CreateGlanceFromSample("quotas/glance_v1beta1_glance_quota.yaml", glanceTest.Instance)
```
fmount, line 74: @abays I fixed the envTest by preventing deletingLimits if no keystone CR is present in the ctlplane, and this actually solved the func test. I'm just moving the sample to a dedicated directory, and reference it from here (it can be done for any other sample file instead of keeping them in the root sample directory).
If it's ok w/ you (given @konan-abhi is off), you can re-apply the lgtm label and merge this.

PR #311, opened 2023-09-15
Always recalculate deployment status

abays:
We've noticed that certain operators are not considering the status of their associated `Deployment` after initially succeeding.  This needs to always be reconsidered in case if the `Deployment` is having problem.

This change follows prior art from Keystone operator [1].

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/main/controllers/keystoneapi_controller.go#L827-L835

### DIFF
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -606,8 +606,14 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
 		return ctrl.Result{}, err
 	}
 
-	if instance.Status.ReadyCount > 0 {
+	if instance.Status.ReadyCount > 0 || *instance.Spec.Replicas == 0 {
 		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
+	} else {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.DeploymentReadyCondition,
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			condition.DeploymentReadyRunningMessage))
 	}
 	// create Deployment - end
 


controllers/glanceapi_controller.go
```diff
@@ -608,6 +608,12 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
 
 	if instance.Status.ReadyCount > 0 {
```
gibizer, line 609: There is a slight variant of this check in couple of places[1]. 
If Instance.Spec.Replicas is set to 0  by the user then ReadyCount will be 0 and the GlanceAPI CR won't be set Ready even though the operator successfully applied the requested Specification to the Deployment resource. 

[1] e.g. https://github.com/openstack-k8s-operators/nova-operator/blob/532bc7f74c7f8893ad9ab53690bed442c843619b/controllers/novametadata_controller.go#L420-L432

gibizer (commented):
We can cover this logic in envtest by simulating that Deployment ReadyReplicas is 0 after GlanceAPI is Ready.

controllers/glanceapi_controller.go
abays: That's a good point.  The user's desired `Replicas` count should also be considered.  I will add that.

fmount: Thanks @gibizer and @abays. It makes sense, and I agree we can easily cover this scenario via envTest.

controllers/glanceapi_controller.go
```diff
@@ -606,8 +606,14 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
 		return ctrl.Result{}, err
 	}
 
-	if instance.Status.ReadyCount > 0 {
+	if instance.Status.ReadyCount > 0 || *instance.Spec.Replicas == 0 {
```
gibizer, line 609: looks good to me.

PR #317, opened 2023-09-18
Fix samples and backends

fmount:
This patch fixes some backends and references used to build a valid patched control plane with Glance.


konan-abhi (approved):
++, Thank you for fixing it.

konan-abhi (approved):

PR #318, opened 2023-09-22
Add httpd sidecontainer for GlanceAPI

fmount:
This patch introduces an additional `sidecontainer` for `GlanceAPI` that is supposed to `ProxyPass` requests to the `glance-api` process behind.
This will also solve the problem of `tls-everywhere` for Glance, where the second segment is `re-encrypted` from the `Route` (created by the `openstack-operator`) to the `Deployment`.
* The `httpd` will serve requests on `9292`, while the `glance-api` service will be run locally on `9293`.
* The `httpd` instance will be customized in a follow up patch to mount certificates and add the related `ssl` directives


konan-abhi (approved):
++ for design decision. Looks good to me.
Thank you!

PR #329, opened 2023-10-02
Single instance glance support

fmount:
This change is part of a series where we want to provide changes to the `glance-operator` to support a list of `GlanceAPI(s)`. 
This patch introduces a new way of deploying `Glance`. It removes the hardcoded `GlanceAPIInternal` and `GlanceAPIExternal` instances from the `Glance` `v1beta1/` API in favor of a single `GlanceAPI` definition that is supposed to handle, behind the scenes, the split of `Glance` in two different `Deployments` (one for internal and one for external).
The new `GlanceAPI` spec looks like the following:

```
kind: Glance
metadata:
  name: glance
spec:
  serviceUser: glance
  databaseInstance: openstack
 ...
  glanceAPI:
    debug:
      service: false
    preserveJobs: false
    replicas: 1
  ...
```

and still follows the same logic and rules previously defined in the operator.
With this change, it is possible to deploy a `single` API instance without splitting (keep in mind this is **not** the default behavior, but it's the first step to reach the bigger goal of deploying a list of `[]GlanceAPI` for `DCN` scenarios).

In case of a single GlanceAPI the main CR looks like the following:

```
kind: Glance
metadata:
  name: glance
spec:
  serviceUser: glance
  ...
  glanceAPI:
    debug:
      service: false
    type: "single"
    preserveJobs: false
    replicas: 1
 ...
```

and, as described in the `design_decisions` section, it will create just a single k8s `Deployment`.
Given that is possible to have a single deployment tied to both `glance-internal` and `glance-public` endpoint, the `APIOverrideSpec` has been changed to support both `public` and `internal` service overrides (a map) instead of a single `*service.RoutedOverrideSpec`, and the change in this logic is reflected in the `openstack-operator` https://github.com/openstack-k8s-operators/openstack-operator/pull/514.

## How to review this change

This patch has of course many moving parts: a change in the `API` has implications over the `CRD` (and the auto-generated code), so I'm going to highlight here the list of the many relevant files that should drive the review:

- ` v1beta1/glance_types.go`: here's where everything started by removing the hardcoded instances in favor of a single glanceAPI definition; we also have an additional `glanceType` called `APISingle` that is the string defining the `glanceAPIType`.

- `v1beta1/glance_webhook.go`: this has also been changed to reflect the new Glance `Spec` that includes `GlanceAPI` instead of the hardcoded Internal/External

- `v1beta1/common.go`: `GlanceAPITemplate`, which is the type used in the previous  `Spec`  definition, now has a `Type` field, which is used to drive the `apiDeployment` (either split or  deploy a single instance). In addition, given a single `glanceAPI` is defined in the main spec, the service overrides is now a `map[service.Endpoint]service.RoutedOverrideSpec`, which is the same pattern applied to all the other operators (manila/cinder/nova/etc ...)

- `config/samples`: All the samples have been update to generate an `openstackcontrolplane` CR that includes the new `Glance` definition, but the interesting thing here is to look at the two sample files in the the root `samples/` directory, one for `split` and the other one for `single` instances.

Both samples are used by:
  - `test/functional/sample_test.go` that performs a sanity check against these Glance CRs,
  - `test/kuttl/glance_{scale, single}` that is now able to run the same kuttl tests against
     the two scenarios (split vs single)

- `controllers/glance_controller`: here's the main place where things have changed. the   `apiDeployment` function implements the logic of splitting the deployment or provide a `single API instance` according to what's defined in the main `CR`. Everything is
backward compatible with the previous definition, however, if `type: single` is passed  (which means `current.Type ==  "single"` within the function), one of the two `Deployments` is skipped, and the `Endpoints` are properly reflected in the main `Glance`
CR `Status`
  
- `controllers/glanceapi_controller`: it mostly updates the endpoints and the service overrides in case we don't split the `API`

- `kuttl/tests`: they are now split between `glance_scale` and `glance_single`, so  we can make sure we cover the `single API` scenario the same way we do for the split API use case

- `test/functional`: `envTests` are now updated to cover the following:
  - new `NAD` definition that includes internal vs external
  - `single glanceAPI` deployment
  - internal/external `glanceAPI` deployment

## How this change is tested

The change in the `GlanceAPI` impacts the `openstack-operator` `CRD` that is used to define a valid `openstackcontrolplane` object required to use the new `GlanceAPI` and have a valid job run for both `kuttl` and `tempest` tests.
The patch https://github.com/openshift/release/pull/44532 in the `Prow` scripts added the ability to checkout a specific `openstack-operator` PR before building it, and thanks to this change we're able to build an `openstack-operator` version that is in line with the current change, and it allows the `kuttl` job to run properly.

- kuttl job
  - [build openstack-operator from PR514](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_glance-operator/329/pull-ci-openstack-k8s-operators-glance-operator-main-glance-operator-build-deploy-kuttl/1716826079433330688/artifacts/glance-operator-build-deploy-kuttl/openstack-k8s-operators-build/build-log.txt)
  - [kuttl execution](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_glance-operator/329/pull-ci-openstack-k8s-operators-glance-operator-main-glance-operator-build-deploy-kuttl/1716826079433330688/artifacts/glance-operator-build-deploy-kuttl/openstack-k8s-operators-build/build-log.txt)

- Also, as can be seen in the kuttl job [1], both scale and single tests are
  executed:

  ```
  === CONT  kuttl
      harness.go:405: run tests finished
      harness.go:513: cleaning up
      harness.go:570: removing temp folder: ""
  --- PASS: kuttl (191.31s)
      --- PASS: kuttl/harness (0.00s)
          --- PASS: kuttl/harness/glance_scale (105.87s)
          --- PASS: kuttl/harness/glance_single (77.14s)
  PASS
  ```

At the same time, the [openstack-operator patch](https://github.com/openstack-k8s-operators/openstack-operator/pull/514) that includes the new `glanceAPI` structure is green because of the `Depends-On` on this patch applied to the Zuul job [2].
Zuul (actually the ci-framework job run through zuul) knows that `Glance` should be built before building the `openstack-operator` bundle, and this one must reference the pending version of the `glance-operator` that is supposed to handle the new API.
The `openstack-operator` Kuttl job [3] is green and reports the Service Ready in the main assertion. 

[1]  [kuttl job](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_glance-operator/329/pull-ci-openstack-k8s-operators-glance-operator-main-glance-operator-build-deploy-kuttl/1716826079433330688/artifacts/glance-operator-build-deploy-kuttl/openstack-k8s-operators-kuttl/build-log.txt)
[2] [GlanceCR in the zuul job](https://logserver.rdoproject.org/14/514/3118988f27b008f080675fdb15410957a68baeba/github-check/podified-multinode-edpm-deployment-crc/4583947/controller/ci-framework-data/logs/crc/crs/glances.yaml)
[3] [OpenStackControlPlaneGlanceReady in the meta operator kuttl job](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_openstack-operator/514/pull-ci-openstack-k8s-operators-openstack-operator-main-openstack-operator-build-deploy-kuttl/1717111437471518720/artifacts/openstack-operator-build-deploy-kuttl/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-9810fe9cf8d0d302413f7876267d8e3149d81e7d4b4c2b88c98f5a60f0c3f800/namespaces/openstack/crs/openstackcontrolplanes.core.openstack.org/openstack-galera.yaml)


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openstack-k8s-operators/glance-operator
-  newTag: latest
+  newName: quay.io/fpantano/glance-operator
```
fmount, line 15: remove this

api/v1beta1/glance_webhook.go
```diff
@@ -59,16 +59,12 @@ func (r *Glance) Default() {
 
 // Default - set defaults for this Glance spec
 func (spec *GlanceSpec) Default() {
-	if spec.ContainerImage == "" {
+	if len(spec.ContainerImage) == 0 {
 		spec.ContainerImage = glanceDefaults.ContainerImageURL
 	}
```
abays, line 64: Did you notice the old form of the if-statement failing to work properly in certain cases?

controllers/glance_controller.go
```diff
@@ -594,92 +594,128 @@ func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glance
...
+	// Mirror single/external GlanceAPI status' APIEndpoints and ReadyCount to this parent CR
 	if glanceAPI.Status.APIEndpoints != nil {
 		instance.Status.APIEndpoints[string(endpoint.EndpointPublic)] = glanceAPI.Status.APIEndpoints[string(endpoint.EndpointPublic)]
+		// if we don't split, both apiEndpoints (public and internal) should be
+		// reflected to the main Glance CR
+		if current.Type == "single" {
+			instance.Status.APIEndpoints[string(endpoint.EndpointInternal)] = glanceAPI.Status.APIEndpoints[string(endpoint.EndpointInternal)]
+		}
 	}
 	instance.Status.GlanceAPIExternalReadyCount = glanceAPI.Status.ReadyCount
```
abays, line 663: Is there any reason we might want to set `instance.Status.GlanceAPIInternalReadyCount` here as well if type is `single`?

api/v1beta1/glance_webhook.go
abays: I'm fine with as you have it now.  Was just curious if you had seen something.

controllers/glance_controller.go
fmount: In theory a `single` instance acts as both `Internal` and `External`, and I though here two things:
1. if type is `single`, setting the `ReadyCount` in the `Status` is something good to have and aligned with the `split` use case (where we set two of them);
2. we can reuse the same variable as we don't really need a different one that should be mirrored in the main CR status, but it's something we can actually do (making this logic a little bit more verbose);

The other option is not setting it at all when instance is `single`, as we only check this info in the `kuttl` assertion, and afaik is not used anywhere to mark the glance condition as Ready.

Do you think that it makes sense to add a `GlanceAPIReadyCount int32 `json:"glanceAPIReadyCount,omitempty"` to [1] ?

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/api/v1beta1/glance_types.go#L155

abays: I think it's fine reusing the variables as you've done for the meantime.  To your point, they are mostly just informative and aren't considered in any calculation.  This was just another case of my curiosity posing a question. :)

controllers/glanceapi_controller.go
```diff
@@ -743,8 +748,9 @@ func (r *GlanceAPIReconciler) generateServiceConfig(
 	}
 
 	// Configure the internal GlanceAPI to provide image location data, and the
-	// external version to *not* provide it.
-	if instance.Spec.APIType == glancev1.APIInternal {
+	// external version to *not* provide it; if we don't split, the resulting
+	// GlanceAPI instance will provide it.
+	if instance.Spec.APIType == glancev1.APIInternal || instance.Spec.APIType == glancev1.APISingle {
```
konan-abhi, line 753: Just curious to know don't we have something similar to "in" function in python here like 'x' in [a,b,c,d] ?

fmount: I don't think so, or at list this can be achieved via the following:

```golang
func checkAPIType(current string, myarray []string) bool {
    for _, b := range myarray.(string) {
        if b == current {
            return true
        }
    }
    return false
}
```
and then I can simply:

```golang
if checkAPIType(instance.Spec.APIType,  []string{string(glancev1.APIInternal), string(glancev1.APISingle)}) {
...
}
 ```
 
but it makes less obvious what we're trying to achieve in that logic, and instead of an additional `||` we're going to add more complexity building an array (on-the-fly) for the purpose of making an iteration.

Comment (konan-abhi) [reply controllers/glanceapi_controller.go:753]: ack, make sense, I thought there could have been inbuilt method for this. May be if worth putting efforts, we can add such small methods later in common modules.

Comment (fmount) [reply controllers/glanceapi_controller.go:753]: +1 on further improvements like that one to improve readability

**Review by konan-abhi — APPROVED** [review]
Looks good me, +1 for later improvements/creating new utility functions

File: controllers/glanceapi_controller.go (hunk 4 above)
Comment (fmount) [reply controllers/glanceapi_controller.go:753]: nice, that's exactly what I had in mind to implement the equivalent of `"string" in array`! Noted for a follow up change.

**Comment by fmount** [comment]
@abays @konan-abhi @stuggi this change is basically ready, I rebased the content to the latest patches (`StatefulSet`, etc) and we have green CI with the `Depends-on: https://github.com/openstack-k8s-operators/openstack-operator/pull/514`.
The current plan is:
1. merge this change
2. rebase [openstack-operator/pull/514 ](https://github.com/openstack-k8s-operators/openstack-operator/pull/514) and remove the `replace` directive, bumping the glance-operator `APIs` in `go.mod` and `apis/go.mod`.
3. merge the `openstack-operator` patch
4. Update https://github.com/openstack-k8s-operators/data-plane-adoption/pull/187 to make sure we don't break the `adoption` use case (/cc @jistr )

I'm not sure I'm overlooking other projects that might need an update, but this is pretty much the list of things that I know I have to change/update.

**Review by konan-abhi — APPROVED** [review]
Looks good to me!

**Comment by abays** [comment]
@fmount One thing I was testing that seems to have trouble is changing the `Glance` (or `OpenStackControlPlane`) CR's `spec.glanceAPI.type` from `split` to `single`.  When doing so, I see this error in the Glance controller:

```
2023-11-13T16:21:46Z	ERROR	Reconciler error	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "GlanceAPI": {"name":"glance-single","namespace":"openstack"}, "namespace": "openstack", "name": "glance-single", "reconcileID": "c5a163cc-8703-494b-963a-43a9c57faa4b", "error": "Object openstack/glance-public is already owned by another GlanceAPI controller glance-external"}
```

**Comment by fmount** [comment]
> @fmount One thing I was testing that seems to have trouble is changing the `Glance` (or `OpenStackControlPlane`) CR's `spec.glanceAPI.type` from `split` to `single`. When doing so, I see this error in the Glance controller:
> 
> ```
> 2023-11-13T16:21:46Z	ERROR	Reconciler error	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "GlanceAPI": {"name":"glance-single","namespace":"openstack"}, "namespace": "openstack", "name": "glance-single", "reconcileID": "c5a163cc-8703-494b-963a-43a9c57faa4b", "error": "Object openstack/glance-public is already owned by another GlanceAPI controller glance-external"}
> ```

This is interesting: the fact that it fails happens because you're trying to change layout on an already deployed `GlanceApi`. Unfortunately this is not something we can do atm because if you change `type`, you still have all the resources deployed (e.g. SVCs) and there's no logic in the operator to delete/terminate the previous deployment (and the associated resources) and rollout the new one.
This means that in order to change the layout of a `GlanceAPI` from the `openstack-operator` we should delete the instances and the associated resources (`enabled: false`) and then redeploy it w/ the new layout.
Changing type from `split` to `single` allows you to create a `glance-single` API, but this is not terminating the previous `StatefulSets`, hence this is a no go for the current implementation.
When we move to a list of `glanceAPIs` (`map[string]GlanceAPI`) it would be easier deleting a specific instance by editing the CR and redeploying, but that doesn't change the fact we can't update the layout.

@abays what do you think? We have two options here:

1. Add a validation (using webhooks?) to prevent the update of the `type` field;
2. Support, within the operator, the update of the `GlanceAPI` layout.

I'm not sure about the best approach here: I would block in the first place the ability to change the layout for an already deployed Glance,  and  dig more into the update topic and POC something as a follow up.

**Comment by abays** [comment]
> > @fmount One thing I was testing that seems to have trouble is changing the `Glance` (or `OpenStackControlPlane`) CR's `spec.glanceAPI.type` from `split` to `single`. When doing so, I see this error in the Glance controller:
> > ```
> > 2023-11-13T16:21:46Z	ERROR	Reconciler error	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "GlanceAPI": {"name":"glance-single","namespace":"openstack"}, "namespace": "openstack", "name": "glance-single", "reconcileID": "c5a163cc-8703-494b-963a-43a9c57faa4b", "error": "Object openstack/glance-public is already owned by another GlanceAPI controller glance-external"}
> > ```
> 
> This is interesting: the fact that it fails happens because you're trying to change layout on an already deployed `GlanceApi`. Unfortunately this is not something we can do atm because if you change `type`, you still have all the resources deployed (e.g. SVCs) and there's no logic in the operator to delete/terminate the previous deployment (and the associated resources) and rollout the new one. This means that in order to change the layout of a `GlanceAPI` from the `openstack-operator` we should delete the instances and the associated resources (`enabled: false`) and then redeploy it w/ the new layout. Changing type from `split` to `single` allows you to create a `glance-single` API, but this is not terminating the previous `StatefulSets`, hence this is a no go for the current implementation. When we move to a list of `glanceAPIs` (`map[string]GlanceAPI`) it would be easier deleting a specific instance by editing the CR and redeploying, but that doesn't change the fact we can't update the layout.
> 
> @abays what do you think? We have two options here:
> 
>     1. Add a validation (using webhooks?) to prevent the update of the `type` field;
> 
>     2. Support, within the operator, the update of the `GlanceAPI` layout.
> 
> 
> I'm not sure about the best approach here: I would block in the first place the ability to change the layout for an already deployed Glance, and dig more into the update topic and POC something as a follow up.

I think validation via a webhook to make the `type` field immutable will suffice for now.  Then we can think about how we might want to handle allowing changing the type in the future, if at all.

**Comment by fmount** [comment]
> > > @fmount One thing I was testing that seems to have trouble is changing the `Glance` (or `OpenStackControlPlane`) CR's `spec.glanceAPI.type` from `split` to `single`. When doing so, I see this error in the Glance controller:
> > > ```
> > > 2023-11-13T16:21:46Z	ERROR	Reconciler error	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "GlanceAPI": {"name":"glance-single","namespace":"openstack"}, "namespace": "openstack", "name": "glance-single", "reconcileID": "c5a163cc-8703-494b-963a-43a9c57faa4b", "error": "Object openstack/glance-public is already owned by another GlanceAPI controller glance-external"}
> > > ```
> > 
> > 
> > This is interesting: the fact that it fails happens because you're trying to change layout on an already deployed `GlanceApi`. Unfortunately this is not something we can do atm because if you change `type`, you still have all the resources deployed (e.g. SVCs) and there's no logic in the operator to delete/terminate the previous deployment (and the associated resources) and rollout the new one. This means that in order to change the layout of a `GlanceAPI` from the `openstack-operator` we should delete the instances and the associated resources (`enabled: false`) and then redeploy it w/ the new layout. Changing type from `split` to `single` allows you to create a `glance-single` API, but this is not terminating the previous `StatefulSets`, hence this is a no go for the current implementation. When we move to a list of `glanceAPIs` (`map[string]GlanceAPI`) it would be easier deleting a specific instance by editing the CR and redeploying, but that doesn't change the fact we can't update the layout.
> > @abays what do you think? We have two options here:
> > ```
> > 1. Add a validation (using webhooks?) to prevent the update of the `type` field;
> > 
> > 2. Support, within the operator, the update of the `GlanceAPI` layout.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm not sure about the best approach here: I would block in the first place the ability to change the layout for an already deployed Glance, and dig more into the update topic and POC something as a follow up.
> 
> I think validation via a webhook to make the `type` field immutable will suffice for now. Then we can think about how we might want to handle allowing changing the type in the future, if at all.

Ok, I think the new patchset has the required change:

```
[stack@osp-storage-04 samples]$ oc patch glance glance --type=json -p="[{'op': 'replace', 'path': '/spec/glanceAPI/type', 'value': 'split'}]"
Error from server (Deployment layout can't be changed): admission webhook "vglance.kb.io" denied the request: Deployment layout can't be changed
```

and on the operator side:

```
2023-11-14T07:26:59-05:00       INFO    controllers.Glance      Reconciled Service 'glance' successfully
2023-11-14T07:27:27-05:00       DEBUG   controller-runtime.webhook.webhooks     received request        {"webhook": "/mutate-glance-openstack-org-v1beta1-glance", "UID": "46c40fc2-a365-430c-a85b-a8b6bb41d22d", "kind": "glance.openstack.org/v1beta1, Kind=Glance", "resource": {"group":"glance.openstack.org","version":"v1beta1","resource":"glances"}}
2023-11-14T07:27:27-05:00       INFO    glance-resource default {"name": "glance"}
2023-11-14T07:27:27-05:00       DEBUG   controller-runtime.webhook.webhooks     wrote response  {"webhook": "/mutate-glance-openstack-org-v1beta1-glance", "code": 200, "reason": "", "UID": "46c40fc2-a365-430c-a85b-a8b6bb41d22d", "allowed": true}
2023-11-14T07:27:27-05:00       DEBUG   controller-runtime.webhook.webhooks     received request        {"webhook": "/validate-glance-openstack-org-v1beta1-glance", "UID": "41d18e26-453c-4ecd-b7bd-ff3dc978c275", "kind": "glance.openstack.org/v1beta1, Kind=Glance", "resource": {"group":"glance.openstack.org","version":"v1beta1","resource":"glances"}}
2023-11-14T07:27:27-05:00       INFO    glance-resource validate update {"name": "glance"}
2023-11-14T07:27:27-05:00       INFO    glance-resource Deployment Type - new   {"Spec.Type": "split"}
2023-11-14T07:27:27-05:00       INFO    glance-resource Deployment Type - old   {"Spec.Type": "single"}
2023-11-14T07:27:27-05:00       INFO    glance-resource Errro Type - new        {"Spec.Type": "split"}
2023-11-14T07:27:27-05:00       DEBUG   controller-runtime.webhook.webhooks     wrote response  {"webhook": "/validate-glance-openstack-org-v1beta1-glance", "code": 403, "reason": "Deployment layout can't be changed", "UID": "41d18e26-453c-4ecd-b7bd-ff3dc978c275", "allowed": false
```

PR #330, opened 2023-10-03
Enable Cleaner and Pruner cronJobs for cache

fmount:
When `cache` is enabled, we might need to build two additional `cronJobs` to enable both `cache-cleaner` and `cache-pruner`.
This patch extends the ability of the current `cronJob` definition to run different commands and schedules, according to the `CronJobType` passed as argument.
However, the main `glance_controller` is still not able to enable the two new cronJobs because `cache_middleware` is still not supported. That will be part of a different patch but all the `CronJob` related functions and variables are in place at this point.

### DIFF
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -670,7 +670,7 @@ func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glance
 	// NOTE: a CronJob is created at this stage to purge all soft deleted records.
 	// This command should be executed periodically to avoid glance database becomes
 	// bigger by getting filled by soft-deleted records.
-	cronjobDef := glance.CronJob(instance, serviceLabels, serviceAnnotations, "purge")
+	cronjobDef := glance.CronJob(instance, serviceLabels, serviceAnnotations, glance.DBPurge)
 	cronjob := cronjob.NewCronJob(
 		cronjobDef,
 		5*time.Second,
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -670,7 +670,7 @@ func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glance
 	// NOTE: a CronJob is created at this stage to purge all soft deleted records.
 	// This command should be executed periodically to avoid glance database becomes
 	// bigger by getting filled by soft-deleted records.
-	cronjobDef := glance.CronJob(instance, serviceLabels, serviceAnnotations)
+	cronjobDef := glance.CronJob(instance, serviceLabels, serviceAnnotations, "purge")
 	cronjob := cronjob.NewCronJob(
 		cronjobDef,
 		5*time.Second,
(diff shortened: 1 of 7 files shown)


controllers/glance_controller.go
```diff
@@ -670,7 +670,7 @@ func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glance
 	// NOTE: a CronJob is created at this stage to purge all soft deleted records.
 	// This command should be executed periodically to avoid glance database becomes
 	// bigger by getting filled by soft-deleted records.
-	cronjobDef := glance.CronJob(instance, serviceLabels, serviceAnnotations)
+	cronjobDef := glance.CronJob(instance, serviceLabels, serviceAnnotations, glance.DBPurge)
```
fmount, line 673: we still don't support cache (additional work is needed on that front), so we force `glance.DBPurge` here because is the only created cronJob.
Logic to build multiple cronJobs (if cache is enabled) will be a follow up patch.

fmount:
tempest job looks good: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_glance-operator/330/pull-ci-openstack-k8s-operators-glance-operator-main-glance-operator-build-deploy-tempest/1709474557351432192/artifacts/glance-operator-build-deploy-tempest/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-923bc944999f722935616f4e3fa620f5a1b037c8cef824258f2bad11f0a178e4/namespaces/openstack/cronjobs/glance-cron.yaml

pkg/glance/const.go
```diff
@@ -59,12 +62,32 @@ const (
...
 	//DBPurgeDefaultSchedule -
 	DBPurgeDefaultSchedule = "1 0 * * *"
+	//CacheCleanerDefaultSchedule -
+	CacheCleanerDefaultSchedule = "1 0 * * *"
+	//CachePrunerDefaultSchedule -
+	CachePrunerDefaultSchedule = "*/30 * * * *"
 )
 
+// DBPurgeCommandBase -
+var DBPurgeCommandBase = [...]string{"/usr/bin/glance-manage", "--debug", "--config-dir /etc/glance/glance.conf.d", "db purge "}
```
abays, line 83: Is it intentional to have the extra space at the end of `"db purge "`?  I see it was there beforehand in the `cronjob.go` file, but just want to make sure.

fmount: Yeah, it was intentional, because then I can just append the `DBPurgeAge` int:

```
 ( cronJobCommand = append(cronJobCommand, strconv.Itoa(DBPurgeAge))
 ```
 
and not worry about dealing w/ extra spaces when I turn it into a cmd string with:

```
commandString := strings.Join(cronCommand, " ")
```

It basically makes the logic clean enough, and even if a `DBPurgeAge` int is not provided
(resulting in an empty string), the command will work

abays (approved):
/lgtm

PR #331, opened 2023-10-03
Use internal endpoint for keystone

olliewalsh:
Internal service traffic should use internal api endpoints, regardless of whether glance-api is internal or external.

Jira: [OSP-26299](https://issues.redhat.com//browse/OSP-26299)


stuggi (approved):
/lgtm

PR #332, opened 2023-10-04
Move policy testing section in the right section of the project

fmount:
This change moves the policy testing example in the right section of the repo.


abays (approved):
/lgtm

PR #334, opened 2023-10-05
Enable image-cache for GlanceAPI

fmount:
This patch is supposed to handle the `image-cache` feature for `GlanceAPI` and it brings three main changes:

- A new PVC that will be bound to the `image-cache` dir is requested and created, and the `apiVolume` related arrays are updated to have this additional `VolumeMount`

- The `GlanceAPISpec` exposes an `imageCacheSize` parameter, which is passed to the top-level `Glance CR` and if it's greater than zero (or not an empty string in terms of kubernetes validation), it means that a human operator is requesting cache for this component, enabling/triggering the related logic

- The global config for `glanceAPI` has been updated to support all the cache related parameters in the `DEFAULT` section, hence new `templateParameters` have been added as part of the cache activation logic

It's still missing the logic of enabling `cronJobs` from the main controller, but, given the size of the change, that will be part of a follow up patch.
Finally, `envTests` are updated to assert whether a `PVC` exists or not and validate we're not introducing regressions on the basic use case (when `Cache` is not enabled).

### DIFF
--- a/api/v1beta1/glance_types.go
+++ b/api/v1beta1/glance_types.go
@@ -121,6 +121,7 @@ type GlanceSpec struct {
 	Quotas QuotaLimits `json:"quotas,omitempty"`
 
 	// ImageCacheSize, provides the size of the cache that will be reflected in the image_cache_max_size parameter
+	// Local storage request, in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
 	// +kubebuilder:default=""
 	ImageCacheSize string `json:"imageCacheSize"`
 }
--- a/api/v1beta1/glance_types.go
+++ b/api/v1beta1/glance_types.go
@@ -119,6 +119,10 @@ type GlanceSpec struct {
 	// Quotas is defined, per-tenant quotas are enforced according to the
 	// registered keystone limits
 	Quotas QuotaLimits `json:"quotas,omitempty"`
+
+	// ImageCacheSize, provides the size of the cache that will be reflected in the image_cache_max_size parameter
+	// +kubebuilder:default=""
+	ImageCacheSize string `json:"imageCacheSize"`
 }
 
 // PasswordSelector to identify the DB and AdminUser password from the Secret
--- a/config/samples/image-cache/image-cache.yaml
+++ b/config/samples/image-cache/image-cache.yaml
@@ -1,0 +1,55 @@
+# Sample using Ceph as a glance backend
+# Requires a running Ceph cluster and its `/etc/ceph` files in secret `ceph-conf-files`
+# This can be achieved with the `ceph` target of `install_yamls`
+apiVersion: core.openstack.org/v1beta1
+kind: OpenStackControlPlane
+metadata:
+  name: openstack
+spec:
+  glance:
+    template:
+      databaseInstance: openstack
+      serviceUser: glance
+      containerImage: quay.io/podified-antelope-centos9/openstack-glance-api:current-podified
+      customServiceConfig: |
+        [DEFAULT]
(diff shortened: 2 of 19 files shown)


templates/glance/config/00-config.conf
```diff
@@ -76,7 +82,7 @@ config_file = /etc/glance/glance-api-paste.ini
 flavor = keystone
 
 [os_glance_staging_store]
-filesystem_store_datadir = /var/lib/glance/os_glance_staging_store/
+filesystem_store_datadir = /var/lib/glance/image-cache
```
fultonj, line 85: This change will be undone in a future patch set.

controllers/glance_controller.go
```diff
@@ -832,6 +824,45 @@ func (r *GlanceReconciler) ensureRegisteredLimits(
...
+		ctrlResult, err := cachePvc.CreateOrPatch(ctx, h)
+
+		if err != nil {
+			return ctrlResult, err
+		} else if (ctrlResult != ctrl.Result{}) {
+			return ctrlResult, nil
+		}
+	}
+	// End PVC creation/patch
+	return ctrlResult, nil
```
abays, line 863: This could be simplified to reuse the `ctrlResult` and `err` variables from L840, and you could also remove the if-statement and just `return ctrlResult, err` always.

fmount: I agree the if statement is actually redundant as the same block is repeated/evaluated after the execution of this function.
I still need the first block because if the local pvc creation fails, I'd like to return and block the execution, but +1 on removing the second one. And yeah, I can actually reuse the existing variables.
I'll push a quick update for this part.

abays: Yeah, sorry, I didn't meant to imply removing both if-blocks.  It looks good now!

abays (approved):
/lgtm

templates/glance/config/00-config.conf
fmount: Done

api/v1beta1/glance_types.go
```diff
@@ -119,6 +119,10 @@ type GlanceSpec struct {
 	// Quotas is defined, per-tenant quotas are enforced according to the
 	// registered keystone limits
 	Quotas QuotaLimits `json:"quotas,omitempty"`
+
+	// ImageCacheSize, provides the size of the cache that will be reflected in the image_cache_max_size parameter
+	// +kubebuilder:default=""
+	ImageCacheSize string `json:"imageCacheSize"`
```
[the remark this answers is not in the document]
fmount: @abays added a new section under `config/samples/` where `image-cache` is described in terms of what the feature does and I also added an example that can be applied using `kustomize`

abays (approved):
/lgtm

config/samples/image-cache/image-cache.yaml
```diff
@@ -0,0 +1,55 @@
...
+  name: openstack
+spec:
+  glance:
+    template:
+      databaseInstance: openstack
+      serviceUser: glance
+      containerImage: quay.io/podified-antelope-centos9/openstack-glance-api:current-podified
+      customServiceConfig: |
+        [DEFAULT]
+        enabled_backends = default_backend:rbd
```
konan-abhi, line 16: Since glance uses caching as an middleware, In addition we also need to set below parameter in glance-api.conf to enable the cache ,

[paste_deploy]
flavor = keystone+cachemanagement

If this is not set then I am doubtful that image caching will work.

fmount: ack and thanks for the feedback, I'm going to update the sample file in a separate patch.

PR #335, opened 2023-10-06
Update PVC samples to be RFC 1123 compliant

fultonj:
The image conversion and decompression PVC samples cannot be created without this patch since they
will fail with:

The PersistentVolumeClaim "image_import_staging_workspace" is invalid: metadata.name: Invalid value:
"image_import_staging_workspace": a lowercase RFC 1123

Change Glance CR extraMounts to use the name in the PVC so that both sample files are consistent.

Also, add `enabled_import_methods` with `glance-direct` to `customServiceConfig` sample because a popular way to test these settings are to use `openstack image create --import` to force the use of the import plugin instead of direct upload.

Jira: [OSP-29527](https://issues.redhat.com//browse/OSP-29527)

### DIFF
--- a/config/samples/import_plugins/image_conversion/image_conversion.yaml
+++ b/config/samples/import_plugins/image_conversion/image_conversion.yaml
@@ -11,6 +11,7 @@ spec:
   customServiceConfig: |
     [DEFAULT]
     enabled_backends = default_backend:rbd
+    enabled_import_methods=[web-download,glance-direct]
     [glance_store]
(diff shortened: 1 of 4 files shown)


fultonj:
holding since I also need to update the `claimName`

https://github.com/openstack-k8s-operators/glance-operator/blob/main/config/samples/import_plugins/image_conversion/image_conversion.yaml#L55

fmount (approved):

PR #338, opened 2023-10-11
Enable pruner and cleaner CronJobs

fmount:
This patch enables both `pruner` and `cleaner` `cronJobs` that are supposed to manage the `/var/lib/glance/image-cache` directory to stay under the upper limit defined by `ImageCacheSize` (that triggers a `PVC` request of that size), as well as clean the incomplete/stalled images from the related folders.
In order to make the above happen, a few things are fixed by this patch:

1. `00-config.conf` (the global one) is updated to include the image-cache related parameters (if cache is enabled);

2. the `CronJobs` creation logic has been moved to a dedicated `ensureCronJobs` function (it's the same approach we adopted for `PVC` creation) and it results cleaner from a `glance_controller` point of view;

3. The `CronJob` function itself mounts an additional `Volume` in case of cleaner/pruner: the `glance-cache` PVC is mounted in `/var/lib/glance/image-cache` because it represents the target directory managed by the `cronJob` command; in case of `DBPurge` this additional volume is not provided as it's not required in that case;

4. The `image_cache_max_size` parameter was broken and filled with a wrong value, hence with this patch we normalize the `ImageCacheSize` parameter (via `resource.MustParse()`) and pass the resulting `Value()` to the `templateParameters` that are supposed to fill the right section of the glance config file [1][2]

[1] https://github.com/openstack/glance/blob/master/glance/image_cache/__init__.py#L68
[2] https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity_example_test.go


fmount:
When cache is enabled we can see cronJobs:

```
[stack@osp-storage-04 gop]$ oc get cronjob
NAME                  SCHEDULE       SUSPEND   ACTIVE   LAST SCHEDULE   AGE
glance-cleaner-cron   1 0 * * *      False     0        <none>          5h50m
glance-pruner-cron    */30 * * * *   False     0        27m             5h50m
glance-purge-cron     1 0 * * *      False     0        <none>          5h50m
```

and if we `oc describe` one of them (cleaner):

```
[stack@osp-storage-04 gop]$ oc describe cronjob glance-cleaner-cron
Name:                          glance-cleaner-cron
Namespace:                     openstack
..
Schedule:                      1 0 * * *
...
Pod Template:
  Labels:           <none>
  Service Account:  glance-glance
  Containers:
   glance-cron:
    Image:      quay.io/podified-antelope-centos9/openstack-glance-api:current-podified
    Port:       <none>
    Host Port:  <none>
    Command:
      /bin/bash
    Args:
      -c
      /usr/bin/glance-cache-cleaner --config-dir /etc/glance/glance.conf.d
    Environment:  <none>
    Mounts:
      /etc/glance/glance.conf.d from db-purge-config-data (ro)
      /var/lib/glance/image-cache from image-cache (rw)
  Volumes:
   db-purge-config-data:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  glance-config-data
    Optional:    false
   image-cache:
    Type:            PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:       glance-cache
    ReadOnly:        false
```

we can see the relevant mounts are present and the referenced `pvc` is the `glance-cache` one, created for this purpose.

controllers/glance_controller.go
```diff
@@ -776,6 +771,15 @@ func (r *GlanceReconciler) generateServiceConfig(
 			glance.DatabaseName,
 		),
 	}
+	// We set in the main 00-config-default.conf the image-cache bits that will
+	// be used by CronJobs cleaner and pruner
+	if len(instance.Spec.ImageCacheSize) > 0 {
+		cacheSize := resource.MustParse(instance.Spec.ImageCacheSize)
```
abays, line 777: It looks like `resource.MustParse()` panics when something goes wrong [1].  Maybe we should catch the panic and report an error instead of allowing this to potentially crash the controller-manager pod here?

[1] https://github.com/kubernetes/apimachinery/blob/1ca6c696c8e470d0745788dad2f69de813b2237d/pkg/api/resource/quantity.go#L136-L142

fmount: Makes sense and it doesn't hurt catching an issue here. Thanks for the hint. 
In theory if we reach this point it means that we already failed during the `pvc` creation, but I agree we should have full control of this block instead of relying in the general flow that could change over the time.
I'm going to update this part in my next iteration.

fmount: ok I think we should survive to the `MustParse` panic there and produce proper log of the issue w/o crushing the Pod [1]
[1] https://go.dev/play/p/HKcBzeRsRXA

abays (approved):
/lgtm

PR #343, opened 2023-10-13
Replace DBPurge Debug boolean with CronJob

fmount:
We previously introduced a `DBPurge` boolean under the `GlanceDebug` struct to make sure we were able to trigger a `cronJob` in `debug` mode. Since the first patch multiple cronjobs have been added (`pruner` and `cleaner`), hence it makes sense to change the boolean variable name to something that is valid and meaningful for a generic `cronjob` object.
This patch concludes the `image-cache` series support in the glance-operator.


abays (approved):
/lgtm

PR #350, opened 2023-10-26
Set PVC access mode ReadWriteOnce

fmount:
Experimenting w/ RWO vs RWX w/o moving (yet) to statefulset to temporary fix local-storage

### DIFF
--- a/config/samples/import_plugins/image_conversion/image_conversion_pvc.yaml
+++ b/config/samples/import_plugins/image_conversion/image_conversion_pvc.yaml
@@ -4,7 +4,7 @@ metadata:
   name: image-import-staging-workspace
 spec:
   accessModes:
-  - ReadWriteMany
+  - ReadWriteOnce
   resources:
     requests:
       storage: 5Gi
--- a/config/samples/import_plugins/image_decompression/image_decompression_pvc.yaml
+++ b/config/samples/import_plugins/image_decompression/image_decompression_pvc.yaml
@@ -4,7 +4,7 @@ metadata:
   name: image-import-staging-workspace
 spec:
   accessModes:
-  - ReadWriteMany
+  - ReadWriteOnce
   resources:
     requests:
       storage: 5Gi
--- a/pkg/glance/pvc.go
+++ b/pkg/glance/pvc.go
@@ -28,7 +28,7 @@ func Pvc(api *glancev1.Glance, labels map[string]string, pvcType PvcType) *corev
 		},
 		Spec: corev1.PersistentVolumeClaimSpec{
 			AccessModes: []corev1.PersistentVolumeAccessMode{
-				corev1.ReadWriteMany,
+				corev1.ReadWriteOnce,
 			},
 			Resources: corev1.ResourceRequirements{
 				Requests: corev1.ResourceList{


fultonj (commented):
Yes, let's see if we can use RWO and enable worker_self_reference_url 
 https://opendev.org/openstack/glance/commit/41e1cecbe63c778ce8e92519993c61588ea1f0cb

fultonj:
@fmount do you want to change this from a DNM to a patch to fully switch to `ReadWriteOnce` in the examples that use it [1]?

Maybe update `config/samples/import_plugins/image_conversion/image_conversion.yaml` with a commented example of `worker_self_reference_url`? 

Or maybe there's a way to get the output of `oc get secret glance-config-data -o json | jq -r '.data."01-config.conf"' | base64 -d` to contain the GlanceInternal endpoint the way TripleO does [2]?

I made a quick update in the architecture example [3] but maybe we can get the operator to configure that setting it if there's >1 replica? 

[1] 
```
[fultonj@runcible glance-operator{pvc_rwx_rwo}]$ grep -R ReadWriteMany 
config/samples/import_plugins/image_conversion/image_conversion_pvc.yaml:  - ReadWriteMany
config/samples/import_plugins/image_decompression/image_decompression_pvc.yaml:  - ReadWriteMany
pkg/glance/pvc.go:				corev1.ReadWriteMany,
[fultonj@runcible glance-operator{pvc_rwx_rwo}]$ 
```

[2] https://github.com/openstack/tripleo-heat-templates/blob/master/deployment/glance/glance-api-container-puppet.yaml#L646C28-L646C53

[3] https://github.com/openstack-k8s-operators/architecture/pull/25

fmount:
> @fmount do you want to change this from a DNM to a patch to fully switch to `ReadWriteOnce` in the examples that use it [1]?
> 
> Maybe update `config/samples/import_plugins/image_conversion/image_conversion.yaml` with a commented example of `worker_self_reference_url`?

+1 on having a dedicated example, but does this parameter applies to `image_conversion` ? If so, we should add `glance-direct` in the same example, @konan-abhi wdyt?
> 
> Or maybe there's a way to get the output of `oc get secret glance-config-data -o json | jq -r '.data."01-config.conf"' | base64 -d` to contain the GlanceInternal endpoint the way TripleO does [2]?

not from the `glance-config-data`, but we can simply query the `glance-public` svc and look for the `Domain` from the `ingresses.config`: I'll try something in my next iteration.

> 
> I made a quick update in the architecture example [3] but maybe we can get the operator to configure that setting it if there's >1 replica?
> 
> [1]
> 
> ```
> [fultonj@runcible glance-operator{pvc_rwx_rwo}]$ grep -R ReadWriteMany 
> config/samples/import_plugins/image_conversion/image_conversion_pvc.yaml:  - ReadWriteMany
> config/samples/import_plugins/image_decompression/image_decompression_pvc.yaml:  - ReadWriteMany
> pkg/glance/pvc.go:				corev1.ReadWriteMany,
> [fultonj@runcible glance-operator{pvc_rwx_rwo}]$ 
> ```
> 
> [2] https://github.com/openstack/tripleo-heat-templates/blob/master/deployment/glance/glance-api-container-puppet.yaml#L646C28-L646C53
> 
> [3] [openstack-k8s-operators/architecture#25](https://github.com/openstack-k8s-operators/architecture/pull/25)

Sure, we can do that, however I have two points:
1.  `worker_self_reference_url` can be set only if you have `glance-direct` import method enabled in your deployment, hence if `glance-direct` is added then we set `worker_self_reference_url` under the `[DEFAULT]` section (which should point to the `glance-public svc`); /cc @konan-abhi to confirm
2. if we set `RWO` with `local-storage` (which is totally fine with me), we're going to hit issues during the scale out of the glance service in multinode environments when the `Pod` is scheduled on a different node, and this issue will be addressed by openstack-k8s-operators/glance-operator/pull/352 in a follow up

konan-abhi:
> > @fmount do you want to change this from a DNM to a patch to fully switch to `ReadWriteOnce` in the examples that use it [1]?
> > Maybe update `config/samples/import_plugins/image_conversion/image_conversion.yaml` with a commented example of `worker_self_reference_url`?
> 
> +1 on having a dedicated example, but does this parameter applies to `image_conversion` ? If so, we should add `glance-direct` in the same example, @konan-abhi wdyt?

Yes, worker_self_reference_url needs to be set only if glance-direct import method is enabled.
> 
> > Or maybe there's a way to get the output of `oc get secret glance-config-data -o json | jq -r '.data."01-config.conf"' | base64 -d` to contain the GlanceInternal endpoint the way TripleO does [2]?
> 
> not from the `glance-config-data`, but we can simply query the `glance-public` svc and look for the `Domain` from the `ingresses.config`: I'll try something in my next iteration.
> 
> > I made a quick update in the architecture example [3] but maybe we can get the operator to configure that setting it if there's >1 replica?
> > [1]
> > ```
> > [fultonj@runcible glance-operator{pvc_rwx_rwo}]$ grep -R ReadWriteMany 
> > config/samples/import_plugins/image_conversion/image_conversion_pvc.yaml:  - ReadWriteMany
> > config/samples/import_plugins/image_decompression/image_decompression_pvc.yaml:  - ReadWriteMany
> > pkg/glance/pvc.go:				corev1.ReadWriteMany,
> > [fultonj@runcible glance-operator{pvc_rwx_rwo}]$ 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > [2] https://github.com/openstack/tripleo-heat-templates/blob/master/deployment/glance/glance-api-container-puppet.yaml#L646C28-L646C53
> > [3] [openstack-k8s-operators/architecture#25](https://github.com/openstack-k8s-operators/architecture/pull/25)
> 
> Sure, we can do that, however I have two points:
> 
>     1. `worker_self_reference_url` can be set only if you have `glance-direct` import method enabled in your deployment, hence if `glance-direct` is added then we set `worker_self_reference_url` under the `[DEFAULT]` section (which should point to the `glance-public svc`); /cc @konan-abhi to confirm

+1, correct
> 
>     2. if we set `RWO` with `local-storage` (which is totally fine with me), we're going to hit issues during the scale out of the glance service in multinode environments when the `Pod` is scheduled on a different node, and this issue will be addressed by [WIP - DNM - Statefulset #352](https://github.com/openstack-k8s-operators/glance-operator/pull/352) in a follow up

fmount:
Closing this patch in favor of https://github.com/openstack-k8s-operators/glance-operator/pull/352 where the problem of getting a `stable pvc` access for any `glanceAPI` instance is solved moving to `StatefulSet`.

PR #351, opened 2023-10-26
Do not require DbSync to mount all the Glance volumes

fmount:
`DBSync` only requires a few files to run. This patch simplifies it only mounting the required files.
For instance now the `PVC` associated with `Glance` is not required to let `dbsync` run and it's not
mounted anymore.


konan-abhi (approved):
Looks good to me!

abays (approved):
/lgtm

PR #358, opened 2023-10-31
Drop container images from config/samples

dprince:
Jira: [OSP-30111](https://issues.redhat.com//browse/OSP-30111)

### DIFF
--- a/config/samples/glance_v1beta1_glance.yaml
+++ b/config/samples/glance_v1beta1_glance.yaml
@@ -4,20 +4,17 @@ metadata:
   name: glance
 spec:
   serviceUser: glance
-  containerImage: quay.io/podified-antelope-centos9/openstack-glance-api:current-podified
   customServiceConfig: |
     [DEFAULT]
     debug = true
   databaseInstance: openstack
   databaseUser: glance
   glanceAPIInternal:
-    containerImage: quay.io/podified-antelope-centos9/openstack-glance-api:current-podified
     debug:
       service: false
     preserveJobs: false
     replicas: 1
   glanceAPIExternal:
-    containerImage: quay.io/podified-antelope-centos9/openstack-glance-api:current-podified
     debug:
       service: false
     preserveJobs: false
--- a/test/functional/sample_test.go
+++ b/test/functional/sample_test.go
@@ -73,7 +73,7 @@ var _ = Describe("Samples", func() {
 			GetGlance(name)
 		})
 	})
-	When("glance_v1beta1_glanceapi.yaml sample is applied", func() {
+    When("glance_v1beta1_glanceapi.yaml sample is applied", func() {
 		It("GlanceAPI is created", func() {
 			name := CreateGlanceAPIFromSample("glance_v1beta1_glanceapi.yaml", glanceTest.Instance)
 			GetGlanceAPI(name)
--- a/test/kuttl/tests/glance_scale/01-assert.yaml
+++ b/test/kuttl/tests/glance_scale/01-assert.yaml
@@ -17,19 +17,16 @@ metadata:
   name: glance
 spec:
   serviceUser: glance
-  containerImage: quay.io/podified-antelope-centos9/openstack-glance-api:current-podified
   customServiceConfig: |
     [DEFAULT]
     debug = true
   databaseInstance: openstack
   databaseUser: glance
   glanceAPIInternal:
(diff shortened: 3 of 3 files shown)


config/samples/glance_v1beta1_glance.yaml
```diff
@@ -4,20 +4,17 @@ metadata:
...
   databaseInstance: openstack
   databaseUser: glance
   glanceAPIInternal:
-    containerImage: quay.io/podified-antelope-centos9/openstack-glance-api:current-podified
     debug:
       service: false
     preserveJobs: false
     replicas: 1
   glanceAPIExternal:
-    containerImage: quay.io/podified-antelope-centos9/openstack-glance-api:current-podified
```
fmount, line 20: @dprince I suspect also kuttl needs to be aligned with this change (as images are hardcoded)

config/samples/glance_v1beta1_glanceapi.yaml
```diff
@@ -5,7 +5,6 @@ metadata:
 spec:
   serviceUser: glance
   serviceAccount: glance
-  containerImage: quay.io/podified-antelope-centos9/openstack-glance-api:current-podified
```
fmount, line 8: Not sure functional tests are going to work with this [1] because the `webhook` is triggered by the `top-level` `CR`, which is not happening with this (/cc @abays to ack/nack what I'm saying)

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_glance-operator/358/pull-ci-openstack-k8s-operators-glance-operator-main-functional/1719366033648652288/artifacts/test/build-log.txt

config/samples/glance_v1beta1_glance.yaml
dprince: @fmount ack. Good call and thanks for pointing it out

config/samples/glance_v1beta1_glanceapi.yaml
dprince: There is actually no webhook for GlanceAPI, only top level Glance. If we want to continue running tests on that sample we'd have to implement a webhook for it (which I think is beyond the scope of this patch). While some operators do implement webhooks on sub-resources it isn't strictly required I think as those resources should not be used externally anyway.

config/samples/glance_v1beta1_glance.yaml
dprince: While kuttl is hard coded though, it should still align with the hard coded defaults in the Go library coded within this operator. So I think it should be okay right?

config/samples/glance_v1beta1_glanceapi.yaml
dprince: I actually disabled the sample test for now. (if that works). If not, the approach should be to first do another PR to create webhooks on the GlanceAPI resource. Then once that is completed follow up with this PR again.

config/samples/glance_v1beta1_glance.yaml
fmount: Yeah, it makes sense: when we bump the `glanceAPI` image in the operator, we can update kuttl tests as well.
We should be good then from that point of view I guess.
Thanks for checking

config/samples/glance_v1beta1_glanceapi.yaml
fmount: I think it's ok to merge this patch as it is, and I can follow up by creating the missing webhook later. I see you already added a note for that, so I'm fine merging this as long as the tests are green. 
If I manage to push a webhooks related patch early next week, we can rebase this one on top of that.

fmount: @dprince I think that in order to unblock this patch (in a more clean way) is to merge PR #363 first and rebase this one that will result in just removing the remaining lines.
Let me know what you think and eventually review #363.

dprince: @fmount thanks! I approved the PR. Once it lands I can rebase. Also, I added a custom kuttl script at the end to verify the containerImage. It works now, but only verifies the top level "Glance" CR. It could be expanded to verify the sub-resources for GlanceAPI and also the Deployments that it creates.

fmount (approved):
/lgtm

PR #363, opened 2023-11-04
Add missing GlanceAPI webhook

fmount:
In general it is not allowed to create a `glanceAPI` instance without defining the top-level `Glance` `CR`.
However, if for any, unknown reason a `glanceAPI` `CR` is created, it will always fail because of the required `.spec.ContainerImage` field that appears to be empty or simply not defined.
This patch introduces the `glanceAPI` missing `webhook` that covers at the `glanceAPI` level the `ContainerImage` propagation/definition.


config/samples/glance_v1beta1_glanceapi.yaml
```diff
@@ -5,7 +5,6 @@ metadata:
 spec:
   serviceUser: glance
   serviceAccount: glance
-  containerImage: quay.io/podified-antelope-centos9/openstack-glance-api:current-podified
```
fmount, line 8: @dprince we can use this patch as a follow up of PR #358 and remove this line here using and not fail because of the `glanceAPI` webhook.

dprince (approved):

api/v1beta1/glanceapi_webhook.go
```diff
@@ -0,0 +1,93 @@
...
+type GlanceAPIDefaults struct {
+	ContainerImageURL string
+}
+
+var glanceAPIDefaults GlanceAPIDefaults
+
+// log is for logging in this package.
+var glanceapilog = logf.Log.WithName("glanceapi-resource")
+
+// SetupGlanceAPIDefaults - initialize Glance spec defaults for use with either internal or external webhooks
```
abays, line 36: nit: `s/initialize Glance/initialize GlanceAPI/`

fmount: Good catch, updating it

api/v1beta1/glanceapi_webhook.go
```diff
@@ -0,0 +1,93 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+// GlanceAPIDefaults -
+type GlanceAPIDefaults struct {
+	ContainerImageURL string
+}
+
+var glanceAPIDefaults GlanceAPIDefaults
```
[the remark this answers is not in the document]
fmount: Right, in the `Default()` function I should have:
```golang
if spec.GlanceAPITemplate.ContainerImage == "" {
    spec.GlanceAPITemplate.ContainerImage = glanceAPIDefaults.ContainerImageURL
}
```

instead of `glanceDefaults.ContainerImageURL`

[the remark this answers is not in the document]
fmount: Done

api/v1beta1/glanceapi_webhook.go
fmount: Done

api/v1beta1/glanceapi_webhook.go
```diff
@@ -0,0 +1,93 @@
...
+
+// log is for logging in this package.
+var glanceapilog = logf.Log.WithName("glanceapi-resource")
+
+// SetupGlanceAPIDefaults - initialize GlanceAPI spec defaults for use with either internal or external webhooks
+func SetupGlanceAPIDefaults(defaults GlanceAPIDefaults) {
+	glanceAPIDefaults = defaults
+	glancelog.Info("Glance defaults initialized", "defaults", defaults)
+}
```
abays, line 40: I think we also need something that pulls-in the ENV vars and sets them in the `GlanceAPIDefaults`, like we do here [1].  Maybe I just am missing where you're already doing this, however. 

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/66fbd23a85871d6c734bd49d0af628e4b50ec23b/api/v1beta1/common.go#L114-L121

fmount: You're right, I need that function too in `common.go`. Updating the patch in a moment.

fmount: Done

dprince (approved):

PR #372, opened 2023-11-16
Do not append "single" to the single glanceAPI

fmount:
This is a cosmetic change that improves the user experience when the glance Pod is deployed.
It's good to have the distinction between internal and external in case we split the API, but if we have a single instance it doesn't make sense to call the pod as "glance-single-api". This patch just improves the naming when a single glanceAPI instance is deployed, resulting simply in "glance-api-0" (it will increase with the replica number).


abays (approved):
/lgtm

PR #376, opened 2023-11-29
Move -log sidecar container to dumb-init

fmount:
As noticed in other operators, the -log sidecar ignores SIGTERM on delete. This patch moves the glance-operator to the dumb-init usage to make sure the SIGTERM is handled properly by the sidecar.
Nova [1] moved to same approach recently and we can see the Pod is immediately terminated when receives the SIGTERM signal.

[1] https://github.com/openstack-k8s-operators/nova-operator/pull/606


fultonj (approved):
/lgtm

PR #384, opened 2023-12-05
Deploy an arbitrary number of glanceAPI

fmount:
This patch introduces a new API interface that allows to specify an arbitrary number of `glanceAPI`. By giving this ability from an interface point of view, the `glance-operator` is now responsible to reconcile the deployed instances in terms of lifecycle, as well as making sure we always have a selected API registered to the keystone `image` service.
As part of this change we are now able to:
- grow or shrink the API deployment 
- update the `keystoneEndpoint` to select which instance should be registered in the `keystone` catalog
- propagate `extraVolumes` to the single instance as done for both `Cinder` and `Manila`

Demo: https://asciinema.org/a/625117


api/v1beta1/glance_webhook.go
```diff
@@ -76,8 +110,16 @@ var _ webhook.Validator = &Glance{}
...
 	glancelog.Info("validate create", "name", r.Name)
+	// At creation time, if the CR has an invalid keystoneEndpoint value (that
+	// doesn't match with any defined backend), return an error.
+	if !r.Spec.isValidKeystoneEP() {
+		return errors.New("KeystoneEndpoint is assigned to an invalid glanceAPI instance")
+	}
 
-	// TODO(user): fill in your validation logic upon object creation.
+	//TODO:
+	// - Two glanceAPI with the same name can't be deployed
```
abays, line 120: Is it even possible to deploy two `GlanceAPI`s with the same name, given that they are specified in a map and the keys are the names?

fmount: No, it's actually not possible, and the CR is not updated, however, I'd like to explicitly deny specific actions like this one, and return an error using `webhooks`, which is something we do for both the `keystoneEndpoint` and  `layout` changes

api/v1beta1/glance_types.go
```diff
@@ -101,8 +101,7 @@ type GlanceSpec struct {
 
 	// +kubebuilder:validation:Required
 	// GlanceAPI - Spec definition for the API service of this Glance deployment
```
abays, line 103: `s/GlanceAPI/GlanceAPIs`

api/v1beta1/glance_webhook.go
```diff
@@ -88,9 +130,19 @@ func (r *Glance) ValidateUpdate(old runtime.Object) error {
...
-	o := old.(*Glance);
-	if (r.Spec.GlanceAPI.Type != o.Spec.GlanceAPI.Type) {
-		return errors.New("GlanceAPI deployment layout can't be updated")
+	o := old.(*Glance)
+	for key, glanceAPI := range r.Spec.GlanceAPIs {
+		// When a new entry (new glanceAPI instance) is added in the main CR, it's
+		// possible that the old CR used to compare the new map had no entry with
+		// the same name. This represent a valid use case and we shouldn't prevent
+		// to grow the deployment
+		if o.Spec.GlanceAPIs[key].Type == "" {
```
abays, line 139: When I first read this line I was expecting the comment above it to mention something about `Type`, but I think the comment makes it clear that we care whether the `GlanceAPI` existed at all in the old `Glance`.  But maybe it would be better to use a check like `if _, ok := o.Spec.GlanceAPIs[key]; ok {` to more clearly-align with what the comment is indicating?  I'm truly just asking, not suggesting that it be one way or the other.

fmount: Yes, I think it makes sense: instead of checking a potential `empty` parameter in an `entry` that might have not been there in the `old` object provided by the runtime, we can just use the way you proposed (thanks, it's also more elegant in terms of style), except for the fact that should be in the form `if _, found := o.Spec.GlanceAPIs[key]; !found {`, (with the `not` to indicate that it's not found) so we can move forward with the iteration.
In my local test I get:

```
[stack@osp-storage-04 layout]$ oc edit glance glance
Edit cancelled, no changes made.
``` 
when I try to add a new instance, so I think it worked as expected. 
Thanks for this suggestion!

api/v1beta1/glance_webhook.go
fmount: This seems to be covered at creation time:

Example:

```
kind: Glance
...
spec:
  serviceUser: glance
  databaseInstance: openstack
  databaseUser: glance
  glanceAPIs:
    api0:
      type: "single"
      preserveJobs: false
      replicas: 1
    api0:
      type: "single"
      preserveJobs: false
      replicas: 1
    api1:
      type: "single"
      preserveJobs: false
      replicas: 1

```

results in:

```
[stack@osp-storage-04 layout]$ oc kustomize single/ | oc create -f -
error: map[string]interface {}(nil): yaml: unmarshal errors:
  line 23: mapping key "api0" already defined at line 17
error: no objects passed to create
```

If I simply `oc create glance-cr.yaml` the deploy starts and it results in the last definition taking place, because it's a `map[string]GlanceAPITemplate`, and it's like you override the value for an existing key.

```
glance-api0-single-0        3/3     Running     0          105s
glance-api1-single-0        3/3     Running     0          105s
```

Sorry for not catching this in my previous comment, but basically by definition `maps` don't allow duplicated keys, even if they're strings, which means that if you define the same API multiple times, it's just an override of the previous definitions.

fmount:
I think there's a valid issue in tempest and it might be related to the endpoint creation:

```
2023-12-05 23:01:10.201 130 WARNING urllib3.connectionpool [-] Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f5c2f9fa160>: Failed to establish a new connection: [Errno -2] Name or service not known')': /[00m
2023-12-05 23:01:10.205 130 WARNING urllib3.connectionpool [-] Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f5c2f9fa5e0>: Failed to establish a new connection: [Errno -2] Name or service not known')': /[00m
2023-12-05 23:01:10.209 130 WARNING urllib3.connectionpool [-] Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f5c2f9fa700>: Failed to establish a new connection: [Errno -2] Name or service not known')': /[00m
2023-12-05 23:01:10.217 130 ERROR config_tempest.constants [-] Request on service 'image' with url 'http://glance-default-external-public.openstack.svc:9292' failed: urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='glance-default-external-public.openstack.svc', port=9292): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f5c2f9fa880>: Failed to establish a new connection: [Errno -2] Name or service not known'))[00m
2023-12-05 23:01:10.220 130 CRITICAL tempest [-] Unhandled error: urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='glance-default-external-public.openstack.svc', port=9292): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f5c2f9fa880>: Failed to establish a new connection: [Errno -2] Name or service not known'))
2023-12-05 23:01:10.220 130 ERROR tempest Traceback (most recent call last):
2023-12-05 23:01:10.220 130 ERROR tempest   File "/usr/lib/python3.9/site-packages/urllib3/connection.py", line 169, in _new_conn
2023-12-05 23:01:10.220 130 ERROR tempest     conn = connection.create_connection(
2023-12-05 23:01:10.220 130 ERROR tempest   File "/usr/lib/python3.9/site-packages/urllib3/util/connection.py", line 73, in create_connection
2023-12-05 23:01:10.220 130 ERROR tempest     for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
2023-12-05 23:01:10.220 130 ERROR tempest   File "/usr/lib64/python3.9/socket.py", line 954, in getaddrinfo
2023-12-05 23:01:10.220 130 ERROR tempest     for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
2023-12-05 23:01:10.220 130 ERROR tempest socket.gaierror: [Errno -2] Name or service not known
2023-12-05 23:01:10.220 130 ERROR tempest 
2023-12-05 23:01:10.220 130 ERROR tempest During handling of the above exception, another exception occurred:
```

controllers/glance_controller.go
```diff
@@ -166,7 +165,7 @@ func (r *GlanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 		instance.Status.Conditions.Init(&cl)
 
 		// Register overall status immediately to have an early feedback e.g. in the cli
-		return ctrl.Result{}, nil
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
fmount: I suspect this is a leftover of my troubleshooting during the development. I'm going to restore this because I'm not going to introduce any change in the patter used across the operator. Thanks, this is a good catch and let me know if you find any other "violation" of common patters that I might avoid in this context. Restoring it in my next PS.

controllers/glance_controller.go
```diff
@@ -930,3 +970,47 @@ func (r *GlanceReconciler) registeredLimitsDelete(
...
+func (r *GlanceReconciler) glanceAPICleanup(ctx context.Context, instance *glancev1.Glance) error {
+	// Generate a list of GlanceAPI CRs
+	apis := &glancev1.GlanceAPIList{}
+	listOpts := []client.ListOption{
+		client.InNamespace(instance.Namespace),
+	}
+	if err := r.Client.List(ctx, apis, listOpts...); err != nil {
+		r.Log.Error(err, "Unable to retrieve GlanceAPI CRs %v")
+		return nil
+	}
```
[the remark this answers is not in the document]
fmount: That makes sense, but In order to avoid issues to the `glanceAPICleanup` function, another way of doing this is to move it outside of the loop that becomes as follow:

```golang
	for name, glanceAPI := range instance.Spec.GlanceAPIs {
		err = r.apiDeployment(ctx, instance, name, glanceAPI, helper, serviceLabels)
		if err != nil {
			return ctrl.Result{}, err
		}
	}
	err = r.glanceAPICleanup(ctx, instance)
	if err != nil {
		return ctrl.Result{}, err
	}
```
by doing this, every time a Reconcile is triggered (even when I `oc patch glance glance --type=json -p="[{'op': 'remove', 'path': '/spec/glanceAPIs/defaultX'}]"` I can make sure the resources are cleaned up.
How does this solution sound?

controllers/glance_controller.go
```diff
@@ -608,103 +618,125 @@ func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glance
...
+		glanceAPI, err = r.ensureAPIDeployment(ctx, instance, glancev1.APIInternal, instanceName, helper)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Get external GlanceAPI's condition status and compare it against priority of internal GlanceAPI's condition
+	apiCondition := glanceAPI.Status.Conditions.Mirror(glancev1.GlanceAPIReadyCondition)
+	// Get internal GlanceAPI's condition status for comparison with external below
+	internalAPICondition := glanceAPI.Status.Conditions.Mirror(glancev1.GlanceAPIReadyCondition)
```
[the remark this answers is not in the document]
fmount: Yeah, I'm not really doing a real distinction in the `GlanceAPIReadyCondition` for two reasons:
1. it applies well to `single` instance deployment;
2. both internal and external logically represent an API that serves a given workload

Do you think I should introduce a:

1. `GlanceAPIInternalReadyCondition`
2. `GlanceAPIExternalReadyCondition`
 
and see if it makes sense growing this more for `edge` and `single` deployments?
This might be content for a follow up patch, @abays wdyt?

api/v1beta1/glance_types.go
fmount: Done

controllers/glance_controller.go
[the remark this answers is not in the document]
abays: That looks good to me!

controllers/glanceapi_controller.go
```diff
@@ -799,3 +742,93 @@ func (r *GlanceAPIReconciler) createHashOfInputHashes(
...
+
+	// If the parent controller didn't set the annotation, the current glanceAPIs
+	// shouldn't register the endpoints in keystone
+	if len(instance.ObjectMeta.Annotations) == 0 ||
+		instance.ObjectMeta.Annotations[glance.KeystoneEndpoint] != "true" {
+		// Mark the KeystoneEndpointReadyCondition as True because there's nothing
+		// to do here
+		instance.Status.Conditions.Set(instance.Status.Conditions.Mirror(condition.KeystoneEndpointReadyCondition))
+		instance.Status.Conditions.MarkTrue(
+			condition.KeystoneEndpointReadyCondition, condition.ReadyMessage)
```
abays, line 765: I'm confused here.  Why do we `Mirror()` if we're just overriding it with with `MarkTrue()` on the next line?

controllers/glanceapi_controller.go
```diff
@@ -799,3 +742,93 @@ func (r *GlanceAPIReconciler) createHashOfInputHashes(
...
+	// mirror the Status, Reason, Severity and Message of the latest keystoneendpoint condition
+	// into a local condition with the type condition.KeystoneEndpointReadyCondition
+	c := ksSvc.GetConditions().Mirror(condition.KeystoneEndpointReadyCondition)
+	if c != nil {
+		instance.Status.Conditions.Set(c)
+	}
+
+	if (ctrlResult != ctrl.Result{}) {
+		return ctrlResult, nil
+	}
```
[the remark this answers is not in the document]
fmount: Makes sense, thanks, removing in my next iteration.

controllers/glanceapi_controller.go
fmount: Right, good catch, thanks. We don't need to `Mirror` the the condition as we're explicitly setting it to `True` (as per the comment in L762). I'm going to remove the `Mirror()` call because we don't need it here.

fmount:
> @fmount +1 from me. I can approve if you're fine with just my review.

I'm fine, thank you @abays. Given I'm mostly afk for now I'd like to trigger a few more tests and coordinate with you both on this one and the `openstack-operator` related patch next week

fmount:
Despite I'm still trying to get the CI green on PR#587 and predict any potential failure
with patch [2], I think this change it's ready to go.
I tested with and without network isolation, with multiple instances deployed through
the `openstack-operator` (see [1]). As an additional test, I deployed `multiple ceph`
clusters (in the `Pod` form [4]) and use `extraVolumes` to attach each of them to
a different `glanceAPI`.

As the demo [1] shows:

```
- api0 => upload an image => ceph0
- patch the `openstackcontrolplane` to switch endpoint to api1
- api1 => upload an image => ceph1
```

The three steps above are basically the common part of the tests I made. I'm going
to create a doc patch as a follow up to better describe the design choices and how
the whole flow works, for both the `glanceAPI` lifecycle and the endpoints management.
As a follow up, we might add more conditions for the `edge` type API, so that we
can skip any public endpoint creation (even if not exposed through a `Route`).
By doing this we can prevent them to have the ability to be registered in `keystone`.

@abays if you are ok with this patch, once is approved, we should rebase/merge [2]
to avoid issues and inconsistencies with the API.
Both `kuttl` and `tempest` test prove that the covered use cases work as expected,
(see also the Prow job logs in [2]).
I'll keep the `do-not-merge/hold` label to block merging it until you've checked [2].
(that of course should be `rebased` on this one, while right now `kuttl` uses my custom
bundle [4] based on this commit).

[1] https://asciinema.org/a/626515
[2] https://github.com/openstack-k8s-operators/openstack-operator/pull/587
[3] https://github.com/openstack-k8s-operators/install_yamls/pull/654
[4] https://quay.io/repository/fpantano/glance-operator-bundle:3598c3d1fad3

abays (approved):
/lgtm

fmount:
Given that this change is ready and the openstack-operator patch is green [1], I'm removing the `hold` label and merge this change.

[1] https://review.rdoproject.org/zuul/buildset/f120eaadda9344ab8efe4c0db053e15d

PR #385, opened 2023-12-07
Clean up jinja2 conditionals in config template

ASBishop:
The parent glance controller needs a minimal config that is adequate for generating jobs for DB sync and cronjobs. This patch uses a new "MinimalConfig" environment variable to support that goal, which eliminates the need to wrap every glanceapi variable with its own conditional.


fmount (approved):
This is a good optimization, with an additional bool in the top-level controller we can avoid most of the checks in the `00-config` template.  We really need the `database` connection for dbsync an cronjobs, hence I'm +1 to this kind of change.
I also checked the logs and they look good: the minimal config results in:

```
[DEFAULT]
verbose=True
[database]
connection = **********
```

fmount:
I'll rebase the barbican patch on top of this, so we don't have to do any additional work there. Thanks @ASBishop for the help!

PR #393, opened 2023-12-17
Update the repo doc and improve the design choices section

fmount:
As a follow up of the `glanceAPI` list feature, this change introduces a doc change to explain the `design choices` behind this new deployment model. It also explains part of the feature from a human operator perspective, describing what is supported (and what is not supported) in this new scenario.


docs/dev/design-decisions.md
```diff
@@ -142,52 +146,81 @@ glance   True     Setup complete
...
-To do that, a Glance CR spec like the following should be created:
+The resulting CRs can be inspected as regular k8s objects.
+The following diagram describes how k8s resources/object are connected.
 
 ```
-...
+                                                      [statefulset(s)]
+                                                +-------------------------+
++------------+       +------------------+ |---> | glance-default-internal |
+| glance CR  | |---> | glance-default CR|       +-------------------------+
```
abays, line 161: Maybe `glance-default CR` should be `glanceAPI-default CR` to clearly indicate that it's a `GlanceAPI` CR?

fmount: Makes sense, it also helps to properly identify the objects created. Thanks, I updated it!

abays (approved):
/lgtm

PR #394, opened 2023-12-17
Do not report network related errors if replicas=0

fmount:
When `replicas` is set to 0 for a particular (or a set of) `glanceAPI(s)`, do not report the network error related about missing Pods network interfaces.

### DIFF
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -600,7 +600,7 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
 	}
 
 	instance.Status.NetworkAttachments = networkAttachmentStatus
-	if networkReady {
+	if networkReady || *instance.Spec.Replicas == 0 {
 		instance.Status.Conditions.MarkTrue(condition.NetworkAttachmentsReadyCondition, condition.NetworkAttachmentsReadyMessage)
 	} else {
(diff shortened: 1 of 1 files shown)


karelyatin:
/lgtm

it would be good if fixed in lib-common itself VerifyNetworkStatusFromAnnotation so it's more generic fix.

fmount:
> /lgtm
> 
> it would be good if fixed in lib-common itself VerifyNetworkStatusFromAnnotation so it's more generic fix.

Right, this is something we were discussing last week, so this is more a short/mid term fix until we propose a lib-common fix that works for all the operators.

konan-abhi (approved):

abays (approved):
/lgtm

abays:
> > /lgtm
> > it would be good if fixed in lib-common itself VerifyNetworkStatusFromAnnotation so it's more generic fix.
> 
> Right, this is something we were discussing last week, so this is more a short/mid term fix until we propose a lib-common fix that works for all the operators.

I was thinking about this and what we would do to the `lib-common` `VerifyNetworkStatusFromAnnotation` function.  It seems like we'd have to change the signature and pass something like `requestedReplicas` or whatever.  Then in the function body we'd just return true if `requestedReplicas < 1` and not proceed with anything else.  But then it seems to me that the caller should just check this and not make the call in the first place (which is what this PR is doing).  Maybe it's still better to have the logic in one common place though.

PR #401, opened 2024-01-08
Optimize route ingress create

stuggi:
A route should get created for the glance-api instance which has
- annotation with glance.KeystoneEndpoint -> true
- it is the service.EndpointPublic
- and the k8s service is corev1.ServiceTypeClusterIP

Right now the service.AnnotationIngressCreateKey: "true" Annotation gets set on all glance-api services.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/432


abays (commented):
+1 from me but will defer to @fmount for approval

fmount:
> +1 from me but will defer to @fmount for approval

As discussed, the change itself looks good and I like the idea of simplifying the `openstack-operator` logic and move the control here.
I'll test it locally and comment further! Thanks @stuggi for this patch!

fmount:
I had the chance to test this patch in combination with https://github.com/openstack-k8s-operators/openstack-operator/pull/615 and it works as expected: in case we update the routes we don't have any leftover of the previous api that was registered in keystone, so I guess we can move forward with these patches.

Thanks @stuggi, feel free to /unhold when it's ready to merge.

fmount (approved):
/lgtm

PR #417, opened 2024-01-24
Detect invalid backend configuration

fmount:
When the `GlanceAPI` is split between `internal` and `external` instances, if `File` is used as a `backend` `Glance` won't work as expected. 
Each instance is going to have its own `RWO` `pvc` where the images are stored, and it might happen that while an image is uploaded via the external instance, an application might request it via the internal instance (which is associated to the `keystone` internal endpoint), resulting in a failure because `no data are associated to the requested image`.
For the reason above, when the `File` backend is used, the operator should prevent the `API` split (`type: single`), or block any attempt to use the `File` backend with this layout.
This patch introduces a `webhook` related improvement that validates the layout and fails if some criteria are not met.
In particular, if a `split` (internal + external instances) layout is used with a `File` backend, the `ValidateCreate()` function raises an exception and prevents the deploy.
Other than that, we know that if an `empty spec` is applied, a `default` `GlanceAPI` instance is created.
The `Default()` function now enforces a `split` layout, it sets `replicas: 0` and creates a `CHANGE_ME` dummy backend like the following:

```
[DEFAULT]
enabled_backends=backend1:type1 # CHANGE_ME
```

This will force the human operator to think in advance about the backend that should be associated to `Glance` as we don't want File to be the default.
Both the samples and the functional tests are aligned with the new webhook functions to make sure we don't get a `Detected an invalid backend` issue.

Depends-on: https://github.com/openstack-k8s-operators/install_yamls/pull/706


konan-abhi (approved):
Looks good to me!

test/functional/glance_controller_test.go
```diff
@@ -239,13 +244,13 @@ var _ = Describe("Glance controller", func() {
...
-			keystone.SimulateKeystoneEndpointReady(glanceTest.GlanceExternal)
+			keystone.SimulateKeystoneEndpointReady(glanceTest.GlanceSingle)
 		})
 		It("Creates glanceAPI", func() {
 			// Default type is "split", make sure that behind the scenes two
 			// glanceAPI deployment are created
-			GlanceAPIExists(glanceTest.GlanceExternal)
-			GlanceAPIExists(glanceTest.GlanceInternal)
+			GlanceAPIExists(glanceTest.GlanceSingle)
+			//GlanceAPIExists(glanceTest.GlanceInternal)
```
abays, line 253: Should the comment here by changed to reflect the test code changes?

fmount: @abays thank you for the review! So yeah, good catch, I'm going to update it in my next update.

api/v1beta1/glance_webhook.go
```diff
@@ -142,6 +205,11 @@ func (r *Glance) ValidateUpdate(old runtime.Object) error {
 		if glanceAPI.Type != o.Spec.GlanceAPIs[key].Type {
 			return errors.New("GlanceAPI deployment layout can't be updated")
 		}
+		// Fail if an invalid configuration/layout is detected for the current
+		// glanceAPI instance
+		if r.isInvalidBackend(glanceAPI, topLevelFileBackend) {
```
fmount, line 210: @abays even during an Update of the existing CR, we want to make sure that we don't allow to have File + Split when replicas > 0, so I reproposed the same check to this function (and moved `isInvalidBackend` into its own func).

api/v1beta1/glance_webhook.go
```diff
@@ -105,6 +120,41 @@ func (spec *GlanceSpec) Default() {
...
+}
+
+// Check if the File is used in combination with a wrong layout
+func (r *Glance) isInvalidBackend(glanceAPI GlanceAPITemplate, topLevel bool) bool {
+	var rep int32 = 0
+	// For each current glanceAPI instance, detect an invalid configuration
+	// made by "type: split && backend: file": raise an issue if this config
+	// is found. However, do not fail if 'replica: 0' because it means the
+	// operator has not made any choice about the backend yet
+	if (*glanceAPI.Replicas != rep && glanceAPI.Type == "split" && isFileBackend(glanceAPI.CustomServiceConfig, topLevel)) {
```
fmount, line 152: During the ctlplane CR creation, we do not want to fail if we set the `type` to split but at the same time `Replicas` is 0.
If we look at VA2 [1], we have this kind of situation where no backend has been chosen yet, no `customConfigService` is passed and `replicas: 0`.
We need, in that case, the `ospctlplane` to converge into a working state, and I think it makes sense evaluate a condition that takes into account the number of replicas.
@abays wdyt?

abays: Yes, that makes sense to me

abays (approved):
/lgtm

PR #422, opened 2024-01-26
Update the NFS backend example

fmount:
This patch introduces the missing documentation for a `NFS` backend.
It includes both an updated `CR` that is consistent with the `extraMounts` usage in the `OpenStackControlPlane`,
and it introduces a script to help testing this backend with `crc`.
There are also instructions and an example that shows how to double check the `NFS` backend is actually used
by the Pod.


konan-abhi (approved):
Looks good, thank you!!

PR #424, opened 2024-01-29
[docs] Sample file to enable rbd thin provisioning

konan-abhi:
Added sample document to show how to enable sparse image upload for ceph backend.

Fixes: [OSPRH-1231](https://issues.redhat.com/browse/OSPRH-1231)

### DIFF
--- a/config/samples/backends/README.md
+++ b/config/samples/backends/README.md
@@ -20,6 +20,7 @@ Currently available samples are:
 - Ceph
 - NFS
 - CEPH + NFS
+- CEPH + Sparse Image Upload
 - Cinder backends
 - Swift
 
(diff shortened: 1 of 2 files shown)


fultonj (commented):
This looks good by my reading (I didn't test it directly though).

config/samples/backends/README.md
```diff
@@ -20,6 +20,7 @@ Currently available samples are:
 - Ceph
 - NFS
 - CEPH + NFS
+- CEPH + Sparse Image Upload
```
fmount, line 23: Thank you for adding this particular sample!

config/samples/backends/ceph_thin_provisioning/ceph.yaml
```diff
@@ -0,0 +1,50 @@
+# Sample using Ceph as a glance backend
```
fmount, line 1: Given that this sample is a variation of the basic `backend/ceph` (we are only adding the `rbd_thin_provisioning = True` config option), I think we can remove this file and point to the other one via `kustomization.yaml`. This would reduce duplication and we only have to maintain a basic Ceph sample.

config/samples/backends/ceph_thin_provisioning/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+resources:
```
fmount, line 1: As per my previous comment, you can remove `ceph.yaml`, and replace the content of `kustomization.yaml` with the follwing:

```
resources:
- ../ceph/

patches:
- target:
    kind: OpenStackControlPlane
  patch: |-
    - op: replace
      path: /spec/glance/template/customServiceConfig
      value: |
        [DEFAULT]
        enabled_backends = default_backend:rbd
        [glance_store]
        default_backend = default_backend
        [default_backend]
        rbd_store_ceph_conf = /etc/ceph/ceph.conf
        store_description = "RBD backend"
        rbd_store_pool = images
        rbd_store_user = openstack
        rbd_thin_provisioning = True
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

```

By doing this we can `kustomize build ceph_thin_provisioning` and get the right control plane CR.

fmount (changes requested):

fmount (approved):
/lgtm

fmount:
pre-commit failure that can be solved by running `pre-commit run --all-files --show-diff-on-failure`

PR #426, opened 2024-01-29
migrate from databaseUsername to databaseAccount and fully use MariaDBAccount

zzzeek:
This moves glance to fully use MariaDBAccount based on the dev work being done for mariadb-operator:

Lead Jira: [OSPRH-4095](https://issues.redhat.com/browse/OSPRH-4095)

1. [x] controller calls EnsureMariaDBAccount up front to make sure MariaDBAccount is present
2. [x] error code from EnsureMariaDBAccount is handled, Conditions are amended when error is returned
3. [x] controller calls NewDatabaseForAccount instead of NewDatabase
4. [x]  GetAccountAndSecret is used to retrieve account /secret to populate template
5. [x]  GetDatabaseByName() , normally used for delete finalizers, replaced with GetDatabaseByNameAndAccount
6. [x]  CreateOrPatchAll() used to patch the Database, replacing CreateOrPatchDB / CreateOrPatchDBByName
7. [x]  controller calls DeleteUnusedMariaDBAccountFinalizers when launched pods are definitely running on a new MariaDBAccount, returns error code if present
8. [x]  PasswordSelectors that refer to database are removed
9. [x]  all databaseUser replaced with databaseAccount inside of all XYZ_types.go
10. [x] all databaseUser replaced with databaseAccount inside of all `kuttl/*.yaml`
11. [x] all default databaseUser names become databaseAccount, replacing underscores with dashes inside of all XYZ_types.go
12. [x] all default databaseUser names become databaseAccount, replacing underscores with dashes inside of all `kuttl/*.yaml`
13. [x] MariaDBAccountSuiteTests are used in controller ginkgo tests if it has them
14. [x] Use configsecrets for database URLs; remove from job hash - was already like this
15. [x] [184](https://github.com/openstack-k8s-operators/mariadb-operator/pull/184) is merged and replaces from go.mod are removed

### DIFF
--- a/api/v1beta1/glance_types.go
+++ b/api/v1beta1/glance_types.go
@@ -50,9 +50,8 @@ type GlanceSpecCore struct {
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=glance
-	// DatabaseUser - optional username used for glance DB, defaults to glance
-	// TODO: -> implement needs work in mariadb-operator, right now only glance
-	DatabaseUser string `json:"databaseUser"`
+	// DatabaseAccount - name of MariaDBAccount which will be used to connect.
+	DatabaseAccount string `json:"databaseAccount"`
 
 	// +kubebuilder:validation:Required
 	// +kubebuilder:default=memcached
@@ -60,11 +59,12 @@ type GlanceSpecCore struct {
 	MemcachedInstance string `json:"memcachedInstance"`
 
 	// +kubebuilder:validation:Required
-	// Secret containing OpenStack password information for glance GlanceDatabasePassword
+	// Secret containing OpenStack password information for glance's keystone
+	// password; no longer used for database password
 	Secret string `json:"secret"`
 
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default={database: GlanceDatabasePassword, service: GlancePassword}
+	// +kubebuilder:default={service: GlancePassword}
 	// PasswordSelectors - Selectors to identify the DB and ServiceUser password from the Secret
 	PasswordSelectors PasswordSelector `json:"passwordSelectors"`
 
@@ -140,11 +140,6 @@ type GlanceSpec struct {
 // PasswordSelector to identify the DB and AdminUser password from the Secret
 type PasswordSelector struct {
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default="GlanceDatabasePassword"
-	// Database - Selector to get the glance database user password from the Secret
-	// TODO: not used, need change in mariadb-operator
-	Database string `json:"database"`
-	// +kubebuilder:validation:Optional
 	// +kubebuilder:default="GlancePassword"
 	// Service - Selector to get the glance service password from the Secret
 	Service string `json:"service"`
@@ -162,7 +157,6 @@ type DBPurge struct {
 	//Schedule defines the crontab format string to schedule the DBPurge cronJob
 	Schedule string `json:"schedule"`
 }
-
 
 // GlanceStatus defines the observed state of Glance
 type GlanceStatus struct {
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
 	rbacv1 "k8s.io/api/rbac/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
@@ -300,7 +299,7 @@ func (r *GlanceReconciler) reconcileDelete(ctx context.Context, instance *glance
 	r.Log.Info(fmt.Sprintf("Reconciling Service '%s' delete", instance.Name))
 
 	// remove db finalizer first
-	db, err := mariadbv1.GetDatabaseByName(ctx, helper, instance.Name)
+	db, err := mariadbv1.GetDatabaseByNameAndAccount(ctx, helper, glance.DatabaseName, instance.Spec.DatabaseAccount, instance.Namespace)
 	if err != nil && !k8s_errors.IsNotFound(err) {
 		return ctrl.Result{}, err
 	}
@@ -678,6 +677,12 @@ func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glance
 		return ctrl.Result{}, err
 	}
 
+	// remove finalizers from unused MariaDBAccount records
+	err = mariadbv1.DeleteUnusedMariaDBAccountFinalizers(ctx, helper, instance.Name, instance.Spec.DatabaseAccount, instance.Namespace)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
 	// create DBPurge CronJob
 
 	// DBPurge is not optional and always created to purge all soft deleted
@@ -846,7 +851,7 @@ func (r *GlanceReconciler) apiDeploymentCreateOrUpdate(
 		GlanceAPITemplate: apiTemplate,
 		APIType:           apiType,
 		DatabaseHostname:  instance.Status.DatabaseHostname,
-		DatabaseUser:      instance.Spec.DatabaseUser,
+		DatabaseAccount:   instance.Spec.DatabaseAccount,
 		Secret:            instance.Spec.Secret,
 		ExtraMounts:       instance.Spec.ExtraMounts,
 		PasswordSelectors: instance.Spec.PasswordSelectors,
@@ -933,18 +938,16 @@ func (r *GlanceReconciler) generateServiceConfig(
 ) error {
 	labels := labels.GetLabels(instance, labels.GetGroupLabel(glance.ServiceName), serviceLabels)
 
-	ospSecret, _, err := secret.GetSecret(ctx, h, instance.Spec.Secret, instance.Namespace)
-	if err != nil {
-		return err
-	}
+	databaseAccount := db.GetAccount()
+	dbSecret := db.GetSecret()
 
 	// We only need a minimal 00-config.conf that is only used by db-sync job,
 	// hence only passing the database related parameters
 	templateParameters := map[string]interface{}{
 		"MinimalConfig": true, // This tells the template to generate a minimal config
 		"DatabaseConnection": fmt.Sprintf("mysql+pymysql://%s:%s@%s/%s?read_default_file=/etc/my.cnf",
-			instance.Spec.DatabaseUser,
-			string(ospSecret.Data[instance.Spec.PasswordSelectors.Database]),
+			databaseAccount.Spec.UserName,
+			string(dbSecret.Data[mariadbv1.DatabasePasswordSelector]),
 			instance.Status.DatabaseHostname,
 			glance.DatabaseName,
 		),
@@ -1130,24 +1133,45 @@ func (r *GlanceReconciler) ensureDB(
 	h *helper.Helper,
 	instance *glancev1.Glance,
 ) (*mariadbv1.Database, ctrl.Result, error) {
-	//
+	// ensure a MariaDBAccount CR and Secret exist.
+	// this mariadb API function will as an interim step actually generate a
+	// new MariaDBAccount and Secret if one does not exist already.   in a
+	// future release, this function may change to emit an error if the
+	// MariaDBAccount was not already created ahead of time (e.g. by openstack-operator
+	// or end-user YAML declaration)
+	_, _, err := mariadbv1.EnsureMariaDBAccount(
+		ctx, h, instance.Spec.DatabaseAccount,
+		instance.Namespace, false, "glance",
+	)
+
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			mariadbv1.MariaDBAccountReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			mariadbv1.MariaDBAccountNotReadyMessage,
+			err.Error()))
+
+		return nil, ctrl.Result{}, err
+	}
+	instance.Status.Conditions.MarkTrue(
+		mariadbv1.MariaDBAccountReadyCondition,
+		mariadbv1.MariaDBAccountReadyMessage,
+	)
+
 	// create service DB instance
 	//
-	db := mariadbv1.NewDatabase(
-		instance.Name,
-		instance.Spec.DatabaseUser,
-		instance.Spec.Secret,
-		map[string]string{
-			"dbName": instance.Spec.DatabaseInstance,
-		},
+	db := mariadbv1.NewDatabaseForAccount(
+		instance.Spec.DatabaseInstance, // mariadb/galera service to target
+		glance.DatabaseName,            // name used in CREATE DATABASE in mariadb
+		glance.DatabaseName,            // CR name for MariaDBDatabase
+		instance.Spec.DatabaseAccount,  // CR name for MariaDBAccount
+		instance.Namespace,             // namespace
 	)
 
 	// create or patch the DB
-	ctrlResult, err := db.CreateOrPatchDBByName(
-		ctx,
-		h,
-		instance.Spec.DatabaseInstance,
-	)
+	ctrlResult, err := db.CreateOrPatchAll(ctx, h)
+
 	if err != nil {
 		instance.Status.Conditions.Set(condition.FalseCondition(
 			condition.DBReadyCondition,
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -827,7 +827,7 @@ func (r *GlanceAPIReconciler) generateServiceConfig(
 ) error {
 	labels := labels.GetLabels(instance, labels.GetGroupLabel(glance.ServiceName), GetServiceLabels(instance))
 
-	db, err := mariadbv1.GetDatabaseByName(ctx, h, glance.DatabaseName)
+	db, err := mariadbv1.GetDatabaseByNameAndAccount(ctx, h, glance.DatabaseName, instance.Spec.DatabaseAccount, instance.Namespace)
 	if err != nil {
 		return err
 	}
@@ -874,6 +874,9 @@ func (r *GlanceAPIReconciler) generateServiceConfig(
 		return err
 	}
 
+	databaseAccount := db.GetAccount()
+	dbSecret := db.GetSecret()
+
 	glanceEndpoints := glanceapi.GetGlanceEndpoints(instance.Spec.APIType)
 	httpdVhostConfig := map[string]interface{}{}
 	for endpt := range glanceEndpoints {
@@ -894,8 +897,8 @@ func (r *GlanceAPIReconciler) generateServiceConfig(
 		"KeystoneInternalURL": keystoneInternalURL,
 		"KeystonePublicURL":   keystonePublicURL,
 		"DatabaseConnection": fmt.Sprintf("mysql+pymysql://%s:%s@%s/%s?read_default_file=/etc/my.cnf",
-			instance.Spec.DatabaseUser,
-			string(ospSecret.Data[instance.Spec.PasswordSelectors.Database]),
+			databaseAccount.Spec.UserName,
+			string(dbSecret.Data[mariadbv1.DatabasePasswordSelector]),
 			instance.Spec.DatabaseHostname,
 			glance.DatabaseName,
 		),
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/openstack-k8s-operators/lib-common/modules/openstack v0.3.1-0.20240229121803-169ced56d56e
 	github.com/openstack-k8s-operators/lib-common/modules/storage v0.3.1-0.20240229121803-169ced56d56e
 	github.com/openstack-k8s-operators/lib-common/modules/test v0.3.1-0.20240229121803-169ced56d56e
-	github.com/openstack-k8s-operators/mariadb-operator/api v0.3.1-0.20240222094307-76fef735f093
+	github.com/openstack-k8s-operators/mariadb-operator/api v0.3.1-0.20240303091826-438dde8600d3
 	golang.org/x/exp v0.0.0-20240213143201-ec583247a57a
 	gopkg.in/yaml.v3 v3.0.1
 	k8s.io/api v0.28.7
--- a/test/functional/glance_controller_test.go
+++ b/test/functional/glance_controller_test.go
@@ -27,7 +27,9 @@ import (
 	memcachedv1 "github.com/openstack-k8s-operators/infra-operator/apis/memcached/v1beta1"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	util "github.com/openstack-k8s-operators/lib-common/modules/common/util"
+	mariadb_test "github.com/openstack-k8s-operators/mariadb-operator/api/test/helpers"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	ptr "k8s.io/utils/ptr"
 )
 
@@ -67,7 +69,7 @@ var _ = Describe("Glance controller", func() {
 		It("initializes Spec fields", func() {
 			Glance := GetGlance(glanceTest.Instance)
 			Expect(Glance.Spec.DatabaseInstance).Should(Equal("openstack"))
-			Expect(Glance.Spec.DatabaseUser).Should(Equal(glanceTest.GlanceDatabaseUser))
+			Expect(Glance.Spec.DatabaseAccount).Should(Equal(glanceTest.GlanceDatabaseAccount.Name))
 			Expect(Glance.Spec.ServiceUser).Should(Equal(glanceTest.GlanceServiceUser))
 			Expect(Glance.Spec.MemcachedInstance).Should(Equal(glanceTest.MemcachedInstance))
 			// No Keystone Quota is present, check the default is 0
@@ -155,8 +157,8 @@ var _ = Describe("Glance controller", func() {
 			)
 		})
 		It("Should set DBReady Condition and set DatabaseHostname Status when DB is Created", func() {
-			mariadb.SimulateMariaDBAccountCompleted(glanceTest.Instance)
-			mariadb.SimulateMariaDBDatabaseCompleted(glanceTest.Instance)
+			mariadb.SimulateMariaDBDatabaseCompleted(glanceTest.GlanceDatabaseName)
+			mariadb.SimulateMariaDBAccountCompleted(glanceTest.GlanceDatabaseAccount)
 			th.SimulateJobSuccess(glanceTest.GlanceDBSync)
 			Glance := GetGlance(glanceTest.Instance)
 			Expect(Glance.Status.DatabaseHostname).To(Equal(fmt.Sprintf("hostname-for-openstack.%s.svc", namespace)))
@@ -181,8 +183,8 @@ var _ = Describe("Glance controller", func() {
 			)
 		})
 		It("Should fail if db-sync job fails when DB is Created", func() {
-			mariadb.SimulateMariaDBAccountCompleted(glanceTest.Instance)
-			mariadb.SimulateMariaDBDatabaseCompleted(glanceTest.Instance)
+			mariadb.SimulateMariaDBDatabaseCompleted(glanceTest.GlanceDatabaseName)
+			mariadb.SimulateMariaDBAccountCompleted(glanceTest.GlanceDatabaseAccount)
 			th.SimulateJobFailure(glanceTest.GlanceDBSync)
 			th.ExpectCondition(
 				glanceTest.Instance,
@@ -216,8 +218,8 @@ var _ = Describe("Glance controller", func() {
 					},
 				),
 			)
-			mariadb.SimulateMariaDBAccountCompleted(glanceTest.Instance)
-			mariadb.SimulateMariaDBDatabaseCompleted(glanceTest.Instance)
+			mariadb.SimulateMariaDBDatabaseCompleted(glanceTest.GlanceDatabaseName)
+			mariadb.SimulateMariaDBAccountCompleted(glanceTest.GlanceDatabaseAccount)
 			th.SimulateJobSuccess(glanceTest.GlanceDBSync)
 			keystoneAPI := keystone.CreateKeystoneAPI(glanceTest.Instance.Namespace)
 			DeferCleanup(keystone.DeleteKeystoneAPI, keystoneAPI)
@@ -278,8 +280,8 @@ var _ = Describe("Glance controller", func() {
 				),
 			)
 			DeferCleanup(keystone.DeleteKeystoneAPI, keystone.CreateKeystoneAPI(glanceTest.Instance.Namespace))
-			mariadb.SimulateMariaDBAccountCompleted(glanceTest.Instance)
-			mariadb.SimulateMariaDBDatabaseCompleted(glanceTest.Instance)
+			mariadb.SimulateMariaDBDatabaseCompleted(glanceTest.GlanceDatabaseName)
+			mariadb.SimulateMariaDBAccountCompleted(glanceTest.GlanceDatabaseAccount)
 			th.SimulateJobSuccess(glanceTest.GlanceDBSync)
 			keystone.SimulateKeystoneServiceReady(glanceTest.Instance)
 			keystone.SimulateKeystoneEndpointReady(glanceTest.GlanceSingle)
@@ -313,8 +315,8 @@ var _ = Describe("Glance controller", func() {
 				),
 			)
 			DeferCleanup(keystone.DeleteKeystoneAPI, keystone.CreateKeystoneAPI(glanceTest.Instance.Namespace))
-			mariadb.SimulateMariaDBAccountCompleted(glanceTest.Instance)
-			mariadb.SimulateMariaDBDatabaseCompleted(glanceTest.Instance)
+			mariadb.SimulateMariaDBDatabaseCompleted(glanceTest.GlanceDatabaseName)
+			mariadb.SimulateMariaDBAccountCompleted(glanceTest.GlanceDatabaseAccount)
 			th.SimulateJobSuccess(glanceTest.GlanceDBSync)
 			keystone.SimulateKeystoneServiceReady(glanceTest.Instance)
 			keystone.SimulateKeystoneEndpointReady(glanceTest.GlanceSingle)
@@ -353,6 +355,7 @@ var _ = Describe("Glance controller", func() {
 				"storageRequest":      glanceTest.GlancePVCSize,
 				"secret":              SecretName,
 				"databaseInstance":    "openstack",
+				"databaseAccount":     glanceTest.GlanceDatabaseAccount.Name,
 				"keystoneEndpoint":    "default",
 				"customServiceConfig": GlanceDummyBackend,
 				"glanceAPIs": map[string]interface{}{
@@ -376,8 +379,8 @@ var _ = Describe("Glance controller", func() {
 					},
 				),
 			)
-			mariadb.SimulateMariaDBAccountCompleted(glanceTest.Instance)
-			mariadb.SimulateMariaDBDatabaseCompleted(glanceTest.Instance)
+			mariadb.SimulateMariaDBDatabaseCompleted(glanceTest.GlanceDatabaseName)
+			mariadb.SimulateMariaDBAccountCompleted(glanceTest.GlanceDatabaseAccount)
 			th.SimulateJobSuccess(glanceTest.GlanceDBSync)
 			keystoneAPI := keystone.CreateKeystoneAPI(glanceTest.Instance.Namespace)
 			DeferCleanup(keystone.DeleteKeystoneAPI, keystoneAPI)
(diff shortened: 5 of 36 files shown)


go.mod
```diff
@@ -2,6 +2,9 @@ module github.com/openstack-k8s-operators/glance-operator
 
 go 1.19
 
+replace github.com/openstack-k8s-operators/mariadb-operator/api => /home/classic/dev/openshift/mariadb-operator/api
```
zzzeek, line 5: I've only run this locally so far.   hmmm I guess I can link to the branch but i wanted to get feedback on the code before we try to test this etc.

gibizer: I think it make sense to link it to your mariadb PR and see the test results.

gibizer (commented):
Looks good to me

go.mod
zzzeek: how do you do that?  I saw you did it the other day but I cant find that commit now

zzzeek: got it, hard to google for

zzzeek: OK this is not passing any tests if you could show me how to do this that would be helpful

fmount: I suspect you should go get https://github.com/zzzeek/mariadb-operator/api@integrate_oo_accounts  and put here `https://github.com/zzzeek/mariadb-operator/api <the resulting hash>`

I didn't review this patch entirely but it would be a good start to have something that passes checks, so we can use logs to understand the bits involved here.
The problem I see is that the job that is supposed to test this particular concatenation of patches can only rebuild glance and the openstack-operator, so we should inject the right mariadb image that is able to handle the new code (not sure we're able to do that in CI right now).
Maybe we can do that by patching the openstack-operator (where we inject an arbitrary mariadb-operator image) and add the `Depends-On` to this patch.
That probably would allow the test the whole chain.
The steps should be:
1. rebuild the operator based on https://github.com/openstack-k8s-operators/mariadb-operator/pull/184 and push it to a private quay.io (make sure to set the image as public) (together w/ the associated bundle and index)
2. hack the openstack-operator script (https://github.com/openstack-k8s-operators/openstack-operator/blob/main/hack/pin-bundle-images.sh) to build it using your quay.io bundle pushed at step1
3. add to this patch a `Depends-on` to the openstack-operator patch created at step2.

I did something similar in the past to just see the CI green, and those are pretty much the steps (we might need a variation of them though).

zzzeek: the go get approach gives me this error:

```
$ go get github.com/zzzeek/mariadb-operator/api@integrate_oo_accounts
go: downloading github.com/zzzeek/mariadb-operator/api v0.3.2
go: downloading github.com/zzzeek/mariadb-operator v0.3.2
go: github.com/zzzeek/mariadb-operator/api@v0.3.2: parsing go.mod:
        module declares its path as: github.com/openstack-k8s-operators/mariadb-operator/api
                but was required as: github.com/zzzeek/mariadb-operator/api

```

The changes on the mariadb-operator side are just new API additions within the `mariadbdatabase_funcs.go` file, and otherwise dont change how that operator works at all, so we may consider that I'd test this some more locally enough that we can merge just the mariadb side of it.    I can go the quay approach here but this looks like a steep hill to climb

zzzeek:
so to get a build to pass:

a. we aren't allowed to have replace directives in go.mod, which means
b. I have to "go get" this against the repo over at https://github.com/zzzeek/mariadb-operator/commits/integrate_oo_accounts , which means
c. I have to modify go.mod in base and api/ at https://github.com/zzzeek/mariadb-operator/commits/integrate_oo_accounts to have `module github.com/zzzeek/mariadb-operator` at the top, rather than openstack-operators, which means....
d. I have to change every source file in the mariadb operator to have that new path...and this seems like not what im supposed to be doing

test/functional/suite_test.go
```diff
@@ -243,4 +245,35 @@ var _ = BeforeEach(func() {
 	DeferCleanup(th.DeleteNamespace, namespace)
 	//Let's create the osp-secret in advance (in common to all the test cases)
 	DeferCleanup(k8sClient.Delete, ctx, CreateGlanceSecret(glanceName.Namespace, SecretName))
+
+	// Create MariaDBAccount and Secret.
```
zzzeek, line 249: so I got functional tests to pass (:tada: ) but I had to do some dark magic to get there.

Obviously I will add a "CreateMariaDBAccount" utility function to mariadb/api/test/crd.  just calling that here might be good enough for now, but the base controller here does create this account object as well.

fmount: This is a great news, I understand it's hard to get a full e2e deployment with dependencies across the operators, so I think functional tests represent a great start.
Thanks for your effort and I'll review it in a bit.

test/functional/glance_controller_test.go
```diff
@@ -158,6 +165,10 @@ var _ = Describe("Glance controller", func() {
 		})
 		It("Should fail if db-sync job fails when DB is Created", func() {
 			mariadb.SimulateMariaDBDatabaseCompleted(glanceTest.Instance)
+			mariadb.SimulateMariaDBAccountCompleted(types.NamespacedName{
```
fmount, line 168: in `glance_test_data.go` we can have something like:

```golang
GlanceDatabaseAccount : types.NamespacedName{
	Namespace: glanceName.Namespace,
	Name:      GlanceServiceUser,
},
```

and we reference it as `glanceTest.GlanceDatabaseAccount` in any `mariadb.SimulateMariaDBAccountCompleted` call.

test/functional/glance_controller_test.go
```diff
@@ -190,6 +201,10 @@ var _ = Describe("Glance controller", func() {
 				),
 			)
 			mariadb.SimulateMariaDBDatabaseCompleted(glanceTest.Instance)
```
fmount, line 203: Is `mariadb.SimulateMariaDBDatabaseCompleted` required as long as we're switching to `mariadb.SimulateMariaDBAccountCompleted` ? (The same comment applies to all the occurrences).

test/functional/glance_test_data.go
```diff
@@ -43,7 +43,7 @@ const (
 // GlanceTestData is the data structure used to provide input data to envTest
 type GlanceTestData struct {
 	ContainerImage              string
-	GlanceDatabaseUser          string
+	GlanceDatabaseAccount       string
```
fmount, line 46: See my previous comment: if this is turned into a `types.NamespacedName` and defined with:

```golang
GlanceDatabaseAccount : types.NamespacedName{
	Namespace: glanceName.Namespace,
	Name:      GlanceServiceUser,
},
```
we can reduce the code and we still have the chance to compare the expected name by accessing `glanceTest.GlanceDatabaseAccount.Name`.

test/functional/suite_test.go
```diff
@@ -243,4 +245,35 @@ var _ = BeforeEach(func() {
...
+	// both controllers rely on reading the values from these objects.
+	// TODO: this set of features should likely be in mariadb-operator/api/test/crd.
```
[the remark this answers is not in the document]
fmount: Also, I'm wondering how we're thinking about this in the long term: keep in mind that both `functional` tests and `kuttl` are not expecting a full deployed `OpenStackControlPlane`, as only the service operator CRs are generated (and the minimal amount of dependencies (mariadb/galera, rabbit, keystone, infra components).
I assume we can have an entry in `install_yamls` that generates the CRs expected as output from the openstack-operator, but I see that right now we're not there yet.
@zzzeek should we capture this conversation somewhere outside of this PR?

api/v1beta1/glance_types.go
```diff
@@ -52,12 +52,16 @@ type GlanceSpec struct {
...
 	// +kubebuilder:default=glance
-	// DatabaseUser - optional username used for glance DB, defaults to glance
-	// TODO: -> implement needs work in mariadb-operator, right now only glance
-	DatabaseUser string `json:"databaseUser"`
+	// DatabaseAccount - name of MariaDBAccount which will be used to connect.
+	// As an interim solution, glance-controller will
+	// generate the MariaDBAccount object of the given name if it's not
+	// present in the cluster.  Later, openstack-operator will perform this
+	// generation and also populate this field for the Glance objects it creates
+	DatabaseAccount string `json:"databaseAccount"`
```
fmount, line 60: +1 to switch from DatabaseUser to DatabaseAccount.

controllers/glance_controller.go
```diff
@@ -886,7 +889,39 @@ func (r *GlanceReconciler) generateServiceConfig(
...
+	// database.  The account
+	// has not yet been linked to a MariaDBDatabase, so it will only exist as a
+	// CR in the cluster.  Once linked to a MariaDBDatabase, the mariadb-operator
+	// will ensure it exists with rights for that database.
+	//
+	// This is an **interim** step, where later openstack-operator will
+	// ensure that MariaDBAccounts are created for all the services it
+	// initializes, and this block would no longer be present here.
+	// At that point the DatabaseAccount field can be set to be required.
+
```
fmount, line 905: So, iiuc the flow here is:

1. `MariaDBAccount` CR is created via `mariadbv1.InterimCreateNewAccount`: the `mariadb-operator` watches for this kind of CRs and "translates" them into an account/database with the right permissions. A secret with the credentials is generated by mariadb and can be "consumed" by Glance;
2. We execute `mariadbv1.GetAccount`  to retrieve the associated CR, selected by `instance.Spec.DatabaseAccount`
3. Finally, we can get the secret via the regular `secret.GetSecret`, and it's selected by name via `databaseAccount.Spec.Secret`

Step 1 won't be required as long as `openstack-operator` will generate (in advance) the `MariaDBAccount` CRs for each deployed service.
As long as the CR exist, then I think we can still run the service as a standalone component w/o the openstack-operator requirement, which is good.
@zzzeek is my understanding correct?

fmount (changes requested):
No special concerns to the overall code. We can still do some optimizations but it's a minor thing compared to the main goal that sees a switch to the new method.
I'm wondering if it makes sense to provide the `openstack-operator` feature first (create the `MariaDBAccount` CRs for the deployed services), and then update this patch to remove the temporary `mariadbv1.InterimCreateNewAccount` code.
At that point we can just `GetAccount` and have a working CI in terms of deployment/kuttl flows.

test/functional/suite_test.go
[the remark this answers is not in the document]
gibizer: for functional test we already have `SimulateMariaDBAccountCompleted` in mariadb-operator/api/helper but we can add there a function that create a MariaDBAccount

test/functional/glance_controller_test.go
gibizer: I think we need both for now.

test/functional/glance_controller_test.go
zzzeek: yup, will do

controllers/glance_controller.go
zzzeek: that's the flow with a slight extra wrinkle due to how we've so far decided we want the creation of username/password vs. the actual "database" to go:

1. `MariaDBAccount` CR is created via `mariadbv1.InterimCreateNewAccount` - ideally, this would be done by the openstack-operator itself when it first rolls out a new glance instance, or can be specified by the user creating the yaml if they want a specific username/password (this is all part of the requested use case at https://issues.redhat.com/browse/OSPRH-92 ).  but right now it's just up at the very top of the glance-operator flow as an interim architecture (this would be done for all operators).   at the moment this is also where the Glance finalizer is being added to the `MariaDBAccount` CR.
2. The `MariaDBAccount` here does not **yet** refer to a `MariaDBDatabase` - the latter is the CR that indicates that the mariadb-operator would run "CREATE DATABASE <name>" in the galera server.     Because of this, the mariadb-operator will update the status of the CR with "MariaDBDatabase not specified" and will not act upon the record yet.   (technically, you can create an account in mariadb without a specific database, then GRANT later for specific databases, however, we've designed the `MariaDBAccount` CR to always be underneath a single `MariaDBDatabase`, which corresponds more closely and simply to how we use these accounts)
3. the glance controller when it reconciles for "init" then calls `mariadbv1.NewDatabaseForAccount` , referencing the name of the `MariaDBAccount` CR as well as the name of the database, and then `db.CreateOrPatchDBByName`  - the mariadb-operator API creates a `MariaDBDatabase` CR within this run.  This is the same flow that was already here and is in every operator right now.    the Glance finalizer is added to both CRs if not already present.

The change for 3 above is that previously, the `MariaDBDatabase` would be given `osp-secret` directly (and FYI would consume `DatabasePassword` only, not `GlanceDatabasePassword`, e.g. it's a bug), whereas now there's more granularity in that there is a `MariaDBDatabase` and `MariaDBAccount` that's separate.   Multiple `MariaDBAccount` CRs can exist for a single `MariaDBDatabase` and they can also be dropped.

4. Now the mariadb-operator goes to galera and runs CREATE DATABASE for the `MariaDBDatabase` CR and CREATE USER / GRANT for the `MariaDBAccount` CR.
5. the glance operator and glance api operator continues, consuming the **username and secret** from the `MariaDBAccount`.  For accounts to be rotateable, we need the username to also be dynamic.   this username/secret is deployed into config templates to be part of the deployed openstack service.  so "username" is directly in `MariaDBAccount`, and the "secret" then refers to a `Secret` instance that has only the `DatabasePassword` field in it for this account.   Technically you can refer to `osp-secret` here since it has that key within it but it's meant to be a distinct secret
6. The glance and glance api operators as they watch the `Glance` CR will see changes in the `DatabaseAccount` name, pointing to an all new `MariaDBAccount`, which means a new username/password is being rotated in, which will occur from the outside (user intervention, or scheduled).   the controllers will need to wait for that `MariaDBAccount` to be ready, and at that point regenerate pod configs and re-deploy with new credentials.  *then* they need to *search for the old account for their glance DB based on label* and remove the `Glance` finalizer from it, so that the old account will be dropped when it's marked as deleted.

fmount: Thank you for taking the time to explain the flow and the extra interactions I didn't catch!

> that's the flow with a slight extra wrinkle due to how we've so far decided we want the creation of username/password vs. the actual "database" to go:
> 
>     1. `MariaDBAccount` CR is created via `mariadbv1.InterimCreateNewAccount` - ideally, this would be done by the openstack-operator itself when it first rolls out a new glance instance, or can be specified by the user creating the yaml if they want a specific username/password (this is all part of the requested use case at https://issues.redhat.com/browse/OSPRH-92 ).  but right now it's just up at the very top of the glance-operator flow as an interim architecture (this would be done for all operators).   at the moment this is also where the Glance finalizer is being added to the `MariaDBAccount` CR.

Ack, makes sense, and I assume this is a simple call that will be removed/retired when we add that logic to the `openstack-operator`. I think this is a design decision that has an impact to the overall work, but I assume there's already an agreement on this part (centralize the `MariaDBAccount` creation/update at the openstack-operator level).

/cc @dprince in case there are comments or we see drawbacks with this approach.
 

> 
>     2. The `MariaDBAccount` here does not **yet** refer to a `MariaDBDatabase` - the latter is the CR that indicates that the mariadb-operator would run "CREATE DATABASE " in the galera server.     Because of this, the mariadb-operator will update the status of the CR with "MariaDBDatabase not specified" and will not act upon the record yet.   (technically, you can create an account in mariadb without a specific database, then GRANT later for specific databases, however, we've designed the `MariaDBAccount` CR to always be underneath a single `MariaDBDatabase`, which corresponds more closely and simply to how we use these accounts)

ack, so we have a 1:1 between an **active** user and database as an assumption, right? the `MariaDBAccount`, from the glance_controller perspective is a way to "request" the creation of a `user` **and** then a `MariaDBDatabase` CR is associated to an account.  Updates shouldn't be prevented because the rotation consists in rolling out a new `MariaDBAccount` for a given service and update the relation with the `MariaDBDatabase`: by updating the `Glance` CR and pointing (by `name`)  to a new `MariaDBAccount` we can implement the rotation, which consists in repeating the flow you described. Not sure we can simulate the rotation as part of the tests (probably easier via kuttl?)

> 
>     3. the glance controller when it reconciles for "init" then calls `mariadbv1.NewDatabaseForAccount` , referencing the name of the `MariaDBAccount` CR as well as the name of the database, and then `db.CreateOrPatchDBByName`  - the mariadb-operator API creates a `MariaDBDatabase` CR within this run.  This is the same flow that was already here and is in every operator right now.    the Glance finalizer is added to both CRs if not already present.

+1, right, this is already in place, we're just adding the `MariaDBAccount` CR to the picture, which reminds me something we did a while ago for the `rabbitmq-transport-url` CR that produces a `Secret` that we consume in the operator.
Should we need to add something like [1] to ensure we reference a secret that has already been created?

> 
> 
> The change for 3 above is that previously, the `MariaDBDatabase` would be given `osp-secret` directly (and FYI would consume `DatabasePassword` only, not `GlanceDatabasePassword`, e.g. it's a bug), whereas now there's more granularity in that there is a `MariaDBDatabase` and `MariaDBAccount` that's separate. Multiple `MariaDBAccount` CRs can exist for a single `MariaDBDatabase` and they can also be dropped.
> 
>     4. Now the mariadb-operator goes to galera and runs CREATE DATABASE for the `MariaDBDatabase` CR and CREATE USER / GRANT for the `MariaDBAccount` CR.
> 
>     5. the glance operator and glance api operator continues, consuming the **username and secret** from the `MariaDBAccount`.  For accounts to be rotateable, we need the username to also be dynamic.   this username/secret is deployed into config templates to be part of the deployed openstack service.  so "username" is directly in `MariaDBAccount`, and the "secret" then refers to a `Secret` instance that has only the `DatabasePassword` field in it for this account.   Technically you can refer to `osp-secret` here since it has that key within it but it's meant to be a distinct secret

Makes sense, as per your patch we're not consuming the `osp-secret` anymore, but a secret "produced" as a result of the mariadb-operator that trigger the creation logic by processing the CR.

> 
>     6. The glance and glance api operators as they watch the `Glance` CR will see changes in the `DatabaseAccount` name, pointing to an all new `MariaDBAccount`, which means a new username/password is being rotated in, which will occur from the outside (user intervention, or scheduled).   the controllers will need to wait for that `MariaDBAccount` to be ready, and at that point regenerate pod configs and re-deploy with new credentials.  _then_ they need to _search for the old account for their glance DB based on label_ and remove the `Glance` finalizer from it, so that the old account will be dropped when it's marked as deleted.

Yeah, from the `glance-operator` side a change in the CR results in a reconciliation loop that will use the new values, and in that case we already assume that an account/secret already exists.
It might be a limitation on my understanding, but in this flow I see a point where Glance should wait for an operation (the user and db creation) that should happen on the mariadb/galera side, that produces a secret that is consumed by Glance, while before we always have assumed `osp-secret` just exists and it's available. Synchronizing this operation as we've done for `transportUrls` is important, unless I overlooked part of this flow.

[1] https://github.com/openstack-k8s-operators/manila-operator/blob/main/controllers/manila_controller.go#L497

zzzeek:
> No special concerns to the overall code. We can still do some optimizations but it's a minor thing compared to the main goal that sees a switch to the new method. I'm wondering if it makes sense to provide the `openstack-operator` feature first (create the `MariaDBAccount` CRs for the deployed services), and then update this patch to remove the temporary `mariadbv1.InterimCreateNewAccount` code. At that point we can just `GetAccount` and have a working CI in terms of deployment/kuttl flows.

it gets a little confusing and it partially depends on if each downstream operator is built on the v0.3.0 version or the v0.3.1 pseudo-version of mariadb-operator (see https://github.com/openstack-k8s-operators/nova-operator/pull/660 for how this bump looked for nova-operator).    When they are built against 0.3.0, they create a MariaDBDatabase with a "secret" directly in the CR, no MariaDBAccount, and that will create a different job hash than that of the separate MariaDBDatabase / MariaDBAccount.    That would lead to conflicts with a MariaDBAccount that was created by openstack-operator, and the MariaDBAccount CR would likely be a zombie that never gets linked with the mariadbdatabase.  So I think doing this in openstack-operator before downstream is all built against 0.3.1pseudo would not work well.

If the operator is built on the 0.3.1 pseudo version as is the case for nova-operator above, then it becomes slightly more feasible that openstack-operator could create the MariaDBAccount up front, matching it with the one that mariadb-operator creates.  However, this requires code changes to the operator anyway, at the very least the operators need changes to their test suites to accommodate the SSL changes.     Since we have to change the operators anyway, I'm hoping that the changes here would be a minimal addition to those changes to simply ensure the operator consumes username/secret from the MariaDBAccount rather than hardwired from osp-secret.       There's already ppl waiting to see what this looks like so I can just have people help out getting all the operators to be on mariadb-operator 0.3.1pseudo.

So I think i know how we should do this.   The idea was that this "interimcreateaccount" thing would be just one line of code that can be put into a given operator making the transition, that allows it to be independent of other changes flying around.   However, the next several lines are then retrieving the MariaDBAccount and dbsecret anyway.   So we can roll this into a single helper renaming InterimCreateNewAccount to EnsureMariaDBAccount   , and then have it return the MariaDBAccount and Secret in one step.  This can at first actually generate the account / secret / pw if not present, and later on can be changed to expect that MariaDBAccount/secret was already created; if returning nil / error, the controller would add a failure message to its Conditions, since a controller should not be proceeding without having these two things set up.

I think that's the best way to go here, that is, an fn is needed here in any case so I can just make this look the way it will be in final form:

```diff
@@ -904,25 +903,10 @@ func (r *GlanceReconciler) generateServiceConfig(
        // initializes, and this block would no longer be present here.
        // At that point the DatabaseAccount field can be set to be required.
 
-       r.Log.Info(fmt.Sprintf("interim create new account start for account %s", instance.Spec.DatabaseAccount))
-
-       err := mariadbv1.InterimCreateNewAccount(
-               ctx, h, instance.Spec.DatabaseAccount, "glance", instance.Namespace,
+       databaseAccount, dbSecret, err := mariadbv1.EnsureMariaDBAccount(
+               ctx, h, instance.Spec.DatabaseAccount, glance.ServiceName, instance.Namespace,
        )
 
-       r.Log.Info("interim create new account end")
-
-       if err != nil {
-               return err
-       }
-       // end section that would be done by openstack-operator eventually
-
-       databaseAccount, err := mariadbv1.GetAccount(ctx, h, instance.Spec.DatabaseAccount, instance.Namespace)
-       if err != nil {
-               return err
-       }
-
-       dbSecret, _, err := secret.GetSecret(ctx, h, databaseAccount.Spec.Secret, instance.Namespace)
        if err != nil {
                return err
        }

``` 

that err above is passed outwards to `instance.Status.Conditions.Set(condition.FalseCondition` in `reconcileNormal`.

test/functional/glance_controller_test.go
zzzeek: so yeah I had it as a namespaced name originally but it seemed to be causing tests to fail, let me look more closely why that was happening

zzzeek: it does this:

```
     <*errors.StatusError | 0xc0007446e0>: 
      admission webhook "mglance.kb.io" denied the request: json: cannot unmarshal object into Go struct field GlanceSpec.spec.databaseAccount of type string
      {
          ErrStatus: {
              TypeMeta: {Kind: "Status", APIVersion: "v1"},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "admission webhook \"mglance.kb.io\" denied the request: json: cannot unmarshal object into Go struct field GlanceSpec.spec.databaseAccount of type string",
              Reason: "",
              Details: nil,
              Code: 400,
          },
      }

```

so i am trying to figure out where that is happening, what's mapping this to that string field

zzzeek: OK. looks like, "dont use that name" is the solution, keep it separate.  got it

zzzeek: oh nm, base_test is using it.    sorry for the chatter

test/functional/suite_test.go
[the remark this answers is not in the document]
zzzeek: ive created a helper in mariadb-operator and moved this to be local to the glanceapi_controller_test only where this doesn't otherwise happen.  for glance_controller_test the "ensuremariadbaccount" step is validated

controllers/glance_controller.go
zzzeek: we had a call with @stuggi and @dciabrin yesterday about TransportURL w/ rabbitMQ and how that might apply to galera.    What we sort of came up with is the answer is "sort of" - we dont really have a single, generic transport URL for galera, we have various bits that dont fit into a single URL like CAs, TLS options, but we do have a single client thing that's the local `.my.cnf` file, which we use in OOO as well for all the client side bits that are more elaborate than just host/username/password.    so we do have a plan that `.my.cnf` is going to be delivered by mariadb-operator, carried along into its destination by each operator like this glance-operator, however the specifics of the file will remain with mariadb-operator, so that changes on that end will propagate to all the controllers without them needing code changes.   we estimated that this would be similar in effect to having a transport URL set up ahead of time, just a little different.

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openstack-k8s-operators/glance-operator
-  newTag: latest
+  newName: quay.io/rhn_engineering_mbayer/glance-operator
+  newTag: 0.0.8
```
zzzeek, line 16: this keeps getting merged in here every time i build an image.   ill take this out when we're ready

go.mod
```diff
@@ -91,3 +91,5 @@ replace github.com/openstack-k8s-operators/glance-operator/api => ./api
 // mschuppert: map to latest commit from release-4.13 tag
 // must consistent within modules and service operators
 replace github.com/openshift/api => github.com/openshift/api v0.0.0-20230414143018-3367bc7e6ac7 //allow-merging
+
+replace github.com/openstack-k8s-operators/mariadb-operator/api => github.com/zzzeek/mariadb-operator/api v0.0.0-89bcc0f6f6941 //allow-merging
```
zzzeek, line 95: this will also go away once we merge the mariadb-operator side of it (which we'll know is done when we're happy with how this looks)

controllers/glance_controller.go
zzzeek: >  but in this flow I see a point where Glance should wait for an operation (the user and db creation) that should happen on the mariadb/galera side, that produces a secret that is consumed by Glance, while before we always have assumed osp-secret just exists and it's available. 

the osp-secret is created ahead of time but it's not useful in the Galera server until the CreateOrPatchDBByName method is called, which applies the password to a specific database that it also creates at that point.   so if the concern is "there's more waiting", there's not really any more waiting.   the MariaDBAccount will later be created up front by openstack-operator as well.

stuggi: > we had a call with @stuggi and @dciabrin yesterday about TransportURL w/ rabbitMQ and how that might apply to galera. What we sort of came up with is the answer is "sort of" - we dont really have a single, generic transport URL for galera, we have various bits that dont fit into a single URL like CAs, TLS options, but we do have a single client thing that's the local `.my.cnf` file, which we use in OOO as well for all the client side bits that are more elaborate than just host/username/password. so we do have a plan that `.my.cnf` is going to be delivered by mariadb-operator, carried along into its destination by each operator like this glance-operator, however the specifics of the file will remain with mariadb-operator, so that changes on that end will propagate to all the controllers without them needing code changes. we estimated that this would be similar in effect to having a transport URL set up ahead of time, just a little different.

right, my idea when we scheduled the call was to have the mariadbdatabase to be the owner of the mariadbaccount and then the mariadbdatabase could create a secret with the connection url and the my.cnf thats then be consumed by the operators. so basically give me DB with all required login details. for now I have submitted a PR to this operator https://github.com/openstack-k8s-operators/mariadb-operator/pull/190 to return the my.cnf

fmount: ok,
so thank you @zzzeek for the many information and to provide a green CI: that helps to see a e2e deployment and double check we don't have any corner case/regression.
I think I have a better idea of the whole flow now.
Also, from the current CI execution I see the `MariaDBAccount` CR [1] and the resulting secrets [2].

Even though I don't like much the secret name being just `glance.yaml` (it could be like `glance-dbaccount.yaml` but this can be changed later), I think that we're going to the right direction and at the service operators level right now we're consuming the secret generated by `mariadbdatabase`.
I think what covers my question about the "wait" part is the fact we have the required info created in advance, so we should be good on that front.

Quick question looking at the CI CR(s)[1] and the secrets [2]: @gibizer does nova already implements this mechanism?
I think we should have a doc patch in [3] to describe the whole flow and the interactions so that other people know the involved CRs and how the sequence diagram of the operators interaction work.

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_glance-operator/426/pull-ci-openstack-k8s-operators-glance-operator-main-glance-operator-build-deploy-tempest/1755748276902039552/artifacts/glance-operator-build-deploy-tempest/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-e39c97b0f556c0f58b33a035028cc0cd9a0e89f2650ef00160db8adfb85d876f/namespaces/openstack/crs/mariadbaccounts.mariadb.openstack.org/glance.yaml

[2] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_glance-operator/426/pull-ci-openstack-k8s-operators-glance-operator-main-glance-operator-build-deploy-tempest/1755748276902039552/artifacts/glance-operator-build-deploy-tempest/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-e39c97b0f556c0f58b33a035028cc0cd9a0e89f2650ef00160db8adfb85d876f/namespaces/openstack/secrets

[3] https://github.com/openstack-k8s-operators/docs/

config/manager/kustomization.yaml
fmount: +1, I think we should start consolidating the mariadb changes first, so we can rebase this patch (that in general looks good) with the new API calls provided by mariadb

controllers/glance_controller.go
```diff
@@ -886,7 +893,17 @@ func (r *GlanceReconciler) generateServiceConfig(
...
 	labels := labels.GetLabels(instance, labels.GetGroupLabel(glance.ServiceName), serviceLabels)
 
-	ospSecret, _, err := secret.GetSecret(ctx, h, instance.Spec.Secret, instance.Namespace)
+	// ensure a MariaDBAccount CR and Secret exist.
+	// this mariadb API function will as an interim step actually generate a
+	// new MariaDBAccount and Secret if one does not exist already.   in a
+	// future release, this function may change to emit an error if the
+	// MariaDBAccount was not already created ahead of time (e.g. by openstack-operator
+	// or end-user YAML declaration)
+	databaseAccount, dbSecret, err := mariadbv1.EnsureMariaDBAccount(
```
fmount, line 902: +1 this looks better and we don't have to revisit this code if something changes in mariadb. Thanks

zzzeek: note the RequireTLS flag (i.e. create the MariaDB username as requiring TLS to log in) hardcoded to false right now in this call.  that...has to come from somewhere at some point

fmount: I think at some point who creates the `MariaDBAccount` CR in advance (`openstack-operator`?) should pass that value to the CR: if we decide to expose that parameter (in a follow up work) we should have a struct for Database where we collect all the elements we need to set in the CR (`databaseAccount: string`, `requireTLS: bool`) and so forth. Having a high level view in the form of a doc patch would help moving forward w/ these kind of conversations in a different context.

zzzeek: I think `Database` would not have extra fields for this, because the requireTLS aspect is going to be transparent to the downstream controllers.  the `my.cnf` file that gets shuttled over will capture whats needed here for the login to proceed.

zzzeek: also AFAICT the `Database` facade seems to be used only for sending instructions to create databases, it does not seem to be used for consumption of parameters, at least here in glance though id have to look at nova and others to see what they do.   so at the moment i noted username/password fields would go away here, see my comment at https://github.com/openstack-k8s-operators/mariadb-operator/pull/184/files#r1484411670

controllers/glance_controller.go
gibizer: @fmount nova does not have this flow implemented yet. @zzzeek selected glance as being the first operator to get it.

zzzeek:
I can confirm this code successfully rotates out a new mariadbaccount if a new one is created with new secret, glance-operator picks it up and rebuilds the pods.

what did not work is the prunefinalizers step on the mariadb-galera side, the glance finalizer is still in the old mariadbaccount.  so I will try to fix that next week

controllers/glance_controller.go
zzzeek, line 902: note this no longer is given a mariadb username; as these usernames are expected to rotate regularly, EnsureMariaDBAccount also generates the username now, as it already generates the name of the Secret and the password.

controllers/glance_controller.go
```diff
@@ -655,6 +658,10 @@ func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glance
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+
+	// remove finalizers from unused MariaDBAccount records
+	mariadbv1.DeleteUnusedMariaDBAccountFinalizers(ctx, helper, glance.DatabaseName, instance.Spec.DatabaseAccount, instance.Namespace)
```
zzzeek, line 663: this is working, and mariadb-operator will not emit DROP USER until the Glance finalizer (and all other finalizers that are not MariaDBAccount itself) are removed

controllers/glanceapi_controller.go
```diff
@@ -809,6 +810,16 @@ func (r *GlanceAPIReconciler) generateServiceConfig(
 		return err
 	}
 
+	databaseAccount, err := mariadbv1.GetAccount(ctx, h, instance.Spec.DatabaseAccount, instance.Namespace)
```
zzzeek, line 813: probably should make a helper function to make this more succinct.

I've looked at how we get things like the database name and the database hostname for generating config, to see if we can consolidate, and I dont think so.  the operators all consider the database name to be a fixed constant in their own code, so that's where that comes from, no need to look up a MariaDBDatabase for that part.    The hostname comes from up front when they create the Database facade, but as we see here, Glance is taking that hostname and putting it into the *spec* of a glanceapi instance, so that's a totally custom thing.   otherwise it seems standard right now that as controllers use Database{} to generate MariaDBDatabase, they are grabbing the hostname right there and putting it into their `instance.Status`.     Then the `Database{}` facade isn't called upon at template construction time.

There's also talk about getting a my.cnf file shuttled through this process.   But for now I think keeping "getaccount/secret" as it's own thing as we have here (I'll add a facade to get both at once) is the way forward for the moment, with possible additions of my.cnf later on.

controllers/glanceapi_controller.go
```diff
@@ -809,6 +810,11 @@ func (r *GlanceAPIReconciler) generateServiceConfig(
 		return err
 	}
 
+	databaseAccount, dbSecret, err := mariadbv1.GetAccountAndSecret(ctx, h, instance.Spec.DatabaseAccount, instance.Namespace)
```
fmount, line 813: @zzzeek I'm wondering if a single `EnsureMariaDBAccount` interface is sufficient for both returning the existing information and create if it doesn't exist: if we can improve the implementation and handle the `getOrCreate` logic there will result in limiting the amount of exposed interfaces.
Note I'm not expecting this to be changed, and I'm fine with current approach, I was just wondering if a generic `EnsureMariaDBAccount` is sufficient to efficiently wrap and handle both the `get` and `creation` logic.

fmount: @zzzeek this comment applies more to PR https://github.com/openstack-k8s-operators/mariadb-operator/pull/184, sorry for writing it here.

controllers/glance_controller.go
zzzeek: i added this so that you can exercise a full create new mariadbaccount + update pods to use new username/password + remove finalizers from old one just by going in the CR and changing "DatabaseAccount" to any random new string

controllers/glanceapi_controller.go
zzzeek: do you mean that we'd use the `Database` interface for this operation also ?

fmount: I meant like re-using 

```golang

databaseAccount, dbSecret, err := mariadbv1.EnsureMariaDBAccount(
     ctx, h,
     instance.Spec.DatabaseAccount,
     instance.Namespace, false,
)
```
provided that the function  efficiently support the `get` logic; otherwise I think it's just fine using the current: 

```golang

databaseAccount, dbSecret, err := mariadbv1.GetAccountAndSecret(
    ctx, h,
    instance.Spec.DatabaseAccount,
    instance.Namespace)
```
I see that both have a similar signature and I was wondering if we can just using a single wrapper to get the same data.

zzzeek: oh...OK well they do really different things.  one of them actually generates a username and password, the other does not.     The fact that EnsureMariaDBAccount returns the two records was added later when I noticed in glance operator we happened to need the two objects right there anyway, but as I work within keystone and nova operators today, this is not the case; they do the "ensure" up top, throw away the output, then later on they have template-filling functions that use the GetAccountAndSecret function.

I think it's important that the role of "we are creating a new account / secret here because openstack-operators doesn't do this yet" is kept separate from the role of "we need to populate a config template and we need to know what the username/password is".   We can of course put these two behaviors into one single function, and then use a boolean flag `createIfNotExist` to differentiate the two behaviors, I would be OK with that, though personally I like keeping the two names / functions *really* separate.

zzzeek: The idea of EnsureMariaDBAccount is that later it will *not* generate anything, openstack-operator will be tasked with doing the generation, and we will not want to change any downstream code.  So the behavior of EnsureMariaDBAccount is expected to transparently change behind the scenes assuming we go forward with moving MariaDBAccount generation up to openstack-operator.

fmount: ack thanks for explaining the reason behind that choice.

controllers/glance_controller.go
```diff
@@ -266,7 +265,7 @@ func (r *GlanceReconciler) reconcileDelete(ctx context.Context, instance *glance
 	r.Log.Info(fmt.Sprintf("Reconciling Service '%s' delete", instance.Name))
 
 	// remove db finalizer first
-	db, err := mariadbv1.GetDatabaseByName(ctx, helper, instance.Name)
+	db, err := mariadbv1.GetDatabaseByNameAndAccount(ctx, helper, instance.Name, instance.Spec.DatabaseAccount, instance.Namespace)
```
fmount, line 268: If the mechanism provided by this patch will be the default, should this function simple be `GetDatabase(...` ?

go.mod
```diff
@@ -91,3 +91,5 @@ replace github.com/openstack-k8s-operators/glance-operator/api => ./api
 // mschuppert: map to latest commit from release-4.13 tag
 // must consistent within modules and service operators
 replace github.com/openshift/api => github.com/openshift/api v0.0.0-20230414143018-3367bc7e6ac7 //allow-merging
+
+replace github.com/openstack-k8s-operators/mariadb-operator/api => github.com/zzzeek/mariadb-operator/api v0.3.1-0.20240213153025-83e6db46eeb8 //allow-merging
```
fmount, line 95: ok I suspect this patch is basically ready to go, we can rebase/update and do another pass once we get the mariadb part merged.

controllers/glance_controller.go
zzzeek: not really sure, I find the Database facade kind of awkward, because it seems like it wants to act like an object you'd "get", but it's more like a package of things that have to be loaded together.     I guess that's why an awkward function name seems fitting at the moment :)

controllers/glance_controller.go
```diff
@@ -890,7 +898,17 @@ func (r *GlanceReconciler) generateServiceConfig(
...
+	// future release, this function may change to emit an error if the
+	// MariaDBAccount was not already created ahead of time (e.g. by openstack-operator
+	// or end-user YAML declaration)
+	databaseAccount, dbSecret, err := mariadbv1.EnsureMariaDBAccount(
+		ctx, h, instance.Spec.DatabaseAccount,
+		instance.Namespace, false,
+	)
+
 	if err != nil {
 		return err
```
zzzeek, line 913: can add status here

zzzeek:
this is all set assuming tests fully pass

fmount (approved):

PR #427, opened 2024-01-30
[docs] Example to configure resources for web-dwonload

konan-abhi:
Provided sample example which will show how to configure resources for web-download import method with file as a glance backend.


fmount (approved):
/lgtm

PR #429, opened 2024-01-31
[docs] Example to enable copy-image import method

konan-abhi:
Added sample example to enable copy-image import method in case multiple stores are configured in environment.


fmount (approved):
/lgtm

PR #430, opened 2024-02-01
Do not maintain a copy of the base ctlplane CR

fmount:
This patch removes the copy of the basic control plane we used to maintain, and it's now added as a reference to the one defined in the `openstack-operator` repository.
The base `kustomization.yaml` is adapted to the new CR to make sure we can add a per-backend overlay on top without extra issues. 
By doing this we can always point to the last available `ctlplane` `CR`.


abays (approved):
/lgtm

PR #432, opened 2024-02-02
Allow to define per api ImageCache size

fmount:
We are able to deploy an arbitrary number of `glanceAPI`. Each api/replicas, however, should have its own `image-cache` pvc as they are separate entities.
This patch improves the API to allow to either inherit the `imageCacheSize` or define it on a per instance basis.

### DIFF
--- a/api/bases/glance.openstack.org_glances.yaml
+++ b/api/bases/glance.openstack.org_glances.yaml
@@ -832,6 +832,9 @@ spec:
                       items:
                         type: string
                       type: array
+                    imageCacheSize:
+                      default: ""
+                      type: string
                     networkAttachments:
                       items:
                         type: string
@@ -961,6 +964,7 @@ spec:
                       type: string
                   required:
                   - containerImage
+                  - imageCacheSize
                   - storageRequest
                   type: object
                 type: object
--- a/api/v1beta1/common_types.go
+++ b/api/v1beta1/common_types.go
@@ -107,6 +107,10 @@ type GlanceAPITemplate struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// TLS - Parameters related to the TLS
 	TLS tls.API `json:"tls,omitempty"`
+
+	// ImageCacheSize, provides the size of the cache that will be reflected in the image_cache_max_size parameter
+	// +kubebuilder:default=""
+	ImageCacheSize string `json:"imageCacheSize"`
 }
 
 // APIOverrideSpec to override the generated manifest of several child resources.
--- a/api/v1beta1/glanceapi_types.go
+++ b/api/v1beta1/glanceapi_types.go
@@ -77,10 +77,6 @@ type GlanceAPISpec struct {
 	// QuotaEnforce if true, per-tenant quotas are enforced according to the
 	// registered keystone limits
 	Quota bool `json:"quota"`
-
-	// ImageCacheSize, provides the size of the cache that will be reflected in the image_cache_max_size parameter
-	// +kubebuilder:default=""
-	ImageCacheSize string `json:"imageCacheSize"`
 }
 
 // GlanceAPIStatus defines the observed state of GlanceAPI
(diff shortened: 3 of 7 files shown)


fmount:
@konan-abhi this should be more aligned for what we need in a regular or DCN deployment.
Suppose I deploy 3 API (default, default2, default3):

```
glance-default-single-0     3/3     Running     0          9m16s
glance-default2-single-0    3/3     Running     0          8m12s
glance-default3-single-0    3/3     Running     0          7m29s
```

I can select which one should have the `image-cache` pvc. In my test I added it on both default and default2, but not on default3, resulting in the following layout:

```
glance-cache-glance-default-single-0
glance-cache-glance-default2-single-0
glance-glance-default-single-0
glance-glance-default2-single-0
glance-glance-default3-single-0
``` 

```
    default:
      imageCacheSize: 2G
      type: single
    default2:
      imageCacheSize: 2G
      type: single
    default3:
      type: single
```

fultonj (commented):
Makes sense to me  +1

konan-abhi:
> @konan-abhi this should be more aligned for what we need in a regular or DCN deployment. Suppose I deploy 3 API (default, default2, default3):
> 
> ```
> glance-default-single-0     3/3     Running     0          9m16s
> glance-default2-single-0    3/3     Running     0          8m12s
> glance-default3-single-0    3/3     Running     0          7m29s
> ```
> 
> I can select which one should have the `image-cache` pvc. In my test I added it on both default and default2, but not on default3, resulting in the following layout:
> 
> ```
> glance-cache-glance-default-single-0
> glance-cache-glance-default2-single-0
> glance-glance-default-single-0
> glance-glance-default2-single-0
> glance-glance-default3-single-0
> ```
> 
> ```
>     default:
>       imageCacheSize: 2G
>       type: single
>     default2:
>       imageCacheSize: 2G
>       type: single
>     default3:
>       type: single
> ```

This makes sense, AFAIK in edge deployment we don't need image to be cached at central node, so can we say cache is/should be disabled on central node?

If yes, then can we simply attach PVC is caching is enabled? 

 Otherwise if we don't want to go this way, then this mechanism completely looks good.

fmount:
> > @konan-abhi this should be more aligned for what we need in a regular or DCN deployment. Suppose I deploy 3 API (default, default2, default3):
> > ```
> > glance-default-single-0     3/3     Running     0          9m16s
> > glance-default2-single-0    3/3     Running     0          8m12s
> > glance-default3-single-0    3/3     Running     0          7m29s
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I can select which one should have the `image-cache` pvc. In my test I added it on both default and default2, but not on default3, resulting in the following layout:
> > ```
> > glance-cache-glance-default-single-0
> > glance-cache-glance-default2-single-0
> > glance-glance-default-single-0
> > glance-glance-default2-single-0
> > glance-glance-default3-single-0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> >     default:
> >       imageCacheSize: 2G
> >       type: single
> >     default2:
> >       imageCacheSize: 2G
> >       type: single
> >     default3:
> >       type: single
> > ```
> 
> This makes sense, AFAIK in edge deployment we don't need image to be cached at central node, so can we say cache is/should be disabled on central node?
> 
> If yes, then can we simply attach PVC is caching is enabled?
> 
> Otherwise if we don't want to go this way, then this mechanism completely looks good.

I think with the current approach we can selectively add `imageCache` for a particular `GlanceAPI`, regardless of the DCN scenario. What you described is totally possible here, but we rely on the fact that the `StatefulSet` automatically creates the `PVC` for us using `VolumeCaimTemplates` (even when you scale replicas up).
In terms of User experience, we can control this behavior using the `imageCacheSize` parameter: in the previous patches, where we only had a single `glanceAPI`, it was a global (top-level) parameter, while now we are able to selectively enable caching for a given instance, which meet the DCN requirements.
There is a follow up [patch](https://github.com/openstack-k8s-operators/glance-operator/pull/435) that also takes care about the associated cronJobs, but I think it's easier review smaller pieces of code that solve a particular problem.

konan-abhi (approved):
Looks good, thank you!!

PR #435, opened 2024-02-04
Fix image cache cronJobs

fmount:
This patch fixes the way `cronJobs` are created.
With a particular focus to the `image-cache` feature, the goal is to perform a `cronJob` creation passing the right `glance-image-cache` `PVC` generated by a given `StatefulSet`.
Each `GlanceAPI` instance (either a replica or a dedicated API) needs an associated `cronJob` when the `ImageCache` is defined. 
A dedicated `cronJob` is needed because the `glance-{cleaner,pruner}` tools act against a path (`/var/lib/glance/image-cache`) created by the `StatefulSet` as `VolumeClaimTemplate`.

**TODO**:
- [x] Rebase on top of https://github.com/openstack-k8s-operators/glance-operator/pull/436
- [x] Expose ImageCacheSchedule parameter(s) through the main API 
- [x] Fix kuttl tests
- [x] Cleanup `GlanceAPI` and remove the cronJob `debug` option


controllers/glanceapi_controller.go
```diff
@@ -979,3 +1014,58 @@ func (r *GlanceAPIReconciler) ensureDeletedEndpoints(
...
+	serviceLabels map[string]string,
+	serviceAnnotations map[string]string,
+	cjType glance.CronJobType,
+) (ctrl.Result, error) {
+
+	var err error
+	var ctrlResult ctrl.Result
+
+	command := glance.GlanceCacheCleaner
+	command = glance.GlanceCacheCleaner
```
abays, line 1031: Looks like some duplication here

controllers/glanceapi_controller.go
```diff
@@ -979,3 +1014,58 @@ func (r *GlanceAPIReconciler) ensureDeletedEndpoints(
...
+	schedule := glance.CacheCleanerDefaultSchedule
+	// TODO: Fix debug option depending on the API
+	debugArg := ""
+
+	if cjType == "pruner" {
```
abays, line 1036: Could use `glance.CachePruner` instead of hardcoded string?

pkg/glance/pvc.go
```diff
@@ -16,7 +16,10 @@ func GetPvc(api *glancev1.GlanceAPI, labels map[string]string, pvcType PvcType)
 	var err error
 	requestSize := api.Spec.StorageRequest
 	pvcName := ServiceName
+	pvcAnnotation := map[string]string{}
+
 	if pvcType == "cache" {
```
abays, line 21: I know this wasn't changed in this PR, but perhaps could use `PvcCache` constant here instead?

pkg/glanceapi/cachejob.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// add CA cert if defined from the first api
+	if instance.Spec.GlanceAPITemplate.TLS.CaBundleSecretName != "" {
+		cronJobVolume = append(cronJobVolume, instance.Spec.GlanceAPITemplate.TLS.CreateVolume())
+		cronJobVolumeMounts = append(cronJobVolumeMounts, instance.Spec.GlanceAPITemplate.TLS.CreateVolumeMounts(nil)...)
+	}
+
+	// The image-cache PVC should be available to the Cache CronJobs to properly
+	// clean the image-cache path
+	if cronSpec.CjType == "pruner" || cronSpec.CjType == "cleaner" {
```
abays, line 83: Same question about perhaps using constants here

controllers/glanceapi_controller.go
```diff
@@ -759,6 +761,39 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
...
+			if err != nil {
+				instance.Status.Conditions.Set(condition.FalseCondition(
+					condition.CronJobReadyCondition,
+					condition.ErrorReason,
+					condition.SeverityWarning,
+					condition.CronJobReadyErrorMessage,
+					err.Error()))
+				return ctrlResult, err
+			}
+			instance.Status.Conditions.MarkTrue(condition.CronJobReadyCondition, condition.CronJobReadyMessage)
```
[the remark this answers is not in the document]
fmount: makes sense, this is not required, good catch.
Updating in my next PS

controllers/glanceapi_controller.go
fmount: yeah, we have `const` for this purpose, updating that part!

pkg/glanceapi/cachejob.go
fmount: Done

pkg/glance/pvc.go
fmount: Makes sense, and done in the last PS

controllers/glanceapi_controller.go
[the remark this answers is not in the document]

api/v1beta1/common_types.go
```diff
@@ -43,6 +43,10 @@ const (
 	DBPurgeDefaultAge = 30
 	//DBPurgeDefaultSchedule is in crontab format, and the default runs the job once every day
 	DBPurgeDefaultSchedule = "1 0 * * *"
+	//CleanerDefaultSchedule is in crontab format, and the default runs the job once every day
```
konan-abhi, line 46: I think it should be runs after every 30 minutes?

controllers/glanceapi_controller.go
```diff
@@ -777,6 +771,47 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
...
+					condition.CronJobReadyCondition,
+					condition.ErrorReason,
+					condition.SeverityWarning,
+					condition.CronJobReadyErrorMessage,
+					err.Error()))
+				return ctrlResult, err
+			}
+		}
+		// Cleanup any existing (but not used) ImageCache cronJob
+		if ctrlResult, err := r.deleteImageCacheJob(
```
konan-abhi, line 801: If I am not wrong, it will only cleanup CachePruner cronJob.

fmount: It will clean both [1]

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/435/files#diff-fe88ad2f79de701bd18c4a83352465de46564ebf02e97c1f6e6c1e0bd3f66554R1136

api/v1beta1/common_types.go
fmount: ack thanks, a copy/paste leftover, I'll fix it in my next PS

fmount: Done

controllers/glanceapi_controller.go
konan-abhi: I was under impression that ctrlResult is from line 781 loop, but its new one, so makes sense now.

konan-abhi (approved):
Thank you!!

controllers/glanceapi_controller.go
```diff
@@ -1000,3 +1034,122 @@ func (r *GlanceAPIReconciler) ensureDeletedEndpoints(
...
+	for _, vc := range cachePVCs.Items {
+		cacheAnnotations := vc.GetAnnotations()
+		if _, ok := cacheAnnotations["image-cache"]; ok {
+			var pvcName string = vc.GetName()
+			// Get the pod (by name) associated to the current pvc
+			var pod corev1.Pod
+			if err := r.Client.Get(ctx, types.NamespacedName{
+				Name:      strings.TrimPrefix(pvcName, "glance-cache-"),
+				Namespace: instance.Namespace,
+			}, &pod); err != nil && k8s_errors.IsNotFound(err) {
```
abays, line 1113: If the `err` is not `nil` and isn't a `NotFound`, could we mistakenly miss catching and returning it here?  i.e. the loop would just iterate to the next `vc` and the old `err` would be lost?  Maybe it's just a pedantic concern and doesn't really matter.

fmount: yeah, basically here we try to get a `Pod` that matches w/ the `Pvc` name (excluding the prefix introduced by the `StatefulSet` templating), and if the pod is not found, we need to perform a delete on the `cronJobs` that are still there.
That means that `err` doesn't matter here, as we want to just move to the next element and be silent in case a `Pod` or a `Cronjob` is not found (that's the only condition where we need the current value of `err`)

abays (approved):
/lgtm

konan-abhi (approved):
Thank you!

PR #436, opened 2024-02-04
Expose DBPurge Age and Schedule

fmount:
This patch updates the current glance CR to expose both `dbPurgeAge` and `dbPurgeSchedule` as customizable parameters. They were previously hardcoded as constants in the code, and it makes sense removing that part as long as the human operator has the ability to customize them. This also aligns the `glance-operator` with the work already done in both `Manila` and `Cinder`.


fmount:
@ASBishop FYI: I have a WIP patch that will be rebased on top of this to simplify the cronJob handling in the operator for image-cache, but I wanted to share this change w/ you as it's supposed to align Glance and the API/interface with what we've done in cinder/manila.

api/v1beta1/glance_types.go
```diff
@@ -139,12 +142,25 @@ type PasswordSelector struct {
 	Service string `json:"service"`
 }
 
+// DBPurge struct is used to model the parameters exposed to the Manila API CronJob
```
abays, line 145: `s/Manila/Glance/`?

abays (approved):
/lgtm

PR #439, opened 2024-02-07
Fix Dummy template formatting

fmount:
When the `webhook` injects the default/dummy configuration, we get:

`oslo_config.cfg.ConfigFileParseError: Failed to parse`

This patch fixes that error providing proper formatting to the dummy string.


konan-abhi (approved):
I think we should add a test for it?

fmount:
> I think we should add a test for it?
Done

konan-abhi (approved):
Looks perfect, thank you for the test!

PR #440, opened 2024-02-07
Remove enabled_backend entry from 00-config.conf

fmount:
With `webhooks` we prevent the `Glance` deploy unless a backend is specified via `customServiceConfig`.
It doesn't fail if `replicas` is set to `0`, but in general a backend must be specified.
This patch reinforce what already done by removing the `enabled_backends` line from `00-config`.


konan-abhi (approved):
Looks good!

templates/common/config/00-config.conf
```diff
@@ -18,7 +18,6 @@ max_logfile_size_mb=50
 log_rotation_type=size
 log_file = {{ .LogFile }}
 {{ end }}
-enabled_backends=default_backend:file
```
fmount, line 21: @konan-abhi wondering if we should also remove the following:

```
[glance_store]
```
on L46-L47

konan-abhi: If we are injecting it at runtime then yes, we can remove it as well.

fmount: ack thank you, updating the patch

fmount: Done

konan-abhi (approved):
Thank you!

abays (approved):
/lgtm

PR #445, opened 2024-02-14
Add GlanceSpecCore struct

dprince:
This version of the struct (called "core") is meant to be used via the OpenStackControlplane. It is the same as GlanceSpec only it is missing the containerImages.

Jira: [OSPRH-4835](https://issues.redhat.com//browse/OSPRH-4835)


fmount:
As per the previous conversation I think this patch is good to go, hence +2 from me. I see that the purpose is to not expose in the `openstack-operator` CR the `containerImage` parameter, but build it according to the `OpenStackVersion` CR.
I'm going to approve the patch, and once you're confident to land it feel free to remove the /hold.
(/cc @abays)

fmount (approved):
/lgtm

PR #447, opened 2024-02-16
Bump golangci-lint

gibizer:
Bump the golangci-lint version to the newest that still supports 1.20 and removed disabled checks that passing now.


fmount (approved):
/lgtm

PR #455, opened 2024-02-19
Document distributed image import feature

fmount:
This patch introduces the documentation associated to the distributed image import feature. 
It also provides a script and the steps required to automate tests.
Finally, within the same patch, a few nits are addressed in the way both the `endpointName` and the `StatefulSet` names are built.

Implements: [OSPRH-1243](https://issues.redhat.com/browse/OSPRH-1243)


konan-abhi (approved):
Perfect!!

PR #456, opened 2024-02-20
Sample: Fix cinder backend samples

Akrog:
This patch updates the glance samples that use the cinder backend, as they are no longer valid, because the `debug` section has been removed.

The patch also makes changes to glance's `type` and `preservejobs` in those samples so they match the new general samples.

### DIFF
--- a/config/samples/backends/cinder/glance-common/glance.yaml
+++ b/config/samples/backends/cinder/glance-common/glance.yaml
@@ -25,10 +25,7 @@ spec:
       databaseInstance: openstack
       glanceAPIs:
         default:
-          debug:
-            service: false
-          preserveJobs: false
           replicas: 1
-          type: split
+          type: single
       storageClass: ""
       storageRequest: 1G


config/samples/backends/cinder/glance-common/glance.yaml
```diff
@@ -25,10 +25,7 @@ spec:
       databaseInstance: openstack
       glanceAPIs:
         default:
-          debug:
-            service: false
-          preserveJobs: false
           replicas: 1
-          type: split
+          type: single
```
fmount, line 29: I think `split` is required for `cinder` in general, right? 
I'm using `single` only for `File` || `NFS` backend, so this field should be reverted I guess

konan-abhi: ++ split is required

fmount: This is covered by: https://github.com/openstack-k8s-operators/glance-operator/pull/459 where we updated and fixed (for the very same reason) most of the samples.

Akrog: What does it mean `split` is required?
As far as I know it is not required, it's an option, right?
It is an option to solve the internal vs public API issue.

Akrog:
This has been obsoleted by the later PR#459

PR #457, opened 2024-02-21
[tlse] TLS database connection

stuggi:
The my.cnf file gets added to the secret holding the service configs. The content of my.cnf is centrally managed in the mariadb-operator and retrieved calling db.GetDatabaseClientConfig(tlsCfg)

Depends-On: https://github.com/openstack-k8s-operators/mariadb-operator/pull/190
Depends-On: https://github.com/openstack-k8s-operators/mariadb-operator/pull/191
Depends-On: https://github.com/openstack-k8s-operators/mariadb-operator/pull/200

Jira: [OSPRH-4547](https://issues.redhat.com//browse/OSPRH-4547)


controllers/glance_controller.go
```diff
@@ -588,12 +532,19 @@ func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glance
 	instance.Status.Conditions.MarkTrue(condition.InputReadyCondition, condition.InputReadyMessage)
 	// run check OpenStack secret - end
 
+	db, result, err := r.ensureDB(ctx, helper, instance)
```
fmount, line 535: @stuggi not sure if there's a reason to not overwrite the same `ctrlResult` variable here and use `result`

controllers/glance_controller.go
```diff
@@ -1074,3 +1036,70 @@ func (r *GlanceReconciler) glanceAPICleanup(ctx context.Context, instance *glanc
 	}
 	return nil
 }
+
+func (r *GlanceReconciler) ensureDB(
```
fmount, line 1040: +1, good to have this logic into a `ensureDB` function

controllers/glance_controller.go
```diff
@@ -919,7 +871,17 @@ func (r *GlanceReconciler) generateServiceConfig(
...
+	var tlsCfg *tls.Service
+	for _, api := range instance.Spec.GlanceAPIs {
+		if api.TLS.CaBundleSecretName != "" {
+			tlsCfg = &tls.Service{}
+			break
+		}
+	}
+	customData := map[string]string{
+		glance.CustomConfigFileName: instance.Spec.CustomServiceConfig,
+		"my.cnf":                    db.GetDatabaseClientConfig(tlsCfg), //(mschuppert) for now just get the default my.cnf
```
fmount, line 883: ack, so the whole thing here is that we can get `my.cnf` as part of an API provided by the mariadb db object. Then we can use this information to set the database connection string.

fmount (commented):
Looks good to me.
We'll merge as long as it's rebased on main

controllers/glance_controller.go
stuggi: no, not really a reason.

stuggi:
need to update for the new added edge tests

controllers/glance_controller.go
stuggi: changed to the ctrlResult

PR #458, opened 2024-02-22
Fix cinder backend sample

ASBishop:
cinder_store_auth_address must be set or else glance will ignore the other cinder_store_XXX credentials (user_name, password, project_name).


Akrog (approved):
/lgtm

fmount (approved):

PR #459, opened 2024-02-23
Fix the samples and remove the debug option

fmount:
As per the recent changes we removed the debug option from the GlanceAPI, switching to the kubernetes approach (oc debug/<component>). This patch aligns the existing samples removing the unused option.


abays (approved):
/lgtm

PR #462, opened 2024-02-25
Enable caching using memcached

fultonj:
Jira: [OSPRH-1524](https://issues.redhat.com//browse/OSPRH-1524)


templates/common/config/00-config.conf
```diff
@@ -60,6 +60,9 @@ username={{ .ServiceUser }}
 {{ if (index . "ServicePassword") }}
 password = {{ .ServicePassword }}
 {{ end }}
+{{ if (index . "MemcachedServersWithInet") }}
+memcached_servers = {{ .MemcachedServersWithInet }}
+{{ end }}
```
fultonj, line 65: Without the `if (index . "MemcachedServersWithInet")` I get:
```
2024-02-25T17:09:29-05:00	ERROR	Reconciler error	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "GlanceAPI": {"name":"glance-default-single","namespace":"openstack"}, "namespace": "openstack", "name": "glance-default-single", "reconcileID": "60084dbf-994d-4bd6-8471-139180efea4f", "error": "template: tmp:63:23: executing \"tmp\" at <.MemcachedServersWithInet>: map has no entry for key \"MemcachedServersWithInet\""}
```
With this it works for me in that I see the property getting set in my configuration:
```
[fultonj@hamfast glance-operator{memcached}]$ oc get secret glance-config-data -o json | jq -r '.data."00-config.conf"' | base64 -d | grep memcached
memcached_servers = inet:[memcached-0.memcached]:11211
[fultonj@hamfast glance-operator{memcached}]$ 
```

fmount:
We might need to improve functional tests to make sure this new dependency is satisfied.
This means we should first get `memcached` into the environment (`suite_test.go`):

```golang
err = memcachedv1.AddToScheme(scheme.Scheme)
Expect(err).NotTo(HaveOccurred())
```

Then, in `glance_test_data.go` we should define the resources used by the *controller related
tests:

```golang

const (
	// MemcachedInstance - name of the memcached instance
	MemcachedInstance = "memcached"
)

type GlanceTestData struct {
...
...
MemcachedInstance      string
GlanceMemcached        types.NamespacedName
```
and later in the same file we can have an instance of the new variables:

```golang
GlanceMemcached: types.NamespacedName{
			Namespace: glanceName.Namespace,
			Name:      MemcachedInstance,
		},
...
...
MemcachedInstance:   MemcachedInstance,
```
At this point in `glance_controller_test.go` we can add this instance as a prerequisite for the whole test to run.
Basically, w/o memcached glance is not going to work anymore as a standalone component.
I think we can initialize a common Spec for all the test blocks:

```golang
var _ = Describe("Glance controller", func() {
	var memcachedSpec memcachedv1.MemcachedSpec

	BeforeEach(func() {
		memcachedSpec = memcachedv1.MemcachedSpec{
			Replicas: ptr.To(int32(3)),
		}
	})
...
...
```
and finally, for each `BeforeEach` block, we can have:

```golang
DeferCleanup(infra.DeleteMemcached, infra.CreateMemcached(namespace, glanceTest.MemcachedInstance, memcachedSpec))
infra.SimulateMemcachedReady(glanceTest.ManilaMemcached)
```

Also, the same applies to `glanceapi_controller.go`.
The above can be slightly different, but it's pretty much what we need to make the functional tests pass (see [1] to have an example).

[1] https://github.com/openstack-k8s-operators/manila-operator/pull/164

templates/common/config/00-config.conf
fmount: interesting, I have to check this one because it shouldn't happen

fmount: ok I think I understand now. It makes sense to have the if condition over there because we rely on the same template (`00-config.conf`) that can be used both in `glance_controller` (by cronjobs, dbsync and other resources), and in `glanceapi_controller` (which is the one that is failing for you). In particular, each `glanceapi` controller build an extended version of that config file, hence we set multiple templates options that are missing there.
I'm wondering if this option is available in `glance-<apiName>-config-data`, otherwise we need to propagate it to the sublevel CR.
Right now I think we're not using memcached at all because the config option is not propagated to the underlying subCR.

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/controllers/glanceapi_controller.go#L843

fmount (changes requested):
I think we need to either find a way to propagate the memcached config to the underlying subCR (via API?) or move the existing controller logic to the glanceapi controller.
The current approach would work only if we pass `strings.Join(memcached.Status.ServerListWithInet, ",")` to the subCR and set it in the `generateConfig` in [1]

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/controllers/glanceapi_controller.go#L843

fultonj:
> I think we need to either find a way to propagate the memcached config to the underlying subCR (via API?) or move the existing controller logic to the glanceapi controller.

Thanks for your help on this. The latest update does this. I confirm now that ` glance-config-data` does not have the `memcached_servers` while `glance-default-single-config-data` (for my testing) does. 

I still need to add memcached to the functional test which I'll do in my next update.

fultonj:
> We might need to improve functional tests to make sure this new dependency is satisfied. 

Thanks for all of your guidance in this. I've added the tests but they're not yet passing. I'll keep debugging tomorrow.

  https://github.com/openstack-k8s-operators/glance-operator/compare/f7c145679abbf9ae6424f789083279325d44ea2c..0fb6a4649c1574edc738327bd331be8486664b58

controllers/glanceapi_controller.go
```diff
@@ -896,6 +896,8 @@ func (r *GlanceAPIReconciler) generateServiceConfig(
 		templateParameters["CacheMaxSize"] = cacheSize.Value()
 		templateParameters["ImageCacheDir"] = glance.ImageCacheDir
 	}
+	memcached, err := getGlanceMemcached(ctx, h, instance.Spec.MemcachedInstance, instance.Namespace)
```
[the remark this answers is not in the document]
fmount: and the above should unblock your testing w/ envTests in the first place. Then I think you can investigate specific failures

templates/common/config/00-config.conf
fultonj: I think this issue is resolved. The `memcached_servers` are now where they should be. Thanks for your help.

```
[fultonj@hamfast functional{memcached}]$ oc get secret glance-config-data -o json | jq -r '.data."00-config.conf"' | base64 -d | grep memcached
[fultonj@hamfast functional{memcached}]$ oc get secret glance-default-single-config-data -o json | jq -r '.data."00-config.conf"' | base64 -d | grep memcache
memcached_servers = inet:[memcached-0.memcached]:11211
[fultonj@hamfast functional{memcached}]$ 
```

fultonj:
All my functional test errors have a timeout waiting for Informers to sync [1]. I'm going to focus on the following error though since memcached is the only change and the same tests pass if I switch to the main branch in my env.

```
"failed to get API group resources: unable to retrieve the complete list of server APIs: memcached.openstack.org/v1beta1: the server could not find the requested resource"
```

[1] 
```
[fultonj@hamfast glance-operator{memcached}]$ make test > log
[fultonj@hamfast glance-operator{memcached}]$ grep ERROR log | awk -F 'T| +' '{print substr($0, index($0, $5))}' | sort  | uniq
after stop sequence was engaged	{"error": "failed to wait for glance caches to sync: cache did not sync"}
get informer from cache	{"error": "failed to get API group resources: unable to retrieve the complete list of server APIs: memcached.openstack.org/v1beta1: the server could not find the requested resource"}
get informer from cache	{"error": "Timeout: failed waiting for *v1beta1.KeystoneEndpoint Informer to sync"}
get informer from cache	{"error": "Timeout: failed waiting for *v1beta1.KeystoneService Informer to sync"}
get informer from cache	{"error": "Timeout: failed waiting for *v1beta1.MariaDBAccount Informer to sync"}
get informer from cache	{"error": "Timeout: failed waiting for *v1beta1.MariaDBDatabase Informer to sync"}
get informer from cache	{"error": "Timeout: failed waiting for *v1.CronJob Informer to sync"}
get informer from cache	{"error": "Timeout: failed waiting for *v1.Job Informer to sync"}
get informer from cache	{"error": "Timeout: failed waiting for *v1.NetworkAttachmentDefinition Informer to sync"}
get informer from cache	{"error": "Timeout: failed waiting for *v1.PersistentVolumeClaim Informer to sync"}
get informer from cache	{"error": "Timeout: failed waiting for *v1.RoleBinding Informer to sync"}
get informer from cache	{"error": "Timeout: failed waiting for *v1.Role Informer to sync"}
get informer from cache	{"error": "Timeout: failed waiting for *v1.Secret Informer to sync"}
get informer from cache	{"error": "Timeout: failed waiting for *v1.ServiceAccount Informer to sync"}
get informer from cache	{"error": "Timeout: failed waiting for *v1.Service Informer to sync"}
get informer from cache	{"error": "Timeout: failed waiting for *v1.StatefulSet Informer to sync"}
wait for Cache to sync	{"controller": "glance", "controllerGroup": "glance.openstack.org", "controllerKind": "Glance", "error": "failed to wait for glance caches to sync: cache did not sync"}
[fultonj@hamfast glance-operator{memcached}]$ 
```

fultonj:
> All my functional test errors have a timeout waiting for Informers to sync

This is now resolved.

fmount (approved):
/lgtm

PR #463, opened 2024-02-26
Deploy an arbitrary number of Edge GlanceAPI

fmount:
To be able to serve workloads behind different backends that live at the `Edge` site, the operator should be able to be aware of a `GlanceAPI` `Edge` instance type.
An `Edge` instance has the following traits:

1. It can't be promoted as a `Central API` and registered in `Keystone` (or switched to a `central` API during a CR update);
2. It has neither a public endpoint (public `Service`), nor a `Route` associated;
3. It can only be reached via `Internal` `Service/endpoint`;
4. It still provides an `headless Service` to take advantage of `distributed image import` (it especially enables HA at application level)

Implements: [OSPRH-2679](https://issues.redhat.com/browse/OSPRH-2679)


konan-abhi:
> 4\. It still provides an `headless Service` to take advantage of `distributed image import` (it especially enables HA at application level)

Since it can't be promoted as Central API we really don't required distributed image import supported at edge site (unless we are going to have more than one replicas which I doubt we will be having for edge site.)

konan-abhi (approved):
Changes looks good to me, we need to verify that copy-image import method is enabled in edge deployment at central site + edge site and image is copied from central site to edge site.

Let me know if you need details about how to copy image between two glance sites.

Thank you!!

fmount:
> copy-image import method is enabled in edge deployment at central s

We can definitely come up w/ a guide for the `copy-image` import method, but I suspect is something we document and ask to enable via `customServiceConfig`. We'll follow up on that part soon!

PR #469, opened 2024-03-02
Add troubleshooting guide for glance-operator

fmount:
As per the recent changes across the `openstack-k8s-operators`, all the `debug` options have been removed from the `CR(s)`. This guide is supposed to update the procedure required to perform some troubleshooting against the `Glance` Service (run it manually and interact with it).

### DIFF
--- a/docs/dev/design-decisions.md
+++ b/docs/dev/design-decisions.md
@@ -125,7 +125,7 @@ spec:
   serviceUser: glance
   databaseInstance: openstack
   databaseUser: glance
-  glanceAPI:
+  glanceAPIs:
     ..
   secret: osp-secret
   storageClass: ""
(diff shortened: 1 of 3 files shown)


docs/dev/design-decisions.md
```diff
@@ -161,6 +161,9 @@ The following diagram describes how k8s resources/object are connected.
 | glance CR  | |---> | glanceAPI-default CR|       +-------------------------+
 +------------+       +---------------------+ |---> | glance-default-external |
                                                    +-------------------------+
+ (headless SVC to allow pod2pod communication)     | glance-default-single   |
```
fmount, line 164: have to change this

docs/dev/troubleshoot_api_setup.sh
abays, line 1: Is the intention naybe that the user copies this script into the duplicate pod mentioned in `docs/dev/troubleshooting.md`?  Or is this just an example.  Just curious.

fmount: Yeah, that's something we should mention in the troubleshooting guide: it started as an example (a script that automates part of these steps for you generating the config and starting the service), but it might be a good idea have a dedicated step to copy it, so less things to do by a human operator and less error prone.
It's a good idea, I'll do that in my next iteration.

abays (approved):
/lgtm

PR #472, opened 2024-03-04
Use bind_host=localhost

rabi:
Using `localhost` would allow bind_host to use the correct loopback address for ipv6.


hjensas (approved):

fultonj:
+2 Makes sense to me. 

Any comments @fmount @konan-abhi ?

fmount:
Works for me, thanks @fultonj for pinging

fmount (approved):
/lgtm

PR #473, opened 2024-03-05
Use PodManagementPolicy: appsv1.ParallelPodManagement for statefulset

stuggi:
With the default PodManagementPolicy: OrderedReadyPodManagement the statefulset controller will only progress pods when the previous/current pod is ready or terminated.

When service configuration changes while the pod is starting and the new configuration requires e.g. additional volume mounts the initial pod will never reach ready and therefore an update won't happen.

With ParallelPodManagement the statefulset controller will not wait for pods to be ready or complete termination.

### DIFF
--- a/pkg/glanceapi/statefulset.go
+++ b/pkg/glanceapi/statefulset.go
@@ -190,7 +190,8 @@ func StatefulSet(
 			Selector: &metav1.LabelSelector{
 				MatchLabels: labels,
 			},
-			Replicas: instance.Spec.Replicas,
+			PodManagementPolicy: appsv1.ParallelPodManagement,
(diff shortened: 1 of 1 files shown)


fmount (approved):
/lgtm

pkg/glanceapi/statefulset.go
```diff
@@ -190,7 +190,8 @@ func StatefulSet(
 			Selector: &metav1.LabelSelector{
 				MatchLabels: labels,
 			},
-			Replicas: instance.Spec.Replicas,
+			PodManagementPolicy: appsv1.ParallelPodManagement,
```
fmount, line 193: totally makes sense, thanks for this patch!

PR #481, opened 2024-03-19
Re-init conditions for each reconcile

fmount:
This patch is part of the updates/upgrades related effort, and as proposed for other operators, it re-inits the conditions when a reconcile is performed. By doing this we can catch issues and set conditions to False if a requirement is not met.

Related: [OSPRH-5698](https://issues.redhat.com/browse/OSPRH-5698)

### DIFF
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -699,8 +699,14 @@ func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glance
 		return ctrlResult, err
 	}
 	instance.Status.Conditions.MarkTrue(condition.CronJobReadyCondition, condition.CronJobReadyMessage)
-
 	// create CronJob - end
+
+	// We reached the end of the Reconcile, update the Ready condition based on
+	// the sub conditions
+	if instance.Status.Conditions.AllSubConditionIsTrue() {
+		instance.Status.Conditions.MarkTrue(
+			condition.ReadyCondition, condition.ReadyMessage)
+	}
 	return ctrl.Result{}, nil
 }
 
@@ -873,6 +879,10 @@ func (r *GlanceReconciler) apiDeploymentCreateOrUpdate(
 	apiSpec.GlanceAPITemplate.StorageRequest = instance.Spec.StorageRequest
 	apiSpec.GlanceAPITemplate.StorageClass = instance.Spec.StorageClass
 	apiSpec.MemcachedInstance = memcached.Name
+	// Make sure to inject the ContainerImage passed by the OpenStackVersions
+	// resource to all the underlying instances and rollout a new StatefulSet
+	// if it has been changed
+	apiSpec.GlanceAPITemplate.ContainerImage = instance.Spec.ContainerImage
 
 	// We select which glanceAPI should register the keystoneEndpoint by using
 	// an API selector defined in the main glance CR; if it matches with the
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -123,6 +123,17 @@ func (r *GlanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 		return ctrl.Result{}, err
 	}
 
+	// initialize status if Conditions is nil, but do not reset if it already
+	// exists
+	isNewInstance := instance.Status.Conditions == nil
+	if isNewInstance {
+		instance.Status.Conditions = condition.Conditions{}
+	}
+
+	// Save a copy of the condtions so that we can restore the LastTransitionTime
+	// when a condition's state doesn't change.
+	savedConditions := instance.Status.Conditions.DeepCopy()
+
 	// Always patch the instance status when exiting this function so we can persist any changes.
 	defer func() {
 		// update the Ready condition based on the sub conditions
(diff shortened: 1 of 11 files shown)


fmount:
First checking a CI execution and see if we have gaps with the current tests.

controllers/glance_controller.go
```diff
@@ -152,31 +152,31 @@ func (r *GlanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
 
-		instance.Status.Conditions.Init(&cl)
+	instance.Status.Conditions.Init(&cl)
 
-		// Register overall status immediately to have an early feedback e.g. in the cli
-		return ctrl.Result{}, nil
-	}
+	// Register overall status immediately to have an early feedback e.g. in the cli
+	//return ctrl.Result{}, nil
+	//}
```
gibizer, line 179: * I guess you don't want to carry commented out code.
* also we need to be aware that this change means there will be a longer time period where the status.Conditions will be nil on the CR after create. This might cause visible failures of pre-existing bugs around condition handling.

fmount: +1 for the commented code, I just started testing the behavior of the operator when those conditions are re-initialized on each reconcile.
I think in general is good if we can catch bugs around wrong conditions handling at this stage. For `nil`, you mean we have a longer period w/ an `UnknownCondition` ?

gibizer: I mean we will have a longer period between CR creation and the first update on the CR that initialize the Status.Condition field from nil to a list of conditions.

controllers/glance_controller.go
```diff
@@ -156,16 +156,18 @@ func (r *GlanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 		instance.Status.Conditions = condition.Conditions{}
 	}
 
-	// initialize conditions used later as Status=Unknown
+	// initialize conditions used later as Status=Unknown, except the ReadyCondition
+	// that should be False when we start
 	cl := condition.CreateList(
+		// Mark ReadyCondition as False from the beginning
+		condition.FalseCondition(condition.ReadyCondition, condition.InitReason, condition.SeverityInfo, condition.ReadyInitMessage),
```
gibizer, line 163: as the reconcile loop always exit via the logic in the deferred call https://github.com/openstack-k8s-operators/glance-operator/blob/ca211863b3824313ae1794af92179b8a4a44214f/controllers/glance_controller.go#L129-L139 this should not be necessary as the deferred call will either set the ReadyCondition to True all mirror the status of the most sever condition in the list.

controllers/glanceapi_controller.go
```diff
@@ -642,6 +670,11 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
 				err.Error()))
 			return ctrlResult, err
 		} else if (ctrlResult != ctrl.Result{}) {
+			instance.Status.Conditions.MarkFalse(
```
gibizer, line 673: nit: I would consider this as Unknown instead of False based on the last bulletpoint of https://github.com/openstack-k8s-operators/docs/blob/main/conditions.md#conditions-in-openstack-k8s-operators-podified-controlplane-operators

controllers/glanceapi_controller.go
```diff
@@ -661,6 +694,11 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
 			err.Error()))
 		return ctrlResult, err
 	} else if (ctrlResult != ctrl.Result{}) {
+		instance.Status.Conditions.MarkFalse(
```
gibizer, line 697: ditto this feels something in progress

controllers/glanceapi_controller.go
```diff
@@ -683,6 +721,12 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
 	} else if hashChanged {
 		// Hash changed and instance status should be updated (which will be done by main defer func),
 		// so we need to return and reconcile again
+		r.Log.Info(fmt.Sprintf("%s... requeueing", condition.ServiceConfigReadyInitMessage))
+		instance.Status.Conditions.MarkFalse(
+			condition.ServiceConfigReadyCondition,
+			condition.InitReason,
+			condition.SeverityInfo,
+			condition.ServiceConfigReadyInitMessage)
 		return ctrl.Result{}, nil
```
gibizer, line 730: This is confusing for me. The log states requeueing but we does not return a requeue result but a success result. Also setting the condition to False and then indicating success is a contradiction to me.

controllers/glance_controller.go
fmount: right, I'm just not there yet because I'm not sure that re-asserting all the conditions is the right thing to do at the end of the reconcile loop. I'm going to update this patch soon, and the plan is to remove this evaluation from the defer call

controllers/glanceapi_controller.go
fmount: right, the wording is probably wrong. I'll remove the `log` message because is a leftover, but the main idea is to mark the condition as False because the next reconcile will update the deployment based on the input that have changed.
Does this makes sense? I think I can't set this condition to True if I still have to reconcile based on the new Input

controllers/glanceapi_controller.go
fmount: Based on this code, `instance.Spec.TLS.CaBundleSecretName` is not empty, `ValidateCACertSecret` fails, and I assume there's something that should be changed somewhere in the CR (or the openstack-operator), so I'm not sure that the glance-operator has the logic to make this progressing to a `Status=True` state without any "external" or "user" related intervention: am I interpreting this scenario wrong?

controllers/glanceapi_controller.go
fmount: same comment I had above

controllers/glance_controller.go
fmount: There is a side conversation on the meaning of the `isReady` here, that's why I did not update this code so far.

controllers/glanceapi_controller.go
gibizer: We discussed this over a call. I think the return statement can be removed and let the rest of the reconcile roll out the new hash

fmount: yeah, this is not required, I can confirm that progressing in the current loop, instead of returning, takes care about rolling new pods out.
I'm going to remove the whole `else if` as part of the next update.

controllers/glanceapi_controller.go
fmount: I had a comment on [1], but if it meets the definition of `Unknown` I'm ok to change it to `Unknown`. Also this is not an `error` situation, we're comparing `ctrlResult != ctrl.Result{}` but actually no error is returned, so I think it makes sense to move it to use `Unknown` instead of `False`. Comparing `ctrlResult != ctrl.Result{}` feels like something still in progress rather than an error that should mark the condition to `False` and return an `err`.
Thank you, this is a good catch! I'll change that in my next iteration.

[1] https://github.com/openstack-k8s-operators/docs/pull/99

controllers/glanceapi_controller.go
fmount: Comparing `ctrlResult != ctrl.Result{}` feels like something still `in-progress` rather than an `error` that should mark the `condition` to `False` and return an `err`.
I'll change this part as well in my next iteration to match the other similar comment I had in a different piece of code that does the same comparison.

dprince (approved):

PR #482, opened 2024-03-19
Do not set image_conversion config options for internal Pods

fmount:
This patch fixes a minor issue in the backends `README`, and also enables `image_conversion` for **external** pods only.
If a pod is single/edge/internal, it won't have an `image_conversion` `PVC`, hence the related config in `00-config` is not required.

Related: [OSPRH-5690](https://issues.redhat.com/browse/OSPRH-5690)

### DIFF
--- a/config/samples/backends/README.md
+++ b/config/samples/backends/README.md
@@ -58,8 +58,8 @@ If we already have a deployment working we can always use
 When Ceph is adopted as a backend, Glance `image-conversion` is enabled by default.
 It's realized through a dedicated `PVC` (built by the `StatefulSet` via templates)
 that is mounted to the `/var/lib/glance/os_glance_staging_store` path.
-A `glance-conversion` PVC can be found with if the Glance **external** Pod is
-inspected via the `oc describe pod .. ` command:
+A `glance-conversion` PVC can be found inspecting the Glance **external** Pod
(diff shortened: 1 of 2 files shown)


fmount:
@konan-abhi can you confirm that in the `internal` Pod (which is still configured with `Ceph`), we do not require the image_conversion config in 00-config?
We do not attach the `glance-conversion` PVC to an internal/edge/single Pod, hence I think the image_conversion options are not required in 00-conifg (we can leave `no_op`).
Are these assumptions correct?

konan-abhi:
> @konan-abhi can you confirm that in the `internal` Pod (which is still configured with `Ceph`), we do not require the image_conversion config in 00-config? We do not attach the `glance-conversion` PVC to an internal/edge/single Pod, hence I think the image_conversion options are not required in 00-conifg (we can leave `no_op`). Also, the internal Pod is not used to do any conversion that only happens in the external instance. Are these assumptions correct?

++ We do not need this enabled to internal pod. Actually we don't need any import plugins or import methods set for internal pods since other services do not use image import workflow at all.

On the other hand even if it remains in config file it is kind of noop for internal pod, it will not cause any regression.

konan-abhi (approved):
Looks good, thank you!

PR #483, opened 2024-03-20
Do not run db-sync as root

fmount:
This patch introduces a `glanceSecurityContext` function to make sure we do not run `db-sync` as root.
This is supposed to avoid issues during the bootstrap process, and a follow up patch will extend the `scc` to both the `-api` and `-httpd` containers.

### DIFF
--- a/pkg/glance/const.go
+++ b/pkg/glance/const.go
@@ -46,12 +46,14 @@ const (
 	// PvcImageConv is used to define a PVC mounted for image conversion purposes
 	// when Ceph is detected as a backend
 	PvcImageConv PvcType = "imageConv"
-
 	// GlancePublicPort -
 	GlancePublicPort int32 = 9292
 	// GlanceInternalPort -
 	GlanceInternalPort int32 = 9292
-
+	// GlanceUID - https://github.com/openstack/kolla/blob/master/kolla/common/users.py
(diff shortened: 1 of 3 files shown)


pkg/glance/const.go
```diff
@@ -90,6 +92,9 @@ const (
 	CachePruner CronJobType = "pruner"
 	//ImageCacheDir -
 	ImageCacheDir = "/var/lib/glance/image-cache"
+
+	// GlanceDBSyncCommand -
+	GlanceDBSyncCommand = "/usr/local/bin/kolla_start"
```
fmount, line 97: we don't need `kolla_set_config` because we already mount `00-config.conf` in the target directory (`/etc/glance/glance.conf.d/`).

pkg/glance/funcs.go
```diff
@@ -53,3 +54,22 @@ func GetGlanceAPIName(name string) string {
 	}
 	return api
 }
+
+// glanceSecurityContext - currently used to make sure we don't run db-sync as
+// root user
+func glanceSecurityContext() *corev1.SecurityContext {
+	trueVal := true
```
konan-abhi, line 61: nit: 
Why can't we define this in const.go?

konan-abhi (approved):
Looks good, thank you!

PR #491, opened 2024-03-27
Add Status.ObservedGeneration to the top-level Glance CR

fmount:
This field is used by minor updates to make sure we know, when a CR changes, if the requested generation matches with the last observed state, then we can check the `IsReady` because we're sure we're not looking at an outdated CR.

Jira: [OSPRH-5919](https://issues.redhat.com//browse/OSPRH-5919)


konan-abhi (approved):
Looks good!

PR #492, opened 2024-03-27
Improve logging and remove placeholder reconcile functions

fmount:
This patch improves logging where necessary and simplify some code by removing both `reconcileUpdate` and `reconcileUpgrade` placeholder functions (that result in less logging).


abays:
Seems reasonable to me

abays (approved):
/lgtm

PR #494, opened 2024-03-29
Improve envTest for validation webhooks

fmount:
This patch fixes the `isValidEndpoint` function provided by the validation `webhook` and introduces `envTests` for Validation webhooks that are one of the most important part of the operator.

### DIFF
--- a/api/v1beta1/glance_webhook.go
+++ b/api/v1beta1/glance_webhook.go
@@ -72,11 +72,9 @@ func (r *Glance) Default() {
 // Check if the KeystoneEndpoint matches with a deployed glanceAPI
 func (spec *GlanceSpec) isValidKeystoneEP() bool {
 	for name, api := range spec.GlanceAPIs {
-		// If it's an Edge API is not a valid endpoint
-		if api.Type == APIEdge {
-			return false
-		}
-		if spec.KeystoneEndpoint == name {
+		// A valid keystoneEndpoint can either be applied to
+		// a single API or split type, but not to an EdgeAPI
+		if api.Type != APIEdge && spec.KeystoneEndpoint == name {
 			return true
 		}
 	}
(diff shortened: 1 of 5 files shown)


api/v1beta1/glance_webhook.go
```diff
@@ -234,12 +232,13 @@ func (r *Glance) ValidateUpdate(old runtime.Object) (admission.Warnings, error)
...
-		// (e.g. an Edge GlanceAPI instance that can't be registered in keystone)
-		// return an error message
-		if !r.Spec.isValidKeystoneEP() {
-			return nil, errors.New("KeystoneEndpoint is assigned to an invalid glanceAPI instance")
-		}
+	}
+	// At update time, if the CR has an invalid keystoneEndpoint set
+	// (e.g. an Edge GlanceAPI instance that can't be registered in keystone)
+	// return an error message
+	if !r.Spec.isValidKeystoneEP() {
```
fmount, line 239: same comment, we already iterate over the API within the function, no need to stay within an external for loop (that represents the same iteration!)

test/functional/validation_webhook_test.go
```diff
@@ -0,0 +1,151 @@
...
+		)
+	})
+
+	It("webhooks reject the request - invalid instance", func() {
+		spec := GetGlanceDefaultSpec()
+
+		gapis := map[string]interface{}{
+			"edge2": map[string]interface{}{
+				"replicas": 1,
+				"type":     "edge",
```
konan-abhi, line 106: I have a question here, in case of edge do we provide split deployment or just internal endpoint?

fmount: Internal only, the edge layout doesn't allow the API split

konan-abhi (approved):
Looks good, thank you!

PR #496, opened 2024-03-29
Mirror ReadyCondition with the actual current condition

fmount:
This patch covers two things:

- `Mirror()` relies on sorting conditions in each condition group by their `LastTransitionTime`, so it swaps the `RestoreLastTransitionTimes()` call with the `Mirror()` call in the defer function

- It setup `ReadyCondition` as `Unknown` instead of `False`, so we can avoid it to be the only one mirrored in the
  top-level status: this allows the other conditions to compete for the top-level mirroring

### DIFF
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -137,11 +137,12 @@ func (r *GlanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 	// Always patch the instance status when exiting this function so we can
 	// persist any changes.
 	defer func() {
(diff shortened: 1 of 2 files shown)


abays (approved):
/lgtm

ASBishop (approved):
/lgtm

I saw a demo of this and it works great. Until reconciliation completes, the ReadyCondition tells you which sub-condition isn't ready.

ASBishop:
/retest

rabbitmq failure

PR #498, opened 2024-04-02
Fix TLSInputReadyCondition

fmount:
When GlanceAPI's reconciler is executed, it checks for the provided Input, including TLS related info. If we're not failing (err is not nil), but the operation is still in progress, we should use the proper InputReady message as Reason instead the InitMessage. This patch fixes it and fix the condition with the proper message.


abays:
```
error: Post "https://api.oko-12.oooci.ccitredhat.com:6443/apis/build.openshift.io/v1/namespaces/openstack/buildconfigs": dial tcp 38.102.83.63:6443: i/o timeout
```

/retest-required

PR #501, opened 2024-04-05
Add ObservedGeneration to the sub Resources

fmount:
This patch does two main things:

1. it adds `observedGeneration` to the sub custom resources
2. it proposes to bump the `observedGeneration` at the beginning of the reconciliation loop (see the [knative conversation](https://github.com/knative/serving/issues/4937))
3. it checks, at the top level, if the `ObservedGeneration` matches with the `metadata.generation` assigned to the `subCR(s)`

Depends-on: https://github.com/openstack-k8s-operators/lib-common/pull/488


controllers/glance_controller.go
```diff
@@ -669,8 +669,11 @@ func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glance
 
 	// We reached the end of the Reconcile, update the Ready condition based on
 	// the sub conditions
-	instance.Status.ObservedGeneration = instance.Generation
-	if instance.Status.Conditions.AllSubConditionIsTrue() {
+	subGen, err := r.checkGlanceAPIsGeneration(ctx, instance)
+	if err != nil {
+		return ctrlResult, err
```
fmount, line 674: I might want to enforce a `False` condition here: if I'm here it means I'm not able to get the subCRs and there's a problem in my deployment. Hopefully I already have mirrored the real issue in the top-level `ReadyCondition`, but explicitly marking a condition as False won't hurt. I might add it in a different commit/follow up because the focus is to see if this pattern can be finalized first.

controllers/glance_controller.go
```diff
@@ -493,7 +493,7 @@ func (r *GlanceReconciler) reconcileInit(
 
 func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glancev1.Glance, helper *helper.Helper) (ctrl.Result, error) {
 	r.Log.Info(fmt.Sprintf("Reconciling Service '%s'", instance.Name))
-
+	instance.Status.ObservedGeneration = instance.Generation
```
gibizer, line 496: I would move this to be right after we reset the conditions at L175. For me this more belongs to the status initialization code than to the Normal reconcile path. While it is not likely to cause trouble I think for consistency we want to keep the observedGeneration bumped during delete too.

controllers/glance_controller.go
```diff
@@ -1192,3 +1195,26 @@ func (r *GlanceReconciler) ensureDB(
...
+	listOpts := []client.ListOption{
+		client.InNamespace(instance.Namespace),
+	}
+	if err := r.Client.List(context.Background(), glances, listOpts...); err != nil {
+		r.Log.Error(err, "Unable to retrieve Glance CRs %w")
+		return cond, err
+	}
+	for _, item := range glances.Items {
+		if item.Generation == item.Status.ObservedGeneration {
+			cond = true
```
gibizer, line 1216: you probably want to return true only when all the glanceAPIs are up-to-date. But this code now returns true if any of the GlanceAPI is up-to-date.

I think something like:
```golang
	for _, item := range glances.Items {
		if item.Generation != item.Status.ObservedGeneration {
			return false, nil
        }
    }
    return true, nil
```

controllers/glanceapi_controller.go
```diff
@@ -540,6 +540,7 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
 	configVars := make(map[string]env.Setter)
 	privileged := false
 	imageConv := false
+	instance.Status.ObservedGeneration = instance.Generation
```
gibizer, line 543: ditto I would move this to right after the status condition reinit

controllers/glanceapi_controller.go
gibizer, line 1: (I hate that I cannot comment to not changed code in github)
I think one more step is necessary to make the whole chain of logic race free from imageURL update perspective. This can be done as a followup too.

https://github.com/openstack-k8s-operators/glance-operator/blob/407c91f5e7aa4f5cbef26d623067e6c38d226ac5/controllers/glanceapi_controller.go#L808-L819
The logic that determines that the StatefulSet is ready needs to consider the generation and observedGeneration of the StatefulSet as well to ensure that the status the glanceAPI controller reads from the StatefulSet is talking about the latest StatefulSet Spec generation.

The linked code uses `instance.Status.ReadyCount` which is copied from the StatefulSet.Status at 
https://github.com/openstack-k8s-operators/glance-operator/blob/407c91f5e7aa4f5cbef26d623067e6c38d226ac5/controllers/glanceapi_controller.go#L779
I think this information should only be considered for the DeploymentReadyCondition if the StatefulSet.Status.ObservedGeneration is up-to-date as well.

controllers/glance_controller.go
fmount: yeah, I think it's more efficient always return false if one of the API is not up-to-date.

controllers/glance_controller.go
fmount: yeah, no issues but +1 for consistency

controllers/glance_controller.go
gibizer: It not just about efficiency as the original code would return true if one GlanceAPI is up-to-date but the other GanceAPI is not up-to-date. My suggested code returns false in that case.

controllers/glanceapi_controller.go
fmount: I'll follow up on this, I'd like to try something, but I didn't do that because I assume that within the glanceAPI, within a particular reconcile loop, the instance reference has the same generation for all the duration of the current reconcile, hence it matches as long as I bump the observedGeneration. This is pure theory, I'll test it first and update that condition with a follow up commit.

gibizer: I think (also just theory) when the CreateOrUpdate is called on the StatefulSet that updates imageURL the returned StatefulSet instance has its Spec.Generation bumped by k8s api server persisting the updated object, but the Status.observedGeneration will not be bumped as that will only be bumped when the StatefulSet controller starts reconciling the update.

fmount: of course the above can also be just a && condition for the `DeploymentReady` condition, it probably makes more sense to not "pollute" the current logic, and have it as a side check.

fmount: Ok, I'll push a new commit with that part so we can review.

controllers/glanceapi_controller.go
```diff
@@ -807,9 +808,14 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
 	}
 	// Mark the Deployment as Ready only if the number of Replicas is equals
 	// to the Deployed instances (ReadyCount), but mark it as True is Replicas
-	// is zero
-	if instance.Status.ReadyCount == *instance.Spec.Replicas || *instance.Spec.Replicas == 0 {
-		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
+	// is zero. In addition, make sure the controller sees the last Generation
+	// by comparing it with the ObservedGeneration set in the StateFulSet.
+	if (instance.Status.ReadyCount == *instance.Spec.Replicas || *instance.Spec.Replicas == 0) &&
+		(depl.GetStatefulSet().Generation == depl.GetStatefulSet().Status.ObservedGeneration) {
```
fmount, line 814: @gibizer https://github.com/openstack-k8s-operators/lib-common/pull/488 is needed to pass this point in envTests

controllers/glanceapi_controller.go
mrkisaolamb: I did some code digging and definitly this looks like good place to check generations and avoid race condition by using similar pattern like in kubectl to ensure that generations are matching

controllers/glanceapi_controller.go
```diff
@@ -807,9 +808,14 @@ func (r *GlanceAPIReconciler) reconcileNormal(ctx context.Context, instance *gla
 	}
 	// Mark the Deployment as Ready only if the number of Replicas is equals
 	// to the Deployed instances (ReadyCount), but mark it as True is Replicas
-	// is zero
-	if instance.Status.ReadyCount == *instance.Spec.Replicas || *instance.Spec.Replicas == 0 {
-		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
+	// is zero. In addition, make sure the controller sees the last Generation
+	// by comparing it with the ObservedGeneration set in the StateFulSet.
+	if (instance.Status.ReadyCount == *instance.Spec.Replicas || *instance.Spec.Replicas == 0) &&
```
gibizer, line 813: side question: Why the Replicas == 0 case needs to be handled separately? I assume that in that case pthe ReadyCount  == Replicas condition would be true eventually as well. (definitely something for a separate PR)

controllers/glanceapi_controller.go
gibizer: thanks. I landed that lib-common change.

pkg/glanceapi/statefulset.go
```diff
@@ -326,5 +326,7 @@ func StatefulSet(
 		statefulset.Spec.Template.Spec.NodeSelector = instance.Spec.NodeSelector
 	}
 
+	// Bump the ObservedGeneration for the StatefulSet CR
+	statefulset.Status.ObservedGeneration = statefulset.Generation
```
gibizer, line 330: We should not do this. The statefulset controller is expected to do such a bump when it detects the new generation.

gibizer (changes requested):

controllers/glanceapi_controller.go
bogdando: Can we be sure that both values we got are the most recent? Operators sdk provides an [api objects cache ](https://logserver.rdoproject.org/70/370/773160e9ba239bb3547715a8e9a406dabf1a1203/github-check/data-plane-adoption-osp-17-to-extracted-crc/ef2f701/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-openstack-wstzp/logs/openstackansibleee.log) "Using controller-runtime’s split client means that read operations (gets and lists) are read from a cache, and write operations are written directly to the API server."

pkg/glanceapi/statefulset.go
fmount: Ok, this is probably the result of an issue on my end. I didn't see this parameter and I wasn't able to compare it w/ Generation, but let me revisit this part.

controllers/glanceapi_controller.go
fmount: it was because `ReadyCount` wasn't always initialized, and if it doesn't exist that comparison will never be true. A different approach is to always initialize `ReadyCount = 0` and remove that additional condition. Definitely something I think it will be part of a follow up.

fmount: https://github.com/openstack-k8s-operators/glance-operator/pull/502

controllers/glanceapi_controller.go
gibizer: If the status is old (cached) that is OK, we will reconcile as it does not match the generation.
If the status is freshly read bu the spec is old, then we could have a situation where we compare an old spec generation with an old status observed generation and we report readiness too early. However I think it will be not the case here as the createOrUpdate is supposed to return the result of the POST API call instead of doing an independent read after the POST.

gibizer (approved):
This looks good to me.

PR #504, opened 2024-04-11
Print object diff in ValidateUpdate webhook call

gibizer:
Knowing what is changing in each resource update helps troubleshooting why a resource keeps reconciling.


gibizer:
abandoning in favor of #503

PR #505, opened 2024-04-12
Verify network only if replicas is greater than zero

fmount:
This patch represents an optimization for the current code where we entirely skip the `NAD` verification in case `replicas == 0`.


abays (approved):
/lgtm

PR #507, opened 2024-04-12
Register Edge instances to dnsmasq

fmount:
We need to make sure that EDPM computes reach the internal GlanceAPI Edge endpoint. This patch let the internal endpoint to get the hostname annotation that allows the API to be registered in dnsmasq.


fmount:
Closing this patch because we have all we need for Edge when network isolation is used and `serviceOverrides` are passed.

PR #511, opened 2024-04-17
Improve how the memcached instance is handled

fmount:
Instead of flooding the log with multiple stack traces, we can gracefully handle the case where the `memcached` instance is not ready, log a message (other than setting the `Condition`) and `requeue` the request. If something changes after the first bootstrap, we should trigger a reconciliation loop  in the underlying controller that is supposed to deploy the API using that memcached instance. For this reason a similar check is added within the `glanceapi_controller`, that now watches for the same resource.

### DIFF
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -184,12 +184,10 @@ func (r *GlanceAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	if instance.Status.NetworkAttachments == nil {
 		instance.Status.NetworkAttachments = map[string][]string{}
 	}
-
 	// Handle service delete
 	if !instance.DeletionTimestamp.IsZero() {
 		return r.reconcileDelete(ctx, instance, helper)
 	}
-
 	// Handle non-deleted clusters
 	return r.reconcileNormal(ctx, instance, helper, req)
 }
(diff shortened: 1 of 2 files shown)


controllers/glanceapi_controller.go
```diff
@@ -306,6 +304,35 @@ func (r *GlanceAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			return nil
+		}
+
+		for _, cr := range glanceAPIs.Items {
+			if o.GetName() == cr.Spec.MemcachedInstance {
+				name := client.ObjectKey{
+					Namespace: o.GetNamespace(),
+					Name:      cr.Name,
+				}
+				r.Log.Info(fmt.Sprintf("Memcached %s is used by SwiftStorage CR %s", o.GetName(), cr.Name))
```
abays, line 326: `s/SwiftStorage/GlanceAPI`

controllers/glanceapi_controller.go
```diff
@@ -306,6 +304,35 @@ func (r *GlanceAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return nil
 	}
 
+	memcachedFn := func(ctx context.Context, o client.Object) []reconcile.Request {
+		result := []reconcile.Request{}
+
+		// get all SwiftStorage CRs
```
abays, line 310: `s/SwiftStorage/GlanceAPI`

controllers/glanceapi_controller.go

controllers/glanceapi_controller.go

abays (approved):
/lgtm

PR #513, opened 2024-04-19
Fix Inject_metadata plugin

fmount:
This patch allows to build a `Glance` `CR` (that can be reused in `kuttl` context) with `inject metadata` plugin. We're not taking assumptions about the backend used by the deployment.

### DIFF
--- a/config/samples/import_plugins/inject_metadata/inject_metadata.yaml
+++ b/config/samples/import_plugins/inject_metadata/inject_metadata.yaml
@@ -1,50 +1,13 @@
-# Sample using Ceph as a glance backend with inject metadata plugin
-# Requires a running Ceph cluster and its `/etc/ceph` files in secret `ceph-conf-files`
-# This can be achieved with the `ceph` target of `install_yamls`
+# Inject inject_metadata config
 apiVersion: glance.openstack.org/v1beta1
 kind: Glance
 metadata:
   name: glance
 spec:
   serviceUser: glance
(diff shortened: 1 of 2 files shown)


config/samples/import_plugins/inject_metadata/inject_metadata.yaml
```diff
@@ -1,50 +1,13 @@
...
-    default_backend = default_backend
-    [default_backend]
-    rbd_store_ceph_conf = /etc/ceph/ceph.conf
-    store_description = "RBD backend"
-    rbd_store_pool = images
-    rbd_store_user = openstack
     [image_import_opts]
     image_import_plugins = [inject_image_metadata]
     [inject_metadata_properties]
     ignore_user_roles = admin,user1
```
konan-abhi, line 12: I think we should avoid user1 role since we are not creating it.
Also when this kuttle test will run which user it will use, hopefully the user should not have admin privileges otherwise the metadata will not be injected.

Which means we need to assert for admin as well as non-admin user as well

fmount: I didn't change that line because it was already in the sample. I think for kuttl we should just add the openstackclient and create the user, otherwise (if it's sufficient), we can just do tests against the admin role.

konan-abhi (approved):
Since this is an example it looks good, need to make sure that we have specified users in our deployment in order this to work.

Thank you!

PR #514, opened 2024-04-19
Build disk_format sample in a kuttl compatible way

fmount:
The previous sample was based on the `OpenStackControlPlane`, but `kuttl` uses a single `Glance CR`. This patch aligns this sample and it can be now built and executed via `kuttl`.


konan-abhi (approved):
Looks good!

PR #518, opened 2024-04-21
Add disk_format kuttl tests

maximsava12:
This patch introduces kuttl tests to check the resources that are supposed to be generated when the disk_format is defined.


fmount:
This should be rebased at least on top of https://github.com/openstack-k8s-operators/glance-operator/pull/520, so we'll test it once the dependencies are in place.

config/samples/disk_formats/create-image.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+set -evx
+# Create a image with supported disk format
+#
+# The scripts assumes:
+#
+# 1. an available openstack cli
+# 2. control plane configured with disk formats
+# 2. a single layout (file / NFS) backend and disk format is deployed
```
fmount, line 9: nit: s/2/3

config/samples/disk_formats/create-image.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+set -evx
+# Create a image with supported disk format
+#
+# The scripts assumes:
+#
+# 1. an available openstack cli
+# 2. control plane configured with disk formats
```
fmount, line 8: nit: instead of control plane we can say: a `glanceAPI` configured with ...

config/samples/disk_formats/create-image.sh
```diff
@@ -0,0 +1,63 @@
...
+        "${IMAGE_NAME}" 
+
+    ID=$(openstack image list | awk -v img=$IMAGE_NAME '$0 ~ img {print $2}')
+    echo "Image ID: $ID"
+    sleep "${TIME}"
+
+    if [ -z "$ID" ]
+    then
+      echo "Could not create image  " >&2
+      openstack image delete "$ID"
```
fmount, line 35: I'm not sure about this `if` statement: why you `delete` an image if `-z "$ID"` (which is supposed to be an empty string).
You can return w/ `exit 1` and cleanup the environment with:

```
openstack image list -c ID -f value | xargs -n 1 openstack image delete
```

config/samples/disk_formats/create-image.sh
```diff
@@ -0,0 +1,63 @@
...
+      status=$(openstack image delete "$ID" | awk '/status/{print $4}')
+      printf "Image Status: %s\n" "$status"
+      exit 1
+    else
```
fmount, line 39: this `else` statement is not required

config/samples/disk_formats/create-image.sh
```diff
@@ -0,0 +1,63 @@
...
+      echo "Continue"
+    fi
+
+    # Stage 2 - Check the image is active
+    openstack image list
```
fmount, line 44: other then printing the list of images in the `stdout`, this `image list` doesn't do nothing special. If you reach this point, the image create succeeded, and you can just show the image status. 
However, I'd like the image status influence the `exit code`: if (somehow) the image is not active, we shouldn't return 0.

config/samples/disk_formats/create-image.sh
```diff
@@ -0,0 +1,63 @@
...
+    status=$(openstack image show "$ID" | awk '/status/{print $4}')
+    printf "Image Status: %s\n" "$status"
+
+    # Stage 3 - Delete the image 
+    openstack image delete "$ID"
+    status=$(openstack image delete "$ID" | awk '/status/{print $4}')
```
fmount, line 50: Does an `image delete` status returns something like `Deleted` (I think the info are retained even if the image is no longer available) ? In this case this should also influence the exit code.
However I think you can just remove this `Stage3` (unless is useful), and create a `Stage 0` that ensures the environment is clean __before__ starting any test.

```
# Stage 0 - Delete any pre-existing image
openstack image list -c ID -f value | xargs -n 1 openstack image delete
```

fmount (changes requested):

config/samples/disk_formats/create-image.sh
maximsava12: Done

config/samples/disk_formats/create-image.sh
maximsava12: Done

config/samples/disk_formats/create-image.sh
```diff
@@ -0,0 +1,66 @@
...
+    sleep "${TIME}"
+
+    if [ -z "$ID" ]
+    then
+      openstack image list -c ID -f value | xargs -n 1 openstack image delete
+      exit 1
+    fi
+
+    # Stage 2 - Check the image is active
+    openstack image list
```
fmount, line 44: do we need `openstack image list` here? If it's just for debug purposes, I'm wondering if we want to have a `DEBUG` parameter that can be set to either 0 or 1 to trigger such commands that increase verbosity for troubleshooting purposes.

test/kuttl/tests/glance_disk_formats/00-deploy.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      oc kustomize ../../../../config/samples/layout/single | oc apply -n $NAMESPACE -f -
```
fmount, line 5: remove this line (L5), and instead change `config/samples/disk_format/disk_format.yaml` to be like the following:

```
# Inject inject_metadata config
kind: Glance
metadata:
  name: glance
spec:
  serviceUser: glance
  databaseInstance: openstack
  databaseAccount: glance
  keystoneEndpoint: default
  customServiceConfig: |
    [image_format]
    disk_formats=raw,iso
  glanceAPIs:
    default:
      type: single
      replicas: 1
```
This way you can directly deploy what you need with a single command.
Also, given this layout covers a kuttl test, can we move `disk_format` under `layout` folder?

test/kuttl/tests/glance_disk_formats/02-cleanup.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      oc kustomize ../../../../config/samples/layout/single | oc delete -n $NAMESPACE -f -
```
fmount, line 5: L5 here can be changed to be:

```
oc kustomize ../../../../config/samples/layout/disk_format | oc delete -n $NAMESPACE -f -
```
after you move everything under `layout` directory.

test/kuttl/tests/glance_disk_formats/01-create_image.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+    - command: |
+        oc -n $NAMESPACE exec -it openstackclient --stdin=false -- /bin/sh -c "DOMAIN=glance-default-single.$NAMESPACE.svc ./scripts/create-image.sh raw"
```
fmount, line 5: Looking at the script I think you don't need `"DOMAIN=glance-default-single.$NAMESPACE.svc`.
Remove it from L5 and L7.

fmount (changes requested):
Made another pass and I think we're going to the right path. It will be something we can test and merge after a few minor changes

config/samples/disk_formats/create-image.sh
fmount: I see you resolved this. Thank you.

test/kuttl/tests/glance_disk_formats/01-create_image.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+    - command: |
```
fmount, line 4: -1:
```
logger.go:42: 11:46:39 | glance_disk_formats/0-deploy | test step failed 0-deploy
case.go:378: failed in step 0-deploy
case.go:380: pods "openstackclient" not found
```

As done for image_cache and dist image import, you need to update the openstackclient configmap to get the script that is run during the test (and deploy the pod).

config/samples/disk_formats/create-image.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
```
fmount, line 1: -1: Can you rename this script? It would be misleading a `create-image.sh` script (that does a different validation) in the openstackclient pod.

fmount (changes requested):

maximsava12:
Closing.
Glance Disk formats tested with tempest.

PR #520, opened 2024-04-22
Include openstackclient in samples for testing purposes (kuttl)

fmount:
This patch introduces the `openstackclient` `Pod` among the `config/samples` to make sure we're able to load a set of scripts and execute them against a deployed `GlanceAPI`: in an environment where the `openstack-operator` is not deployed (e.g., `kuttl` or `dev`), this helps to test new functionalities, we can verify that we have no regressions for each proposed patch and it represents a basic building block to establish a framework for proper `kuttl` test(ing).

### DIFF
--- a/config/samples/openstackclient/README.md
+++ b/config/samples/openstackclient/README.md
@@ -1,0 +1,142 @@
+# OPENSTACKCLIENT
+
+The `openstackclient` is used to interact with an existing (even minimal)
+control plane, and it can be used by `kuttl` tests to execute scripts that are
+supposed to verify that a given functionality works as expected.
+It assumes that `keystone` is available and the ConfigMap containing cloud
+credentials already exist and can be mounted to the target `Pod`.
+
+## Deploy the openstackclient via kustomize
+
+It is possible to inject scripts providing the right mounts to the
+`openstackclient` via [kustomize](https://kustomize.io).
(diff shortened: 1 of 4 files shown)


config/samples/openstackclient/README.md
```diff
@@ -0,0 +1,140 @@
...
+```bash
+oc -n <namespace> kustomize --load-restrictor LoadRestrictionsNone $pwd/openstackclient | oc apply -f -
+```
+
+**Note:**
+
+All the theory describe above assumes the `scripts` directory is used as source
+to build a `ConfigMap` that is then mounted to the `Pod`. **Before** going
+through the described steps, make sure a `ConfigMap` containing the scripts
+directory exists, or at leat each script entry is mentioned in kustomization.yaml.
```
abays, line 75: nit: `s/leat/least`

fmount: thank you, fixed in the last PS.
Note that this change is connected with a follow up that I still have to propose  [1], and it establishes an easy way to help QEs to build a set of kuttl tests that are supposed to test the service functionalities.

[1] https://github.com/fmount/glance-operator/commit/db5a16bd3ea0c2d19e505b8efcfb6ff1b04b67a5

abays (approved):
/lgtm

PR #523, opened 2024-04-24
Move to golangci-lint-full

fmount:
This patch moves the pre-commit-config to golangci-lint-full and clean up most of the code leftovers.
In particular:
- unused func params are removed
- better use of ctx context.Context
- fix leftovers from previous PRs (e.g., ineff assignments)
- fix revive error for envTests

### DIFF
--- a/pkg/glance/pvc.go
+++ b/pkg/glance/pvc.go
@@ -14,8 +14,8 @@ func GetPvc(api *glancev1.GlanceAPI, labels map[string]string, pvcType PvcType) 
 	// that will be customized in case the pvc is requested
 	// for cache purposes
 	var err error
-	requestSize := api.Spec.StorageRequest
-	pvcName := ServiceName
+	var requestSize string
+	var pvcName string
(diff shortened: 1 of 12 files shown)


pkg/glance/pvc.go
```diff
@@ -14,8 +14,10 @@ func GetPvc(api *glancev1.GlanceAPI, labels map[string]string, pvcType PvcType)
 	// that will be customized in case the pvc is requested
 	// for cache purposes
 	var err error
-	requestSize := api.Spec.StorageRequest
-	pvcName := ServiceName
+	var requestSize string
+	var pvcName string
+	//requestSize := api.Spec.StorageRequest
+	//pvcName := ServiceName
```
gibizer, line 20: drop commented out code

gibizer (commented):
thanks for applying the fix to glance

pkg/glance/pvc.go
fmount: Done

gibizer (approved):

PR #524, opened 2024-04-24
Bump glance-api probe timeout

SeanMooney:
when deploying locally in crc glance can crashloop
indefintly because the readiness and liveness probe
were set to strictly.

The nova-operator relaxed the probes becuase of tempest failures
in https://github.com/openstack-k8s-operators/nova-operator/pull/381

This change updates the glance values to match.


fultonj:
Seems reasonable to me +2

pkg/glanceapi/statefulset.go
```diff
@@ -60,12 +61,13 @@ func StatefulSet(
 		PeriodSeconds:    10,
 	}
 	livenessProbe := &corev1.Probe{
-		PeriodSeconds:       3,
-		InitialDelaySeconds: 3,
+		TimeoutSeconds:      30,
+		PeriodSeconds:       30,
+		InitialDelaySeconds: 30,
```
gibizer, line 66: that seems to be excessive compared to other patches

SeanMooney: fixed

gibizer (approved):
thanks

SeanMooney:
/test glance-operator-build-deploy-kuttl

Service Unavailable, image "docker.io/rdotripleomirror/openstack-operator-bundle:d8636775a73da79d0187-1783476226313162752": not found]

fmount:
so looks like this represents atm the worst case scenario (crc on a node w/ a limited amount of resources and the whole ctlplane deployed): I recently did something similar in [manila-operator/pull/263](https://github.com/openstack-k8s-operators/manila-operator/pull/263). Considering that I'm not sure what should be a good default, I'm ok with this patch.

SeanMooney:
ya the original values were chosen for nova when we first added Tempest which was using a nested CRC VM in the zull vms. so without nested virt or on a slightly newer laptop, the manila values might be fine. i was debating if we should we consider normalising this across operator or not.

30 was more or less chosen as the upper reasonable limit.

i.e. if it takes more than 30s to get a response from the API then your API is effectively non-functional and we should treat it as such.

lowering that to 10s might also be appropriate. i think we initially tired a lower value but I'm not quite sure. its been a while since we tweaked nova to stablisie tempest and not long after we moved to extracted crc but we never revisited the values.

i think what we do know is the 3/5 second defaults we get form the SDK are to agressive.

PR #529, opened 2024-05-08
Update webhook

stuggi:
Updates webhook to be able to be call from the openstack-operator webhook on the GlanceSpecCore, like for other operators in [1].

Also adds check for valid service override endpoint type on create and update.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/505

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/d2703d3a321c979dacaca95b5d4a634bf116e0db/apis/core/v1beta1/openstackcontrolplane_webhook.go#L181

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/google/go-cmp v0.6.0
 	github.com/gophercloud/gophercloud v1.11.0
 	github.com/openstack-k8s-operators/keystone-operator/api v0.3.1-0.20240429164853-7e1e3b111ee9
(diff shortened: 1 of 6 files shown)


fmount (approved):
I didn't try this patch but from the code point of view seems correct.
I think we can improve the validation webhook tests but what we have looks ok.

fmount:
ok I checked locally and looks safe, sorry for the delay, I knew it was ok in terms of code but I wanted to test it locally to make sure we preserve the current layout.

PR #530, opened 2024-05-13
kuttl multiple test asserts

gibizer:
- **[kuttl]Fix false positive asserts**
- **[pre-commit][kuttl]Check for multiple TestAsserts**

PR #531, opened 2024-05-14
Fix kuttl TestAsserts that look for glanceAPI endpoints

fmount:
We currently have issues in the `GlanceAPI` `TestAsserts` because they do not take into account the api name. As done for `tls` in the past, this patch changes the `go-template` to match the pattern `<apiName>-<endpointType>`. However, given the "-" character is not allowed, a workaround is used with the `.index` on the `.status.apiEdnpoints`.

### DIFF
--- a/test/kuttl/tests/glance_single/01-assert.yaml
+++ b/test/kuttl/tests/glance_single/01-assert.yaml
@@ -149,8 +149,8 @@ kind: TestAssert
 namespaced: true
 commands:
   - script: |
-      template='{{.status.apiEndpoint.internal}}{{":"}}{{.status.apiEndpoint.public}}{{"\n"}}'
-      regex="http:\/\/glance-internal.$NAMESPACE.*:http:\/\/glance-public.$NAMESPACE.*"
+      template='{{ index .status.apiEndpoint "default-internal" }}{{ ":" }}{{ index .status.apiEndpoint "default-public" }}'
+      regex="http:\/\/glance-default-internal.$NAMESPACE.*:http:\/\/glance-default-public.$NAMESPACE.*"
       apiEndpoints=$(oc get -n $NAMESPACE Glance glance -o go-template="$template")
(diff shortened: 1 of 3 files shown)


test/kuttl/tests/glance_single/01-assert.yaml
```diff
@@ -149,8 +149,8 @@ kind: TestAssert
 namespaced: true
 commands:
   - script: |
-      template='{{.status.apiEndpoint.internal}}{{":"}}{{.status.apiEndpoint.public}}{{"\n"}}'
-      regex="http:\/\/glance-internal.$NAMESPACE.*:http:\/\/glance-public.$NAMESPACE.*"
+      template='{{ index .status.apiEndpoint "default-internal" }}{{ ":" }}{{ index .status.apiEndpoint "default-public" }}'
```
fmount, line 152: I see the following error locally: 

```
error: error executing template "{{ index .status.apiEndpoint \"default-internal\" }}{{ \":\" }}{{ index .status.apiEndpoint \"default-public\" }}": template: output:1:3: executing "output" at <index .status.apiEndpoint "default-internal">: error calling index: index of untyped nil
```
when the `NAMESPACE` is not set.
Let's see the CI execution, but we might need to properly set it!

fmount: I see the same error in #530, and I have to investigate how the `namespace` is passed to these scripts.

konan-abhi (approved):

test/kuttl/tests/glance_single/01-assert.yaml
fmount: @gibizer if you rebase #530 on this I think we can see if we still get this error ^

gibizer: Done

gibizer (approved):

PR #532, opened 2024-05-17
Setting quota limits fails with certificate error

fmount:
This patch fixes the interaction with keystone when quota is registered. As in [1], this change solves the problem by getting a scoped client from keystone.

Fixes: [OSPRH-6766](https://issues.redhat.com/browse/OSPRH-6766)

[1]https://github.com/openstack-k8s-operators/keystone-operator/blob/main/api/v1beta1/keystoneapi.go#L143


fmount:
> do we need the `GetAdminServiceClient()` in here? can we just use the GetScopedAdminServiceClient() from keystone? [openstack-k8s-operators/keystone-operator#417](https://github.com/openstack-k8s-operators/keystone-operator/pull/417) will add cert validation for the client connection.

Interesting, I didn't know about that function, especially that we can pass the scope! I will try that and it would simplify the code here in the operator.

konan-abhi (approved):

PR #536, opened 2024-05-22
Set EndpointID required for glance configuration

fmount:
This patch revisits the config generation and retrieves the `endpointID` to make sure the `oslo_limits` section is properly configured.
For this reason the `generateServiceConfig` function is moved later in the process (after the `ensureKeystoneEndpoints`)
to make sure all the data are present when the config template is rendered and the secret generated.


controllers/glanceapi_controller.go
```diff
@@ -1320,6 +1330,36 @@ func (r *GlanceAPIReconciler) deleteJob(
...
+		r.Log.Info("EndpointID not ready")
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.ServiceConfigReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			condition.ServiceConfigReadyErrorMessage,
+			err.Error()))
+		return epID, nil
+	}
+	epID = ep.Status.EndpointIDs[string(epType)]
```
abays, line 1359: If `err` was "not found", and therefore `ep` is still as it was post-L1338, is there a zero-value in `ep.Status.EndpointIDs` such that it won't hit a null pointer error?

fmount: @abays I tried something like this [1] to see how golang deals with `omitempty` and fields that are not set in a struct (it would be `.Status` of the CR in our case), but if this changes (e.g., we remove `omitempty` in keystone), we might have a problem here, so I think we can do the assignment only when `EndpointIDs` is not `nil`.

[1] https://go.dev/play/p/Qd1HwxmuW6y

controllers/glanceapi_controller.go
```diff
@@ -1039,6 +1043,12 @@ func (r *GlanceAPIReconciler) generateServiceConfig(
 		"VHosts":       httpdVhostConfig,
 	}
 
+	// Only set EndpointID parameter when the Endpoint has been created and the
+	// associated ID is set in the keystoneapi CR
+	if len(endpointID) > 0 {
```
fmount, line 1048: I'm still not sure about this part: should I delay the config generation here if the `endpointID` is not present yet? The alternative is to update the secret when it's generated, but that might be an unnecessary patch and Pods rollout.
On the other hand this config is only required for Quota, so I'd like to reduce the scope of this code.

fmount: I agree this seems a correct and safe check.
In case we have multiple APIs, let's say `default` and `default1`:

```yaml
glanceAPIs:
  keystoneEndpoint: default
  default1:
    type: single
   default:
    type: single
```
and `keystoneEndpoint: default`, the `keystoneendpoint` CR will be:

```
$ oc get keystoneendpoints
NAME                    STATUS   MESSAGE
glance-default-single   True     Setup complete
```

and the other API (`default1`), shouldn't be stuck because it waits for an endpoint that won't be set.
In case the endpoint is switched in the main CR (`keystoneendpoint: default1`), the previous endpoint CR is deleted, and a new one is created, with a different `endpointsID`.

controllers/glanceapi_controller.go
```diff
@@ -1320,6 +1330,38 @@ func (r *GlanceAPIReconciler) deleteJob(
...
+	// instead of looking for a different API that might not exist or have issues
+	// we can't control here
+	if instance.Spec.APIType == glancev1.APIExternal {
+		epType = endpoint.EndpointPublic
+	}
+	err := r.Get(ctx, types.NamespacedName{Namespace: instance.Namespace, Name: instance.Name}, ep)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		// Request object not found, GlanceAPI can't be executed with this config
+		r.Log.Info("EndpointID not ready")
+		instance.Status.Conditions.Set(condition.FalseCondition(
```
fmount, line 1351: -1 to this: if we have multiple APIs this is not going to work: we won't have a CR because it shouldn't be created, but we shouldn't block the process.

controllers/glanceapi_controller.go
```diff
@@ -1320,6 +1330,36 @@ func (r *GlanceAPIReconciler) deleteJob(
...
+	}
+	err := r.Get(ctx, types.NamespacedName{Namespace: instance.Namespace, Name: instance.Name}, ep)
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
+			// Just log the keystoneEndpoint CR does not exist so we have evidence
+			// in the operator output: it's not necessarily an error that should
+			// trigger a reconciliation loop
+			r.Log.Info(fmt.Sprintf("EndpointID not found for glanceAPI %s", instance.Name))
+		}
+		return epID, nil
```
konan-abhi, line 1355: here epID will be blank, right?
And why we are returning err as nil, just to trigger reconcile loop?

fmount: Yes, L1339 explicitly initialize it to an empty string, also we do not return err because it might happen that a different glanceAPI (not registered in keystone) is reconciling: in this case it won't find a `keystoneEndpoint` CR and we shouldn't block the deployment because `epID` is not available.

konan-abhi (approved):

PR #542, opened 2024-05-29
Enable TLS for distributed image import

fmount:
When a `Glance` `SVC` is created, an associated `headless service` is generated: this is used to make `glance-direct` work and enable distributed image import.
However, when `certificates` are created, we need to include the `SubjectName` related to the `headless service`, otherwise instances of the same `Pod` (replicas) are able to resolve each other, but not able to proxy requests via `HTTPS`.
This change annotates the created `SVC` with the corresponding `headless` service, and the `openstack-operator` can look at this
annotation to build a list of additional `SubjectName(s)` that needs to be included in the `CertificateRequest`.

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/817
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/516

Jira: https://issues.redhat.com/browse/OSPRH-7276

### DIFF
--- a/controllers/glance_common.go
+++ b/controllers/glance_common.go
@@ -152,7 +152,7 @@ func GetHeadlessService(
 	helper *helper.Helper,
 	instance *glancev1.GlanceAPI,
 	serviceLabels map[string]string,
-) (ctrl.Result, error) {
+) (ctrl.Result, string, error) {
 
 	endpointName := instance.Name
 	// The endpointName for headless services **must** match with:
@@ -186,7 +186,7 @@ func GetHeadlessService(
 			condition.SeverityWarning,
 			condition.ExposeServiceReadyErrorMessage,
 			err.Error()))
-		return ctrl.Result{}, err
+		return ctrl.Result{}, endpointName, err
 	}
 
 	svc.AddAnnotation(map[string]string{
@@ -209,17 +209,17 @@ func GetHeadlessService(
 			condition.ExposeServiceReadyErrorMessage,
 			err.Error()))
 
-		return ctrlResult, err
+		return ctrlResult, endpointName, err
 	} else if (ctrlResult != ctrl.Result{}) {
 		instance.Status.Conditions.Set(condition.FalseCondition(
 			condition.ExposeServiceReadyCondition,
 			condition.RequestedReason,
 			condition.SeverityInfo,
 			condition.ExposeServiceReadyRunningMessage))
-		return ctrlResult, nil
+		return ctrlResult, endpointName, nil
 	}
 
-	return ctrlResult, nil
+	return ctrlResult, endpointName, nil
 }
 
 // GetPvcListWithLabel -
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -429,6 +429,21 @@ func (r *GlanceAPIReconciler) reconcileInit(
 			},
 		)
 
+		// For each StatefulSet associated with a given glanceAPI (single, internal, external)
+		// we create a headless service that allow to resolve pods by hostname (using kube-dns)
+		// and it allows to enable the glance-direct import method
+		ctrlResult, headlessSvcName, err := GetHeadlessService(
+			ctx,
+			helper,
+			instance,
+			serviceLabels,
+		)
+		if err != nil {
+			// The ExposeServiceReadyCondition is already marked as False
+			// within the GetHeadlessService function: we can return
+			return ctrlResult, err
+		}
+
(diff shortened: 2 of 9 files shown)


controllers/glanceapi_controller.go
```diff
@@ -458,6 +473,11 @@ func (r *GlanceAPIReconciler) reconcileInit(
 
 		svc.AddAnnotation(map[string]string{
 			service.AnnotationEndpointKey: endpointTypeStr,
+			// Add to the current service an annotation that refers to the associated
+			// headless service: this information will be used by the openstack-operator
+			// to have an additional SubjectName that allows to reach every replica with
+			// an https endpoint
+			"headless": headlessSvcName,
```
fmount, line 480: Holding the patch for now because I might need @stuggi's input on this: here I annotate the SVC with this key that will be resolved by the openstack-operator to build an additional `sn` entry in the `CertificateRequest` (I'll come up with a patch today for that part).
Should we need to choose a more generic annotation structure/naming to pass that information?
e.g.

```
tls:
  additionalSubjectName:
     - {{ headlessSvcName }}
```

controllers/glance_common.go
```diff
@@ -152,7 +152,7 @@ func GetHeadlessService(
 	helper *helper.Helper,
 	instance *glancev1.GlanceAPI,
 	serviceLabels map[string]string,
-) (ctrl.Result, error) {
+) (ctrl.Result, string, error) {
```
konan-abhi, line 155: Since we are passing instance as a parameter to this function, do we really need to fetch endpointName inside this function and return it (since it is not used for any other purpose), can't we directly use instance.Name outside?

fmount: I see what you mean, and in theory we can compute it.
However, it's a little price to pay to not have to rewrite (in the controller) the if statement of L161 to distinguish between `split` vs `single` layout.
Since all the logic is already in place here, I can just simplify the logic in the main controller and not have to worry about the nature of the `glanceAPI` (that has an impact over the name: `-external` vs `-internal` vs `-single` vs `-edge`).
Hopefully when we don't have to split (in 19?) we can simplify further this logic!

konan-abhi: Ack

konan-abhi (approved):

controllers/glanceapi_controller.go
stuggi: - yes lets add `additionalSubjectName`, or maybe `additionalSubjectNames` to https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/tls/tls.go#L40 . maybe plural to be able to add a comma separated list of additional SubjectNames?

- Where is the headless service name used/configured? do we have to configure it in the httpd.conf as a `ServerAlias`, iiuc right now we only pass in the svc name https://github.com/openstack-k8s-operators/glance-operator/blob/main/controllers/glanceapi_controller.go#L1015

fmount: Yes, sorry for being late to this work. I do agree we should have the lib-common part to manage a comma separated list of subject names. They can be built by the service operators, and the openstack-operator can apply them by accessing the annotation.
I'll follow up here in a bit.

fmount:
I see services annotated [1][2] as expected.

[1] [glance-default-public](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_glance-operator/542/pull-ci-openstack-k8s-operators-glance-operator-main-glance-operator-build-deploy-tempest/1798702788817981440/artifacts/glance-operator-build-deploy-tempest/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-acd8cc6209e20758e44299941b76028c61e4edc0f6c89200915592cd2118b35e/namespaces/openstack/services/glance-default-public.yaml)
[2] [glance-default-internal](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_glance-operator/542/pull-ci-openstack-k8s-operators-glance-operator-main-glance-operator-build-deploy-tempest/1798702788817981440/artifacts/glance-operator-build-deploy-tempest/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-acd8cc6209e20758e44299941b76028c61e4edc0f6c89200915592cd2118b35e/namespaces/openstack/services/glance-default-internal.yaml)

stuggi (approved):
/lgtm

PR #545, opened 2024-05-30
Rollout new pods when InputHash changes

fmount:
During the glanceapi_controller restructuring we lose the ability to rollout new pods when the config changes. This is because the inputHash is generated before the  generateConfig function is called. This patch restores the right behavior.


konan-abhi (approved):

abays (approved):
/lgtm

PR #548, opened 2024-06-05
Wait for DB writes only on reads

fmount:
In this patch we change the value of mysql_wsrep_sync_wait to 1 to only wait on reads.


konan-abhi (approved):
thank you!

PR #550, opened 2024-06-06
Add configurable API Timeouts

Akrog:
Multiple limitations exist with existing code:

- Human operator cannot configure Apache Timeout, so it's fixed at the defaults of 60 seconds.

- Default HAProxy timeout is 30 seconds, which is different than the default in OSP17 (60 seconds)

This patch adds an `apiTimeout` field to the `GlanceSpecCore` as well as the `GlanceAPITemplate` to allow human operators to simultaneously configure the timeouts for HAProxy and Apache.

The `apiTimeout` of the `GlanceSpecCore` defaults to 60 seconds, and the one in the `GlanceAPITemplate` defaults to whatever value the `GlanceSpecCore` has (default or set by the human operator).

The Apache timeout is set using its `TimeOut` parameter because that way it affects the httpd core timeouts as well as the `ProxyTimeout` (it defaults to `TimeOut`).

Having different timeouts for HAProxy and Apache is possible setting the HAProxy timeout in the `apiOverrides` and the Apache one with `apiTimeout`.

To be able to change the HAProxy value based on the `apiTimeout` with any update (and not just the first time) the code adds a custom annotation "api.glance.openstack.org/timeout" with the value that was initially set, this way flags it as being set by the glance-operator.

Jira: https://issues.redhat.com/browse/OSPRH-7415


fmount (approved):
/lgtm

PR #554, opened 2024-06-11
Simplify glance-operator pvc interface

fmount:
This patch simplifies the way we build the storage interface for glance.
In general `Glance` requires a `staging area` to manipulate data during an import operation.
It is possible to copy image data into multiple stores using image import workflow, and all the
`import methods` and `plugins` are based on this flow, hence some persistence is required for
the resulting Glance `Pod`.
With this PR, a `Storage` struct is exposed by the `glance-operator` API, and it helps to choose
the best layout for a particular `glanceAPI`.
When defining the storage requirements for a `GlanceAPI`, with this patch is possible to choose
between two different models:
- External
- PVC

`External` means that no PVCs are provisioned by the operator, and the human administrator should
take care about providing persistence via `ExtraMounts`. `PVC` is still the default interface, and they
are provisioned by the operator.

Jira: https://issues.redhat.com/browse/OSPRH-6719
Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/843


docs/dev/design-decisions.md
```diff
@@ -454,3 +454,279 @@ in the glance-operator we make the following assumptions:
...
+- PVC
+
+### External
+
+If `External` is chosen, no `PVCs` are created, and from a Storage point of
+view the `GlanceAPI` acts like a stateless instance, where no persistence is
+provided. In this case, it's expected that the human operator provides some
+persistence via `extraMounts`.
+For this particular scenario, the most popular choice is `NFS`, and it can be
+mapped to `/var/lib/glance`:
```
konan-abhi, line 489: I think it should be /var/lib/glance/os_glance_staging_store

docs/dev/design-decisions.md
```diff
@@ -454,3 +454,279 @@ in the glance-operator we make the following assumptions:
...
+      name: nfs
+    volumes:
+    - name: nfs
+      nfs:
+        path: <NFS PATH>
+        server: <NFS IP ADDRESS>
+```
+
+It's important to note that the configuration described in the sample above
+conflicts with the distributed image import feature that uses `glance-direct`.
```
konan-abhi, line 512: `glance-direc` import method.

docs/dev/design-decisions.md
fmount: +1 thanks

docs/dev/design-decisions.md
fmount: Added! Thanks!

config/samples/layout/base/glance_v1beta1_glance.yaml
```diff
@@ -11,5 +11,3 @@ spec:
   databaseInstance: openstack
   databaseAccount: glance
   secret: osp-secret
-  storageClass: local-storage
-  storageRequest: 10G
```
abays, line 15: Should this also have a `storage` replacement like the changes to the other samples?

fmount: I removed it because the other files (under layout/) replace this part w/ kustomize, but I think it's a good idea to keep it in case we need it for future layouts.

fmount: I will update it in the next PS

fmount: Done

abays (commented):
+1 from me.  Will give others a chance to review before approving

fmount:
> +1 from me. Will give others a chance to review before approving

ack thank you, I think I asked @konan-abhi to review and he was ok w/ this change, but we can wait for him to re-approve. Thank you @abays !

konan-abhi (approved):
Looks good, thank you!

PR #555, opened 2024-06-12
Fix glance when Glance CR name is not glance

Akrog:
This patch fixes Glance to support a name different from `glance` for the Glance CR.

There were 2 kind of issues when the name was different:

- Secrets were not consistently addressed, for example they could be created using the instance name but then addressed in the pod volumes using the `ServiceName` (`glance`).

- API name calculations assumed names always started with `ServiceName`.

This patch fixes this, and the way we figure out the name of the API is by making the `Glance` controller add a label (`api-name`) to the `GlanceAPI` CR.

By having the API name as a label the `GlanceAPI` controller no longer needs to figure out the name, it can just get it from the label.

Jira: [OSPRH-7396](https://issues.redhat.com//browse/OSPRH-7396)

### DIFF
--- a/api/v1beta1/glanceapi_types.go
+++ b/api/v1beta1/glanceapi_types.go
@@ -27,6 +27,8 @@ import (
 const (
 	// DeploymentHash hash used to detect changes
 	DeploymentHash = "deployment"
+	// APINameLabel - Label on a GlanceAPI that signals the name of the API
+	APINameLabel = "api-name"
 )
 
 // GlanceAPISpec defines the desired state of GlanceAPI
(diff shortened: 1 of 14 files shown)


fmount:
@Akrog I tested this change and I'm ok to merge it, I think it simplifies the logic of setting/getting the API names and it doesn't introduce any issue w/ the previous versions.

Akrog:
> I suggest to change the functional test to use the new name pattern with a random UID fragment in

Done

fmount:
This patch looks good to me, I extensively tested even with multiple API and seems to work fine.
We have logs where the label is properly set [1] and simplifies the get/set of an API name.
@Akrog I'm going to approve it.

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_glance-operator/555/pull-ci-openstack-k8s-operators-glance-operator-main-glance-operator-build-deploy-tempest/1803723316582682624/artifacts/glance-operator-build-deploy-tempest/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-998b38d11a0107b173fcca6d22b99231c0d7be2a2f43bf78fd3285315c295e4b/namespaces/openstack/crs/glanceapis.glance.openstack.org/glance-default-external.yaml

fmount (approved):
/lgtm

PR #558, opened 2024-06-17
Revert "Wait for DB writes only on reads"

stuggi:
This reverts commit 980c318f3f65662a03d74e8b61222539499f80ee.

The openstack-k8s-operators/mariadb-operator#229
switched the mariadb-operator to deploy in Active/Passive mode per default, instead of multimaster.


fmount (approved):
/lgtm

PR #560, opened 2024-06-20
Remove node_staging_uri config option

fmount:
Multistore is the default [1], and we do not need the old staging directory config option (that is in any case ignored in favor of os_glance_staging_store).

[1] https://github.com/openstack/glance/blob/master/doc/source/admin/multistores.rst


fmount:
@maximsava12 FYI

konan-abhi (approved):
Good idea to get it removed as it is of no use.

PR #561, opened 2024-06-20
Improve api logs

Akrog:
When the log gets rotated in the api container our logs will get
polluted with the following messages from the `tail` command:

```
tail: '/var/log/glance/glance-default-single.log' has become inaccessible: No such file or directory
tail: '/var/log/glance/glance-default-single.log' has appeared;  following new file
```

To prevent his from happening we will redirect the `stderr` from the
`tail` command into `/dev/null`.

This PR also reduces the size used by useless rotated logs.
 
The log file configured in glance api is only used to pass the data to
the log sidecar, which then outputs the data to stdout and then goes to
/var/log/pods and/or a centralized logging service.

Considering this, there is no need to keep 5 rotated logs, since we'll
also have them in the other locations and there is no way to get them
without actually going manually into the specific container within the
pod.

This patch proposes having a single rotated log (because passing 0 as
max_logfile_count doesn't work as we want in oslo.log) and reducing the
size to rotate from 50MB to 20MB.

### DIFF
--- a/pkg/glanceapi/statefulset.go
+++ b/pkg/glanceapi/statefulset.go
@@ -196,6 +196,8 @@ func StatefulSet(
 	if instance.Spec.APIType != glancev1.APISingle {
 		stsName = fmt.Sprintf("%s-api", instance.Name)
 	}
+
+	LogFile := string(glance.GlanceLogPath + instance.Name + ".log")
 	statefulset := &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      stsName,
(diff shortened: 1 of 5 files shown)


fmount:
I'm sorry there's extra work to do here, but we should update kuttl to let CI pass [1].

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_glance-operator/561/pull-ci-openstack-k8s-operators-glance-operator-main-glance-operator-build-deploy-kuttl/1804064753119662080/artifacts/glance-operator-build-deploy-kuttl/openstack-k8s-operators-kuttl/build-log.txt

fmount (approved):
/lgtm

PR #564, opened 2024-06-25
Add Required Infrastructure Annotations to CSV

fmount:
Implements: https://issues.redhat.com/browse/OSPRH-7968

### DIFF
--- a/config/manifests/bases/glance-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/glance-operator.clusterserviceversion.yaml
@@ -6,6 +6,11 @@ metadata:
     capabilities: Basic Install
     features.operators.openshift.io/disconnected: "true"
     features.operators.openshift.io/fips-compliant: "true"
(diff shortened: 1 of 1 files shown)


config/manifests/bases/glance-operator.clusterserviceversion.yaml
```diff
@@ -6,6 +6,11 @@ metadata:
     capabilities: Basic Install
     features.operators.openshift.io/disconnected: "true"
     features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "false"
+    features.operators.openshift.io/tls-profiles: "false"
```
fmount, line 10: I'm not sure about `tls-profiles` yet.

fmount: looks like the recommendation is to keep them to false until we have the opportunity to check and eventually change the bool values. In this case I think we're good to go.

abays (approved):
/lgtm

PR #569, opened 2024-07-02
Move to VerifySecret and build the hash of selected fields

fmount:
Currently we get the `osp_secret` and we build the hash of the entire secret. If an unrelated field is added or changed, this cause a rollout of a new glance pod.

This change does two main things:
1. replace the `GetSecret` call with `VerifySecret`
2. build a common `ensureSecret` function that is called by both controllers (top level and `glanceAPI`)

`EnvTests` are updated accordingly.

Jira: https://issues.redhat.com/browse/OSPRH-8193

### DIFF
--- a/controllers/glance_common.go
+++ b/controllers/glance_common.go
@@ -19,9 +19,9 @@ package controllers
 import (
 	"context"
 	"fmt"
-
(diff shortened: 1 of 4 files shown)


konan-abhi (approved):
Looks good to me!!

stuggi (approved):
/lgtm

PR #572, opened 2024-07-04
Adding patch rbac perm for serviceaccounts

stuggi:
We also took the opportunity and added patch to all the existing rbac rules that had update already to avoid similar issues in the future

Resolves: https://issues.redhat.com/browse/OSPRH-8363


fmount (approved):
/lgtm

fmount:
Thank you @stuggi !

PR #580, opened 2024-07-11
Prevent to create/update a Single API with a backend other than File

fmount:
File (or NFS) does not need split API, while all the other backends need a split layout. This patch adds an extra condition to prevent a GlanceAPI creation (or an update in case it exists) where a single layout can expose a backend different than File,
and it returns the following error:

```
OpenStackControlPlane Glance error occured admission webhook "vglance.kb.io" denied the request: Glance.glance.openstack.org "glance" is invalid: spec.glanceAPIs[default]: Invalid value: "default": Invalid backend configuration detected
```

Jira: [OSPRH-8501](https://github.com/fmount/glance-operator/pull/new/osprh8501)

### DIFF
--- a/api/v1beta1/glance_webhook.go
+++ b/api/v1beta1/glance_webhook.go
@@ -187,11 +187,22 @@ func isFileBackend(customServiceConfig string, topLevel bool) bool {
 // Check if the File is used in combination with a wrong layout
 func (r *GlanceSpecCore) isInvalidBackend(glanceAPI GlanceAPITemplate, topLevel bool) bool {
 	var rep int32 = 0
-	// For each current glanceAPI instance, detect an invalid configuration
+
+	// Do not take assumptions about the backend when replicas: 0, because it
+	// means the human operator has not made any choice or has no backend
+	// available yet.
+	if *glanceAPI.Replicas == rep {
+		return false
+	}
+	// For the current glanceAPI instance, detect an invalid configuration
 	// made by "type: split && backend: file": raise an issue if this config
-	// is found. However, do not fail if 'replica: 0' because it means the
-	// operator has not made any choice about the backend yet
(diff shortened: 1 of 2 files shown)


konan-abhi (approved):
May be we can add kuttl test later for verification.

api/v1beta1/glance_webhook.go
```diff
@@ -194,6 +194,11 @@ func (r *GlanceSpecCore) isInvalidBackend(glanceAPI GlanceAPITemplate, topLevel
 	if *glanceAPI.Replicas != rep && glanceAPI.Type == "split" && isFileBackend(glanceAPI.CustomServiceConfig, topLevel) {
 		return true
 	}
+	// Do not allow to deploy a glanceAPI with a backend other than File
+	// (Cinder, Swift, Ceph) with "type: single": we must split in that case
+	if glanceAPI.Type == APISingle && !isFileBackend(glanceAPI.CustomServiceConfig, topLevel) {
```
fmount, line 199: @konan-abhi I probably need to have here `*glanceAPI.Replicas != rep` as well, to not fail when the backend is still not available (and the human operator put `replicas: 0`)

konan-abhi (approved):
Thank you for the test!!

api/v1beta1/glance_webhook.go
```diff
@@ -309,6 +320,11 @@ func (r *GlanceSpecCore) ValidateUpdate(old GlanceSpecCore, basePath *field.Path
 		// the same name. This represent a valid use case and we shouldn't prevent
 		// to grow the deployment
 		if _, found := old.GlanceAPIs[key]; !found {
+			// Fail if an invalid configuration/layout is detected for the current
```
fmount, line 323: this part handles the situation where the CR exists and is update: if a new glanceAPI is added, we can't simply "continue" and allow any kind of configuration: we need to check if the backend is not valid (e.g. single + !file || split + file) and append an error to the APIs evaluation (that is a loop).

konan-abhi (approved):

PR #587, opened 2024-07-19
[validation] check glanceAPI names are valid

stuggi:
The glanceAPI controller creates StatefulSet for glanceapi to run. This adds a StatefulSet pod's label
"controller-revision-hash": "<statefulset_name>-<hash>" to the pod.
The kubernetes label is restricted under 63 char and the revision hash is an int32, 10 chars + the hyphen. Which results in a default statefulset max len of 52 chars. The statefulset name also contain the glance name and the glanceAPI type + 2 hyphens. So the max len also need to be reduced bye the length of those.

Also the name of the created rabbitmq instance must match a lowercase RFC 1123.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/532

Jira: https://issues.redhat.com/browse/OSPRH-8063


api/v1beta1/glance_webhook.go
```diff
@@ -215,6 +217,26 @@ func (r *Glance) ValidateCreate() (admission.Warnings, error) {
 	glancelog.Info("validate create", "name", r.Name)
 	var allErrs field.ErrorList
 	basePath := field.NewPath("spec")
+
+	for key, glanceAPI := range r.Spec.GlanceAPIs {
```
stuggi, line 221: I first had it in ValidateCreate for the GlanceSpecCore but then we'd have to pass down the glance name. I have no real preference, except that it we call the GlanceSpecCore from openstack-operator webhook and changing the params of the func will result in some CI dep madness.

abays (approved):
/lgtm

PR #589, opened 2024-07-26
Restore LastTransitionTime after setting Ready Condition

rabi:
instance.Status.Conditions.MarkTrue() always sets the current time as the LastTransitionTime and this results in infinite loop as the LastTransitionTime changes after every reconcile.

I think this issue is there in many other operator controllers, but we don't see the issue as the next reconcile is probably too quick to not change the LastTransitionTime (precision for which is in seconds). The issue showed up in dataplane controllers as we reconcile a number of services in a loop for every reconcile of nodeset.

Jira: [OSPRH-8811](https://issues.redhat.com//browse/OSPRH-8811)

### DIFF
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -139,12 +139,13 @@ func (r *GlanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 	// Always patch the instance status when exiting this function so we can
 	// persist any changes.
 	defer func() {
-		condition.RestoreLastTransitionTimes(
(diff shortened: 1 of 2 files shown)


rabi:
Looks like not needed here

stuggi:
yes I don't think its required for glance as we set the ReadyCondition at the end of the reconcile https://github.com/openstack-k8s-operators/glance-operator/blob/main/controllers/glanceapi_controller.go#L943 and restore the timestamps at the defer top. only if ready is unknown we mirror, which is wanted

PR #600, opened 2024-08-06
Index named input for NAD, unify across Watches and return nil error in case NAD not found to use RequeueAfter 

stuggi:
* Adds to watch NADs so that the controller reconciles when the configured NAD changes on the GlanceAPI.
* Also unifies the Watches across the named input parameters.
* Currently the reconciler returned both a non-zero result and a non-nil error.
The result will always be ignored if the error is non-nil and the non-nil error causes reqeueuing with exponential backoff.
In case of NotFound return nil that the ReqeueAfter is used. For more details, see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler

~~~~
2024-08-05T15:55:35Z    INFO    Warning: Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes reqeueuing with exponential backoff. For more details, see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler   {"controller": "glance", "controllerGroup": "glance.openstack.org", "controllerKind": "Glance", "Glance": {"name":"glance","namespace":"openstack"}, "namespace": "openstack", "name": "glance", "reconcileID": "d615ee4e-7b07-44d3-a9e8-160540927609"}
~~~~

### DIFF
--- a/controllers/glance_common.go
+++ b/controllers/glance_common.go
@@ -125,13 +125,14 @@ func ensureNAD(
 		_, err = nad.GetNADWithName(ctx, helper, netAtt, helper.GetBeforeObject().GetNamespace())
 		if err != nil {
 			if k8s_errors.IsNotFound(err) {
+				helper.GetLogger().Info(fmt.Sprintf("network-attachment-definition %s not found", netAtt))
 				conditionUpdater.Set(condition.FalseCondition(
 					condition.NetworkAttachmentsReadyCondition,
 					condition.RequestedReason,
 					condition.SeverityInfo,
 					condition.NetworkAttachmentsReadyWaitingMessage,
 					netAtt))
-				return serviceAnnotations, ctrl.Result{RequeueAfter: time.Second * 10}, fmt.Errorf("network-attachment-definition %s not found", netAtt)
+				return serviceAnnotations, glance.ResultRequeue, nil
 			}
 			conditionUpdater.Set(condition.FalseCondition(
 				condition.NetworkAttachmentsReadyCondition,
--- a/controllers/glance_common.go
+++ b/controllers/glance_common.go
@@ -19,10 +19,11 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"time"
+
 	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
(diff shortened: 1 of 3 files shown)


controllers/glance_common.go
```diff
@@ -19,10 +19,11 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"time"
```
fmount, line 22: not required as long as you use `ResultRequeue`: https://github.com/openstack-k8s-operators/glance-operator/blob/main/pkg/glance/const.go#L121

controllers/glance_common.go
```diff
@@ -117,13 +125,14 @@ func ensureNAD(
...
 			if k8s_errors.IsNotFound(err) {
+				helper.GetLogger().Info(fmt.Sprintf("network-attachment-definition %s not found", netAtt))
 				conditionUpdater.Set(condition.FalseCondition(
 					condition.NetworkAttachmentsReadyCondition,
 					condition.RequestedReason,
 					condition.SeverityInfo,
 					condition.NetworkAttachmentsReadyWaitingMessage,
 					netAtt))
-				return serviceAnnotations, ctrl.Result{RequeueAfter: time.Second * 10}, fmt.Errorf("network-attachment-definition %s not found", netAtt)
+				return serviceAnnotations, ctrl.Result{RequeueAfter: time.Second * 10}, nil
```
fmount, line 135: not sure it will raise an error but this could be:

```
return serviceAnnotations, ResultRequeue, nil
```

I don't remember why I didn't use that notation here, but it can be done in a follow up.

fmount (commented):
Just a minor comment on the usage of the existing `const` to return `RequeueAfter` but in general I'm +1 to this change.
@stuggi I can approve and follow up on that part, let me know your thoughts around that.

controllers/glance_common.go
stuggi: will update to use it

fmount (approved):
/lgtm

stuggi:
NAD annotations won't change on NAD details. so it does not make sense right now to watch them. closing in favor of https://github.com/openstack-k8s-operators/glance-operator/pull/601

PR #602, opened 2024-08-08
Fix VerifySecret so it properly requeues when necessary

abays:
We've come to realize three things recently:

1. Returning an error to the reconciler along with a `RequeueAfter` will cause the reconciler to abandon the explicit requeue and only consider the error [1].
2. The `VerifySecret` function in `lib-common` was returning an error along with a requeue, thus contributing to the problem listed in item 1. [2]
3. Also, the error `VerifySecret` was returning was _not_ a "not found" error, but a generic k8s error -- so it wouldn't be properly handled by callers looking for a "not found" error. [3]

`VerifySecret` only returns a non-empty `ctrl.Result{}` if the `Secret` is missing, so let's handle that case in a separate if-clause.

[1] https://github.com/openstack-k8s-operators/keystone-operator/pull/459
[2] https://github.com/openstack-k8s-operators/lib-common/blob/ad3edb6e67ab738cfdbf4e2e78ac403784a08ab7/modules/common/secret/secret.go#L423-L425
[3] https://github.com/openstack-k8s-operators/lib-common/blob/ad3edb6e67ab738cfdbf4e2e78ac403784a08ab7/modules/common/secret/secret.go#L425

PR #603, opened 2024-08-08
602 follow-up: actually use the new lib-common VerifySecret

abays:
I realized that I made the changes in #602 without actually pulling in the latest `lib-common` that had the `VerifySecret` function fixes.  My bad.


fmount (approved):
/lgtm

PR #609, opened 2024-08-19
Run Glance services with glance user/group when possible

fmount:
This patch improves the scc related to Glance services: it updates the code to use Glance user/group provided by kolla in both `dbsync` and `cronjob(s)`, and the same tuning is applied to `Pod/containers` that don't require special privileges.


fmount:
@abays @konan-abhi still working on moving both `httpd` and `api` away from root user: it might require a follow up patch (or commit within this patch) to make the review easier.

konan-abhi (approved):

PR #610, opened 2024-08-20
Run GlanceAPI with GlanceUID user

fmount:
When the backend is not `Cinder` (Cinder still has to be fully tested), `GlanceAPI` can reduce the permissions required for `glance-api` container, and run as `GlanceUID`/`GlanceGID`. This patch introduces `scc` for both `glanceAPI` and `httpd`.

- https://github.com/openstack-k8s-operators/glance-operator/pull/609
- https://github.com/openstack-k8s-operators/glance-operator/pull/620

Jira: https://issues.redhat.com/browse/OSPRH-9842
Jira: https://issues.redhat.com/browse/OSPRH-10149

### DIFF
--- a/pkg/glanceapi/statefulset.go
+++ b/pkg/glanceapi/statefulset.go
@@ -179,6 +179,9 @@ func StatefulSet(
 					Labels:      labels,
 				},
 				Spec: corev1.PodSpec{
+					SecurityContext: &corev1.PodSecurityContext{
+						FSGroup: &userID,
+					},
 					ServiceAccountName: instance.Spec.ServiceAccount,
 					// When using Cinder we run as privileged, but also some
 					// commands need to be run on the host using nsenter (eg:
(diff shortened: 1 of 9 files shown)


konan-abhi (approved):
Looks good to me!

pkg/glanceapi/statefulset.go
```diff
@@ -52,10 +52,11 @@ func StatefulSet(
 	annotations map[string]string,
 	privileged bool,
 ) (*appsv1.StatefulSet, error) {
-	runAsUser := int64(0)
-
+	userID := glance.GlanceUID
+	if privileged {
+		userID = int64(0)
```
fmount, line 57: This is actually used when Cinder is the backend, and the concern is mostly related to the os-brick locks here. I'm wondering if we can still run it as glance user, and let the `kolla` group access the os-brick locks shared by glance and cinder (/cc @eharney )

abays (approved):
/lgtm

fmount:
Tested with uni*eta to make sure Glance + Cinder is not a problem. I think we can merge this patch, and follow up in case of unexpected issues.

fmount:
Thank you @abays for the review.

PR #614, opened 2024-08-29
Update Image Cache sample

fmount:
We recently changed the interface name. This patch updates the example to make sure it's aligned with the current code.


konan-abhi (approved):
Looks good!

PR #619, opened 2024-09-09
Update glance samples to use OpenStackControlPlane

fmount:
We previously based our current samples on `Kind: Glance`.
However, it's easier to apply/patch the different configurations using `Kind: OpenStackControlPlane`.
This patch updates the existing samples to use the top-level control plane and make them consistent.


fmount:
@maximsava12 FYI

PR #620, opened 2024-09-09
Run httpd using kolla

fmount:
Instead of running the `httpd -DFOREGROUND` command as entry point for the `-httpd` `sidecar` container, this change moves the file copy and deployment logic to `kolla`.
This is a requirement to **not** run the container as `root` user, because `kolla` helps to apply the right permissions to the config files (and `pid`) used by the process.
The switch from root user to `GlanceUID` (already present as const) will be part of a different patch, which is already in progress (https://github.com/openstack-k8s-operators/glance-operator/pull/610) 

Jira: https://issues.redhat.com/browse/OSPRH-10040

### DIFF
--- a/pkg/glance/volumes.go
+++ b/pkg/glance/volumes.go
@@ -268,20 +268,13 @@ func GetHttpdVolumeMount() []corev1.VolumeMount {
 	return []corev1.VolumeMount{
 		{
 			Name:      "config-data",
-			MountPath: "/etc/httpd/conf/httpd.conf",
-			SubPath:   "httpd.conf",
+			MountPath: "/var/lib/config-data/default",
 			ReadOnly:  true,
 		},
 		{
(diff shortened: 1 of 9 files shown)


fmount:
@maximsava12 FYI

abays (commented):
I don't know much about kolla, but so far the changes look reasonable to me

fmount:
> I don't know much about kolla, but so far the changes look reasonable to me

Yeah, the goal is to have flexibility in setting permissions to directories, and use the builtin scripts to copy files. The next step, after this patch lands (I'm going to fix kuttl), is to add the appropriate securityContext to the httpd sidecar and run the process as Glance user (provided by kolla) instead of root.
I'm trying to do smaller and focused changes instead of a big messy patch.

abays (approved):
/lgtm

PR #622, opened 2024-09-13
[18.0.0-proposed] [validation] check glanceAPI names are valid

stuggi:
The glanceAPI controller creates StatefulSet for glanceapi to run. This adds a StatefulSet pod's label
"controller-revision-hash": "<statefulset_name>-<hash>" to the pod.
The kubernetes label is restricted under 63 char and the revision hash is an int32, 10 chars + the hyphen. Which results in a default statefulset max len of 52 chars. The statefulset name also contain the glance name and the glanceAPI type + 2 hyphens. So the max len also need to be reduced bye the length of those.

Also the name of the created rabbitmq instance must match a lowercase RFC 1123.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/562

Jira: https://issues.redhat.com/browse/OSPRH-8063

Signed-off-by: Martin Schuppert <mschuppert@redhat.com>
(cherry picked from commit 7d471da0bed691acd28c72c16b776a61c447a2dd)


fmount (approved):
/lgtm

PR #627, opened 2024-09-23
Add cronJob envTests

fmount:
This patch improves the existing `envTests` to include `cronJob` verification and update. It shows how an update in the `Glance` top level `CR` is reflected in an update of the `CronJob` schedule. In addition, the `cronJob` name now uses `instance.Name` instead of hardcoding `ServiceName` (if the `unique CR` name feature is used it will reflect the name of the global CR).


fmount:
I see a failure in fips check:

```
+---------------------------+-----------------+-----------------------------------------------+-------------------------------------------------------------------------------------+
| OPERATOR NAME             | EXECUTABLE NAME | STATUS                                        | IMAGE                                                                               |
+---------------------------+-----------------+-----------------------------------------------+-------------------------------------------------------------------------------------+
| glance-operator-container | /manager        | go binary does not contain required symbol(s) | docker.io/rdotripleomirror/glance-operator:84e3e3e940ef9219475f-1838860824643375104 |
+---------------------------+-----------------+-----------------------------------------------+-------------------------------------------------------------------------------------+
```
which is not related to kuttl, but I suspect is related to the recent bump: I probably need to rebase this patch.

fmount:
ok as mentioned in a diff thread, we need to wait for a patch that pins the go-toolset

abays:
> ok as mentioned in a diff thread, we need to wait for a patch that pins the go-toolset

I think there's a workaround in place now: https://github.com/openshift/release/pull/57159

abays (approved):
/lgtm

PR #628, opened 2024-09-23
Fix policy sample

fmount:
Policy sample missed the default API name, and it generated a failure when kustomize is executed.

Jira: https://issues.redhat.com/browse/OSPRH-10229


abays (approved):
/lgtm

PR #639, opened 2024-10-16
Automate OLM cleanup for running local operator w/ webhooks

abays:
When using make `run-with-webhook`, local versions of the operator and its webhooks are added to the cluster. If the operator was previously installed via OLM, then there might be lingering webhooks from that installation. We've previously been assuming that the user would manually remove them, but we should just get rid of them ourselves since those OLM webhooks need to be deleted anyhow for the local webhooks to function unimpeded.  We can also automatically scale down the operator's OLM deployment in such a scenario.

### DIFF
--- a/hack/configure_local_webhook.sh
+++ b/hack/configure_local_webhook.sh
@@ -142,3 +142,11 @@ webhooks:
 EOF_CAT
 
 oc apply -n openstack -f ${TMPDIR}/patch_webhook_configurations.yaml
+
+# Scale-down operator deployment replicas to zero and remove OLM webhooks
+CSV_NAME="$(oc get csv -n openstack-operators -l operators.coreos.com/glance-operator.openstack-operators -o name)"
+
+if [ -n "${CSV_NAME}" ]; then
+    oc patch "${CSV_NAME}" -n openstack-operators --type=json -p="[{'op': 'replace', 'path': '/spec/install/spec/deployments/0/spec/replicas', 'value': 0}]"
+    oc patch "${CSV_NAME}" -n openstack-operators --type=json -p="[{'op': 'replace', 'path': '/spec/webhookdefinitions', 'value': []}]"
+fi
--- a/Makefile
+++ b/Makefile
@@ -375,12 +375,11 @@ operator-lint: gowork ## Runs operator-lint
 	go vet -vettool=$(LOCALBIN)/operator-lint ./... ./api/...
 
 # Used for webhook testing
-# Please ensure the glance-controller-manager deployment and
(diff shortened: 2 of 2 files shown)


hack/configure_local_webhook.sh
```diff
@@ -142,3 +142,11 @@ webhooks:
 EOF_CAT
 
 oc apply -n openstack -f ${TMPDIR}/patch_webhook_configurations.yaml
+
+# Scale-down operator deployment replicas to zero and remove OLM webhooks
+CSV_NAME="$(oc get csv -n openstack-operators -l operators.coreos.com/glance-operator.openstack-operators -o name)"
+
+if [ -n "${CSV_NAME}" ]; then
+  oc patch "${CSV_NAME}" -n openstack-operators --type=json -p="[{'op': 'replace', 'path': '/spec/install/spec/deployments/0/spec/replicas', 'value': 0}]"
+  oc patch "${CSV_NAME}" -n openstack-operators --type=json -p="[{'op': 'replace', 'path': '/spec/webhookdefinitions', 'value': []}]"
```
fmount, line 151: As a follow up, but just out of curiosity, do we want to update [1] to reflect this approach? (`replace` w/ an empty array instead of `remove`).

[1] https://github.com/openstack-k8s-operators/dev-docs/blob/main/webhooks.md#disabling-webhooks

fmount (approved):
/lgtm

hack/configure_local_webhook.sh
abays: Yes.  In fact, I had consulted that doc while creating this and realized that this was a better way to do it. :smile:

hack/configure_local_webhook.sh
```diff
@@ -142,3 +142,11 @@ webhooks:
 EOF_CAT
 
 oc apply -n openstack -f ${TMPDIR}/patch_webhook_configurations.yaml
+
+# Scale-down operator deployment replicas to zero and remove OLM webhooks
+CSV_NAME="$(oc get csv -n openstack-operators -l operators.coreos.com/glance-operator.openstack-operators -o name)"
+
+if [ -n "${CSV_NAME}" ]; then
+  oc patch "${CSV_NAME}" -n openstack-operators --type=json -p="[{'op': 'replace', 'path': '/spec/install/spec/deployments/0/spec/replicas', 'value': 0}]"
```
fmount, line 150: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openstack-k8s-operators_glance-operator/639/pull-ci-openstack-k8s-operators-glance-operator-main-precommit-check/1846479769856643072

PR #648, opened 2024-11-05
Compute GlanceAPI storage configuration hash

fmount:
Add  `.Spec.Storage` and `.Spec.ImageCache` to the Hash based on the `enabled_backends`.
This will trigger an API refresh not only when the `enabled_backends` string changes, but also if the `Storage interface` or the `ImageCache` config (e.g. `storageClass`, `storageRequest`, `externalStorage`) is updated as part of any day2 operation.

Jira: https://issues.redhat.com/browse/OSPRH-11211


konan-abhi (approved):
Looks good, thank you!

PR #651, opened 2024-11-08
Add s3 backend samples and documentation

fmount:
Add `s3 backend` sample and provide a guide for users to go through the s3 configuration in the `Glance` `CR`.


konan-abhi (approved):
Looks sufficient to configure s3 backend, thank you!!

PR #662, opened 2024-11-27
[make][pre-commit]Check CRD schema to avoid update issues

mrkisaolamb:
The new crd-schema-check make target compares the CRD schema of the patch with the schema on the tip of main and report errors on non backward compatible changes.

This make target now also run in pre-commit both locally and in CI.

This make target uses https://github.com/openshift/crd-schema-checker to do the actual checking.

Related: [OSPRH-11833](https://issues.redhat.com//browse/OSPRH-11833)


fmount (approved):
/lgtm

PR #665, opened 2024-12-03
Add extraMounts functional tests

fmount:
This change adds a functional test for the `Glance` `extraMounts`. It ensures we're able to validate the abstraction of `corev1.VolumeSource` struct introduced in  `lib-common/storage` module.

Jira: https://issues.redhat.com/browse/OSPRH-11210


konan-abhi (approved):
looks good!!

PR #670, opened 2024-12-11
Consume Topology CR

fmount:
This patch provides more granular control over `Pod` placement and scheduling through `Topology` CR integration introduced in [1].
In particular it provides:
-  a new `API` parameter (`TopologyRef`) that allows referencing an existing `Topology` CRs in the same namespace
-  the operator logic that retrieves and processes the referenced `Topology` CR
-  an enhanced `StatefulSet` configuration that incorporates the processed Topology
-  a set of envTest to test the lifecycle (add/update/override/remove) of the resulting `StatefulSets` when a `Topology` is referenced

Jira: https://issues.redhat.com/browse/OSPRH-13256

[1] https://github.com/openstack-k8s-operators/infra-operator/pull/325


controllers/glance_controller.go
```diff
@@ -866,6 +866,13 @@ func (r *GlanceReconciler) apiDeploymentCreateOrUpdate(
 	if instance.Spec.KeystoneEndpoint == apiName {
 		apiAnnotations[glance.KeystoneEndpoint] = "true"
 	}
+
```
stuggi, line 869: can the topology of a deployment be updated after it get created, or is it immutable? We should probably watch the topology, like we do with the secrets. if it can be updated, all good, if not we have to think about what we should do in that case.

fmount: It's not immutable and I'm able to see reconciliation loops being triggered by a change in the topology.
Adding the same watching logic here would  simply be redundant, because `glanceapi_controller` (where I pass the topology reference) has already such watcher, and it triggers a reconciliation if something changes even when the `topologyRef` is a glance top level parameter (tested locally).

stuggi: ah right, have not looked closely at the api controller

controllers/glanceapi_controller.go
```diff
@@ -331,6 +336,45 @@ func (r *GlanceAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+					r.Log.Info(fmt.Sprintf("Topology %s is used by GlanceAPI CR %s", o.GetName(), cr.Name))
+					result = append(result, reconcile.Request{NamespacedName: name})
+				}
+			}
+		}
+		if len(result) > 0 {
+			return result
+		}
+		return nil
+	}
```
stuggi, line 377: wondering if we should align all those watch funcs and instead use IndexField where we have named objects to watch and use the findObjectsForSrc. I could see indexing the field is more efficient, but can not confirm.

api/v1beta1/common_types.go
```diff
@@ -108,6 +113,15 @@ type GlanceAPITemplate struct {
 	APITimeout int `json:"apiTimeout,omitempty"`
 }
 
+// TopologyRef -
+type TopologyRef struct {
+	// +kubebuilder:validation:Optional
+	// Name -
+	Name string `json:"name"`
+	// Namespace -
+	Namespace string `json:"namespace,omitempty"`
```
stuggi, line 122: I know that this is how it is used in k8s, just wondering if this provides the impression that we start supporting to spread a deployment across multiple namespaces, which we don't, at least right now. maybe we add that to the description of the parameter. could also add into the defaulting webhook to set it to the namespace of the cr.

controllers/glanceapi_controller.go
fmount: Let me try this one and I get back with an update

api/v1beta1/common_types.go
fmount: yeah it makes sense. Both comment and webhooks help here to make sure we don't risk to create false assumptions. I'm going to do that in my next PS.

fmount: Done, and both `validationCreate` and `validationUpdate` catches a different `namespace`, reporting an error with the associated message (that we do not support customizing the namespace).

controllers/glanceapi_controller.go
fmount: Hi @stuggi, to close this one, I had the opportunity to dig more into this part, and I confirm it is possible to update this code to get rid of specific topology functions [`topologyFn` and `tpFn`]. (the same thing can be extended, in a follow up, to memcached reference).
In other words I'm able to:
- index the `.spec.Topology.Name` field (we're not interested in changes for the namespace because we do not support something != from the current one)
- use the existing `predicate.GenerationChangedPredicate{}` to trigger a reconcile due to a change in the referenced Topology .Spec

Basically we reconcile in 2 cases:
1. the `topologyRef` Glance / GlanceAPI .Spec field changes (I'm updating a glanceAPI to refer to a different Topology or remove it entirely)
2. there's a change in the Topology .Spec that Glance references

I think we can resolve this thread if you're ok with this implementation.

stuggi: sounds good to me, thanks.

pkg/glanceapi/statefulset.go
```diff
@@ -290,19 +296,32 @@ func StatefulSet(
...
+					glance.ServiceName,
+				},
+				corev1.LabelHostname,
+			)
+		} else {
+			statefulset.Spec.Template.Spec.Affinity = ts.Affinity
+		}
+	}
+
 	return statefulset, err
```
olliewalsh, line 326: do we need to set topology on the other PodSpecTemplates too? e.g Jobs

api/v1beta1/conditions.go
```diff
@@ -43,4 +43,13 @@ const (
...
 	InvalidBackendErrorMessageSingle = "The GlanceAPI layout type: single can only be used in combination with File and NFS backend"
+
+	// TopologyConfigReadyInitMessage
+	TopologyConfigReadyInitMessage = "Topology config create not started"
+	// TopologyConfigReadyMessage
+	TopologyConfigReadyMessage = "Topology config create completed"
+	// TopologyConfigReadyErrorMessage
+	TopologyConfigReadyErrorMessage = "Topology config create error occurred %s"
+	// TopologyConfigReadyCondition Status=True condition which indicates a CR exists and is referenced by the Glance
+	TopologyConfigReadyCondition condition.Type = "TopologyConfigReady"
```
stuggi, line 54: maybe these conditions should go in the infra-operator PR as part of the topology api to be used by any consumer

api/v1beta1/glance_webhook.go
```diff
@@ -441,3 +457,15 @@ func GetCrMaxLengthCorrection(name string, apiType string) int {
 
 	return (defaultCrMaxLengthCorrection + len(name) + len(apiType) + 2)
 }
+
+// ValidateTopologyNamespace - returns a field.Error when Glance / GlanceAPI
+// references a Topoology deployed on a different namespace
+func ValidateTopologyNamespace(topology *TopologyRef, basePath field.Path, ns string) (*field.Error) {
```
stuggi, line 463: should this also go into the infra-operator so that we don't have to copy it to all operators?

api/v1beta1/glanceapi_types.go
```diff
@@ -115,6 +115,9 @@ type GlanceAPIStatus struct {
 	// then the controller has not processed the latest changes injected by
 	// the opentack-operator in the top-level CR (e.g. the ContainerImage)
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// LastAppliedTopology - the last applied Topology
+	LastAppliedTopology string `json:"lastAppliedTopology,omitempty"`
```
stuggi, line 120: should we also keep track of the `LastAppliedTopologyGeneration`, or is that detail of tracking to much?

controllers/glanceapi_controller.go
```diff
@@ -1427,3 +1486,94 @@ func (r *GlanceAPIReconciler) glanceAPIRefresh(
...
+	if err == nil {
+		if controllerutil.RemoveFinalizer(topology, fmt.Sprintf("%s-%s", h.GetFinalizer(), instance.APIName())) {
+			err = r.Update(ctx, topology)
+			if err != nil && !k8s_errors.IsNotFound(err) {
+				return ctrl.Result{}, err
+			}
+			util.LogForObject(h, "Removed finalizer from Topology", topology)
+		}
+	}
+	return ctrl.Result{}, err
```
stuggi, line 1578: should be the same like this, right?


```
        if err != nil && !k8s_errors.IsNotFound(err) {
		return ctrl.Result{}, err
	}

	if !k8s_errors.IsNotFound(err){
		if controllerutil.RemoveFinalizer(topology, fmt.Sprintf("%s-%s", h.GetFinalizer(), instance.APIName())) {
			err = r.Update(ctx, topology)
			if err != nil && !k8s_errors.IsNotFound(err) {
				return ctrl.Result{}, err
			}
			util.LogForObject(h, "Removed finalizer from Topology", topology)
		}
	}
	return ctrl.Result{}, nil
```

pkg/glanceapi/statefulset.go
```diff
@@ -49,6 +50,7 @@ func StatefulSet(
 	labels map[string]string,
 	annotations map[string]string,
 	privileged bool,
+	topologyOverride *topologyv1.Topology,
```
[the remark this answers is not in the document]
fmount: +1

pkg/glanceapi/statefulset.go
```diff
@@ -290,19 +296,32 @@ func StatefulSet(
...
 	}
 
+	if topologyOverride != nil {
+		ts := topologyOverride.Spec
+		if ts.TopologySpreadConstraint == nil {
+			ts.TopologySpreadConstraint = &tplg
+		} else {
+			statefulset.Spec.Template.Spec.TopologySpreadConstraints = *topologyOverride.Spec.TopologySpreadConstraint
+		}
+		if ts.Affinity == nil {
```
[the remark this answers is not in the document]
fmount: Right, otherwise we risk to not set anything. Sorry, that was a quick test I did in parallel w/ many other things. I'm fixing this and we can discuss about it further.

controllers/glanceapi_controller.go
fmount: good catch I guess is the same yes. Updating the code.

api/v1beta1/glanceapi_types.go
fmount: No we don't, `LastAppliedTopology` is basically enough to "try" to remove a finalizer if is part of the list, otherwise it's skipped. I think this is enough to make sure we free resources in case we delete the `topologyRef` key from the CR.

api/v1beta1/glance_webhook.go
fmount: This is an interesting perspective. Definitely we can move the content of the function in infra-operator and provide a basic validation that can be called from other operators. I'll try that (probably tomorrow).

api/v1beta1/conditions.go
fmount: Right, in a world where this kind of change should be spread across N operators I think best to have these conditions in infra-operator, I agree.
I'll work out this part (probably tomorrow), to make sure we can minimize the code we need in each operator to handle topologies.

pkg/glanceapi/statefulset.go
```diff
@@ -290,19 +296,32 @@ func StatefulSet(
...
+		ts := topology.Spec
+		if ts.TopologySpreadConstraint == nil {
+			ts.TopologySpreadConstraint = &tplg
+		} else {
+			statefulset.Spec.Template.Spec.TopologySpreadConstraints = *topology.Spec.TopologySpreadConstraint
+		}
+		if ts.Affinity != nil {
+			statefulset.Spec.Template.Spec.Affinity = ts.Affinity
+		}
+	} else {
```
fultonj, line 313: I like it much better to only apply the default antiaffinity if a topology was set. Merging them as in 587 could lead to unexpected behavior until the human operator figures out that the antiaffnity was getting merged.

pkg/glanceapi/statefulset.go
olliewalsh: hrmm, thinking about it further we shouldn't, at least for now....  The job pods definitely should not be considered in the same topologyconstraint rule as the statefulset or they would mess up the count for skew etc...

api/v1beta1/common_types.go
```diff
@@ -60,6 +60,11 @@ type GlanceAPITemplate struct {
 	// NodeSelector to target subset of worker nodes running this service
 	NodeSelector *map[string]string `json:"nodeSelector,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// Topology to apply the Policy defined by the associated CR referenced by
+	// name
+	Topology *TopologyRef `json:"topologyRef,omitempty"`
```
olliewalsh, line 66: Maybe name it TopologyRef to match the json field, and move the ToplogyRef type definition to lib-common

pkg/glanceapi/statefulset.go
olliewalsh: > I like it much better to only apply the default antiaffinity if a topology was set.

Not sure I understand. Are you saying that you prefer the previous behaviour?
Or is it just a typo and "only apply the default antiaffinity if a topology was *not* set"

fmount: I think it's a typo. After some testing (we started in December) we had this feeling about always having `PodAntiAffinity` created by default might causes "conflicts/contradictions" in the scheduler and prevent the wanted behavior. I think this is the best approach, and both me and John are ok with that.

api/v1beta1/common_types.go
fmount: I'll follow up on this.

controllers/glanceapi_controller.go

pkg/glanceapi/statefulset.go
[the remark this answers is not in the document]
fmount: Done

pkg/glanceapi/statefulset.go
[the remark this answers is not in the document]
fmount: Done

api/v1beta1/common_types.go
fmount: @stuggi @olliewalsh this is what I'm going to do next: I'm going to create a lib-common patch, and create `modules/common/topology.go` with the following extracted bits:
1. a `topologyRef` struct (which is basically the same you see here in Glance);
2. a `ensureTopology` function 
3. a `ensureDeleteTopology` function 

Both functions are going to have the logic defined here, but they will be generic, so we don't have to duplicate this content when we spread the pattern across the operators, but we only need the logic to check the `TopologyRef` object and call those functions. We have a similar pattern to get secrets, I assume this move will not only optimize the current code, but allow to not duplicate the same content.

Are you ok with that? Now that the `Topology` CR is available, we need to promote this from a POC to something we can land and replicate across the operators, and the above seems the required work to make it happen. Let me know your thoughts in case you do not agree (so we can discuss before I start this update).

controllers/glanceapi_controller.go
```diff
@@ -797,10 +820,44 @@ func (r *GlanceAPIReconciler) reconcileNormal(
...
+	// switches to a new Topology, remove the finalizer from the previous
+	// Topology
+	if instance.Spec.Topology == nil ||
+		(instance.Spec.Topology.Name != instance.Status.LastAppliedTopology) {
+		_, err = r.ensureDeletedTopology(ctx, instance, helper)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+	if instance.Spec.Topology != nil {
```
stuggi, line 839: should we only initialize the topology hat this point if it is != nil? with this we'd only get the topology condition reflected if it is really used. also then can move the MarkTrue inside this condition. another option would be to delete the condition from the list if the ref == nil

api/v1beta1/common_types.go
stuggi: sound good to me

fmount: First implementation is here: https://github.com/openstack-k8s-operators/lib-common/pull/594.
Let's review a few details there and I will update this patch accordingly.

controllers/glanceapi_controller.go
```diff
@@ -797,10 +830,81 @@ func (r *GlanceAPIReconciler) reconcileNormal(
...
 	}
+
 	// At this point the config is generated and the inputHash is computed
 	// we can mark the ServiceConfigReady as True and rollout the new pods
 	instance.Status.Conditions.MarkTrue(condition.ServiceConfigReadyCondition, condition.ServiceConfigReadyMessage)
 
+	//
+	// Handle Topology
+	//
+	var topology *topologyv1.Topology
```
fmount, line 841: @stuggi @olliewalsh do you think we should still wrap this `Handle Topology` logic, from L841 to L907 in a dedicated function? The advantage would be, in other operators, that we can reuse the same, common function, across components (e.g. `manilaAPI`, `manilaScheduler`, `manilaShares` controller might call the same big block). This would help to ensure consistency.
I'm looking if the `EnsureDeleteTopologyRef` logic (we have two calls) can be simplified here and reduce it to a single call (we call it both when `.Spec.TopologyRef` is `nil` and when `!= nil`, so I think I still have to push an update to simplify this further.

controllers/glanceapi_controller.go
fmount: I'm not sure about the conditions, but I can check.
When the reconciliation loop starts, we initialize the conditions to `Unknown`, and then, during the flow, if we do not mark them as True, it means that particular bit is not ready, and we should requeue another loop (waiting for resources, etc ...).
If I move `MarkTrue` within the if statement, we might need to get ourselves in the business of distribute more logic in the controller to check:
1.  the Topology is != nil, and `init` a condition to `Unknown` (and do not initialize it regardless a Topology is passed or not)
2.  remove the Condition if during an update a `topologyRef` has been deleted: this is required to avoid we still keep waiting for a Topology in case is not passed at all

Let me check this and I might update the patch to "conditionally" have Conditions around topology.

fmount: Technically, like we do for `networkAttachements` (when we compute `networkReady`), an empty list or not having this info at all in the CR is equivalent to a positive result (and we Mark the condition to true because the assume no issues are present with the network). If `TopologyRef` does not exist, in theory, we could claim that the `Topology` for the service is ready (not having anything will result in calling `DistributePods` with an antiaffinity, that is still a default Topology in a way).
I'm not sure about your thoughts on this, if this is a blocker or you're simply in disagreement with my idea or `Readiness` we can `conditionally` set a topology condition.
If you think this might complicate the current code, we can follow up to focus on that particular change.
Before creating a new commit for it, I'd like to hear from you (and @olliewalsh of course), to see if there are strong opinions on this area.

api/go.mod
```diff
@@ -70,3 +70,5 @@ require (
 // mschuppert: map to latest commit from release-4.16 tag
 // must consistent within modules and service operators
 replace github.com/openshift/api => github.com/openshift/api v0.0.0-20240830023148-b7d0481c9094 //allow-merging
+
+replace github.com/openstack-k8s-operators/lib-common/modules/common => github.com/fmount/lib-common/modules/common v0.0.0-20250121110137-f3ec1120ae76 //allow-merging
```
stuggi, line 74: just a side note, adding `allow-merging` to this kind of replace is dangerous as the pre-commit replace check won't report it and the PR could be merged by the automation

controllers/glanceapi_controller.go
stuggi: If I read it correct, it should be generic to all consumers? if the list of parameters we'd have to pass in is not super big, also we could use a struct type for this for passing in the details. this would allow us to extend it and not have to change the func definition. condition handling would be outside the func. so yes I think its worth looking into it.

api/go.mod
fmount: I can remove it, sure. I thought the `do-not-merge` label was enough to prevent the automation merging this patch automatically. I'll remove it in my next iteration.

stuggi: yes sure, the do not merge is enough. its not critical, just wanted to mention that basically the `allow-merging` will override our pre-commit protection check.

controllers/glanceapi_controller.go
fmount: @stuggi I checked this part as well and added a `TopologyReadyCondition` outside of the default group. By doing this only when a topology is referenced this condition is evaluated.

api/go.mod
fmount: Removed

controllers/glanceapi_controller.go
```diff
@@ -1427,3 +1498,70 @@ func (r *GlanceAPIReconciler) glanceAPIRefresh(
 	}
 	return nil
 }
+
+// ensureGlanceAPITopology -
+func (r *GlanceAPIReconciler) ensureGlanceAPITopology(
+	ctx context.Context,
+	helper *helper.Helper,
+	tpRef *topologyRef.Topology,
```
fmount, line 1506: @stuggi this function wraps the whole logic in the `glanceapi_controller`. It takes three input parameters:
1. a topologyRef, which is the struct that eventually contains both Name and Namespace params
2. lastAppliedTopology (it comes, if set, from .Status, and used to handle the `EnsureDeleteTopologyRef` flow)
3. finalize: because glanceAPIs have arbitrary names and is a list, we pass the API name so we know which Topology is used by which API

I'm not sure wrapping those parameters in an additional struct results in overcomplicating the understanding of the code, as I don't think we are going to grow those parameters. If `topologyRef` grows in lib-common, we automatically get new values there, but I wanted to avoid mixing `TopologyRef`, `lastAppliedTopology` and `finalizer` within the same group.
What do you think? Any opinion about the reason we should group them on a struct? I think it's good to explicitly add some parameters in the struct, and what is referenced is already hidden by `topologyRef`, but I might be missing something (/cc @olliewalsh)

api/v1beta1/glance_types.go
```diff
@@ -72,6 +73,11 @@ type GlanceSpecCore struct {
 	// NodeSelector to target subset of worker nodes running this service
 	NodeSelector *map[string]string `json:"nodeSelector,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// TopologyRef to apply the Topology defined by the associated CR referenced
+	// by name
+	TopologyRef *topology.TopoRef `json:"topologyRef,omitempty"`
```
ASBishop, line 79: I wondered why we need a TopologyRef in the GlanceCoreSpec structure when there's also one in the GlanceAPITemplate structure in common_types.go. That had me looking at the NodeSelector field (asking myself the same question; why is it needed in this structure when there's one in the GlanceAPITemplate?).

From what I can tell, the NodeSelector field in this structure is only referenced by the pkg code that handles the dbsync and cronjob stuff. I guess dbsync and cronjobs can't (re)use the NodeSelector from a GlanceAPI? That's a separate question unrelated to this PR, but it has me asking whether we need a TopologyRef in this structure. Will it be referenced anywhere, for example in some future update to the dbsync or cronjob code?

pkg/glanceapi/statefulset.go
```diff
@@ -290,19 +296,32 @@ func StatefulSet(
...
-		corev1.LabelHostname,
-	)
 	if instance.Spec.NodeSelector != nil {
 		statefulset.Spec.Template.Spec.NodeSelector = *instance.Spec.NodeSelector
 	}
 
+	if topology != nil {
+		ts := topology.Spec
+		if ts.TopologySpreadConstraints == nil {
+			ts.TopologySpreadConstraints = &tplg
```
ASBishop, line 306: I don't understand what this is doing.
-  Is the reference to "&tplg" just to an empty array per L80?
-  How does this affect the statefulset?

fmount: Long time I don't check this part. I think the main problem I had was to make sure we do not dereference something that is not present in the Topology CR to avoid segfaults. This was the high level idea, but, considering we don't have an Affinity override that is applied regardless of the TopologySpreadConstraint, I think I can really simplify this part further.
Let me push an update to make sure this is simple enough and easy to read.

fmount: Done, my last update simplifies this part and improves readability.

pkg/glanceapi/statefulset.go
```diff
@@ -290,19 +293,33 @@ func StatefulSet(
...
+			statefulset.Spec.Template.Spec.TopologySpreadConstraints = *topology.Spec.TopologySpreadConstraints
+		}
+		// Process Affinity if defined in the referenced Topology
+		if ts.Affinity != nil {
+			statefulset.Spec.Template.Spec.Affinity = ts.Affinity
+		}
+	} else {
+		// If possible two pods of the same service should not
+		// run on the same worker node. If this is not possible
+		// the get still created on the same worker node.
```
stuggi, line 314: do we have to explicite set the `statefulset.Spec.Template.Spec.TopologySpreadConstraints` to be empty in the case someone removed the ref back to default, since we update it using `CreateOrPatch`?

test/functional/suite_test.go
```diff
@@ -47,6 +47,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
```
stuggi, line 48: probably should add some tests for adding/removing the topologyref to see that the statefulset gets updated properly when move from default to toporef and back?

controllers/glanceapi_controller.go
```diff
@@ -86,6 +88,7 @@ type GlanceAPIReconciler struct {
 // +kubebuilder:rbac:groups=keystone.openstack.org,resources=keystoneendpoints,verbs=get;list;watch;create;update;patch;delete;
 // +kubebuilder:rbac:groups=k8s.cni.cncf.io,resources=network-attachment-definitions,verbs=get;list;watch
 // +kubebuilder:rbac:groups=memcached.openstack.org,resources=memcacheds,verbs=get;list;watch;
+// +kubebuilder:rbac:groups=topology.openstack.org,resources=topologies,verbs=get;list;watch;update;patch
```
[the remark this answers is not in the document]
fmount: yeah, I think this is a leftover related to the previous iterations (before moving the Topology logic to lib-common). Let me update this and double check we have no issues. Thanks for the suggestion!

controllers/glanceapi_controller.go
```diff
@@ -797,10 +833,49 @@ func (r *GlanceAPIReconciler) reconcileNormal(
...
+	)
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.TopologyReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			condition.TopologyReadyErrorMessage,
+			err.Error()))
+		r.Log.Info("GlanceAPI config is waiting for Topology requirements, requeueing...")
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
fmount: yes, not against that, I'm going to add it in my next iteration.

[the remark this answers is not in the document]

controllers/glanceapi_controller.go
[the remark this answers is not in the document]

pkg/glanceapi/statefulset.go
fmount: No, if you remove the reference, a new `StatefulSet` is rolled out (because the `.Spec` changed), so we don't need to explicitly set that parameter to an empty `TopologySpreadConstraints` because it is already `nil` and `omitted` from the resulting `StatefulSet`.
The behavior is exactly the same we see with `NodeSelector`.

stuggi: cool, sounds good!

test/functional/suite_test.go
fmount: Yes, the current tests only prove that we have no regression and 100% backward compatibility we the previous CR versions and the behavior is preserved (we don't get any unexpected result). I agree that adding / removing a `topologyRef` to actually test the new code introduced both here and in lib-common would be a nice improvement here. Let me experiment with that and I will update this patch.

controllers/glanceapi_controller.go
```diff
@@ -86,6 +88,8 @@ type GlanceAPIReconciler struct {
 // +kubebuilder:rbac:groups=keystone.openstack.org,resources=keystoneendpoints,verbs=get;list;watch;create;update;patch;delete;
 // +kubebuilder:rbac:groups=k8s.cni.cncf.io,resources=network-attachment-definitions,verbs=get;list;watch
 // +kubebuilder:rbac:groups=memcached.openstack.org,resources=memcacheds,verbs=get;list;watch;
+// +kubebuilder:rbac:groups=topology.openstack.org,resources=topologies,verbs=get;list;watch
```
fmount, line 91: @stuggi I think I have an issue with this: in theory your suggestion is correct since we `get/list/watch` topologies, and we only update the CR by adding/removing `finalizers`.
However, in my testing, when I start from a deployment without any topology, and then I add the reference, I get:
```
2025-01-29T14:06:26Z    ERROR   Reconciler error        {"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "GlanceAPI": {"name":"glance-default-external","namespace":"openstack"}, "namespace": "openstack", "name": "glance-default-external", "reconcileID": "0a7f579d-3e12-4715-a714-23ce9aa817f3", "error": "waiting for Topology requirements: topologies.topology.openstack.org \"glance-default-spread\" is forbidden: User \"system:serviceaccount:openstack-operators:glance-operator-controller-manager\" cannot update resource \"topologies\" in API group \"topology.openstack.org\" in the namespace \"openstack\": RBAC: clusterrole.rbac.authorization.k8s.io \"glance-operator-proxy-role\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.6/pkg/internal/controller/controller.go:329
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.6/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
```

fmount: This happens because we call `.Update` from [1] and [2], therefore we need that `verb` in glance.
Topologies are not created by any "entity" atm, they can be manually generated, and they are not reconciled with a controller, so we don't have a `createOrPatch` for that resource.
We can think about adding it as some point, and see if we can apply your suggestion, but right now I would stick with `get/list/watch/update` on the Topology CRD.


[1] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/topology/topology.go#L57
[2] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/topology/topology.go#L91

stuggi: right there we have the updated. missed that. that's fine. lets keep what you have and we can revisit it.

test/functional/suite_test.go
fmount: Done

stuggi:
thanks @fmount , looks good to me. maybe we could squash some of the commits, but its really not critical. lets see if @olliewalsh can have a look today, too.

ASBishop:
Everything lgtm, so I can approve this once the other reviewers are satisfied.

fmount:
@stuggi @ASBishop thanks for the multiple reviews. I simply pushed an update with an additional envTest to trigger webhooks and reject an invalid reference. @olliewalsh this patch is ready to land, let me know if we can move forward.

test/functional/base_test.go
```diff
@@ -345,3 +369,38 @@ func GetExtraMounts() []map[string]interface{} {
...
+	// Build the topology Spec
+	topologySpec := map[string]interface{}{
+		"topologySpreadConstraints": []map[string]interface{}{
+			{
+				"maxSkew":           1,
+				"topologyKey":       corev1.LabelHostname,
+				"whenUnsatisfiable": "ScheduleAnyway",
+				"labelSelector": map[string]interface{}{
+					"matchLabels": map[string]interface{}{
+						"service": glanceTest.GlanceService.Name,
```
olliewalsh, line 385: Looks like the cache job uses a more specific labelSelector https://github.com/openstack-k8s-operators/glance-operator/blob/be560a599e2dce0d0eb0bd57d794a9172a862c8e/pkg/glanceapi/funcs.go#L55 that includes the APIType. Which is correct?

fmount:
@olliewalsh FYI I still think I'd like to incorporate the default `LabelSelector` introduced in [1] in a follow up PR.
For the other operators it would be ok to include them in a single PR, but for glance I'd like to keep the focus on what we tested/validated.

[1] https://github.com/openstack-k8s-operators/lib-common/pull/597

test/functional/base_test.go
olliewalsh: If the cache job style selector is correct that would be best to use that in the test

fmount: Yes, because that's a `RequiredDuringSchedulingIgnoredDuringExecution` `PodAffinityTerm` because the cache job must be colocated with the API that refers to. This will go away with 19, but required for the imageCache.

fmount: > style selector is correct it would be best to use that in the test

I won't to avoid confusion. `ImageCache` follows the `GlanceAPI`, and it uses a different key (`glance.GlanceAPIName`) the affinity for that particular job can't be passed as an override. In addition, I think that jobs/cronJobs should be out of scope for topology.
At least for Storage operators, jobs only last a few seconds, and they need eventually db connection. No reason to overcomplicate the logic unless there's a particular reason.

olliewalsh: Yes, but isn't the same true for any constraints? E.g the pods for internal API service need to be spread from each other, the pods for the external API need to spread from each other, so the glanceAPI label should be used. The service label would group the pods from both into a single set IIUC

fmount: even though you have multiple internal/external API (type: split, but it's the same logical API), you want an balanced distribution like it was a single application (`service: glance`). What you're saying is an override at API level when we have a particular case (e.g. nodeAffinity) of a specific set of API that shouldn't follow the default topology rule, valid at service level. I might not understand the entire point here, but [1] is the theory behind my tests and the behavior we need for ToplogySpreadConstraints (for the rest we have affinity, and in that case yes, we use GlanceAPI as LabelSelector).
Anyway I think we're going out of scope with this, I'd like to clarify two points:
1. `GetGlanceAPIPodAffinity` is only used for `ImageCache` cronjob (entirely out of scope here, as they have a unique behavior for the Glance use case)
2. This function in the `envTests` is just a sample: if the key selector follows the service or the API is not relevant for the test itself. I understand you might have a different idea, and we can change it in a follow up

Not sure I'm really understanding your point, hope the above ^^ clarifies a bit the goal of the test. I can put a comment there to explain it better if you think it might generate confusion for the reader (/me noted for a follow up patch).

[1] https://gist.github.com/fmount/2c5e7b99d3e1bcc1a2afdb619c7ad9d6

olliewalsh:
> @olliewalsh FYI I still think I'd like to incorporate the default `LabelSelector` introduced in [1] in a follow up PR. For the other operators it would be ok to include them in a single PR, but for glance I'd like to keep the focus on what we tested/validated so far.
> 
> [1] [openstack-k8s-operators/lib-common#597](https://github.com/openstack-k8s-operators/lib-common/pull/597)

It shouldn't affect any existing testing/validation. The would not have used a nil LabelSelector as it means "match no pods" in this context.

fmount:
> > @olliewalsh FYI I still think I'd like to incorporate the default `LabelSelector` introduced in [1] in a follow up PR. For the other operators it would be ok to include them in a single PR, but for glance I'd like to keep the focus on what we tested/validated so far.
> > [1] [openstack-k8s-operators/lib-common#597](https://github.com/openstack-k8s-operators/lib-common/pull/597)
> 
> It shouldn't affect any existing testing/validation as they would not have used a nil LabelSelector as it means "match no pods" in this context.
> 
> It's an API change in lib-common so would like to land it before any of the operators depend on the function

As I mentioned, it's glance-operator, not every other operator!
I'm ok to land your patch first and rebase everything on top of it. I will incorporate the default `LabelSelector` change on my current patches if this is a blocker.

test/functional/base_test.go
olliewalsh: ack that it's just a test but comment says it's an opinionated sample which suggests it should be done this way.... which is wrong I suspect.

E.g when treating all 6 pods as the same set it could result in 
```
+--------+      +--------+      +--------+
|        |      |        |      |        |
| ZONE A |      | ZONE B |      | ZONE C |
|        |      |        |      |        |
+--------+      +--------+      +--------+
 |_ api-ext-0    |_ api-int-0    |_ api-ext-2
 |_ api-ext-1    |_ api-int-1    |_ api-int-2
```

olliewalsh:
> As I mentioned, it's glance-operator, not every other operator! I'm ok to land your patch first and rebase everything on top of it. I will incorporate the default `LabelSelector` change on my current patches if this is a blocker.

Lets follow-up tomorrow. It's a relatively minor change, testing it on top of this PR in https://github.com/olliewalsh/glance-operator/commit/8386bff9a356bb8ff88a7cfa0b274b400c26b3b5. Decided what the default should be might not be straightforward though...

test/functional/base_test.go
fmount: If you have multiple nodes per zone that might happen, but using `glanceAPI` as key is not sufficient to ensure the following:

```
+--------+      +--------+      +--------+
|        |      |        |      |        |
| ZONE A |      | ZONE B |      | ZONE C |
|        |      |        |      |        |
+--------+      +--------+      +--------+
 |_ api-ext-0    |_ api-ext-1    |_ api-ext-2
 |_ api-int-0    |_ api-int-1    |_ api-int-2
```

If you use glanceAPI as key I suspect that Pods with the **same labelselector** are spread across zones, which is not the wanted behavior.
For example if I have `glanceAPI=api0`, which is made by `api0-ext` and `api0-int`, a `TopologySpreadConstraint` will simply schedule a set of `Pods` of the **same** API on **different zones**, and this is why spreading as a `service` instead of `api` is a viable solution. You can achieve isolation for a set of Pods of the same API using `affinity`: I can have an affinity based on the `APIName`, and if I properly merge it with TopologySpreadConstraint I can reach that particular result. All of this is not simply something we can automate by adding `cluster-level` constraints, because each service has its own characteristics.
Hopefully with 19 we're removing the api split and this should simplify the scenario a lot.

fmount:
> > As I mentioned, it's glance-operator, not every other operator! I'm ok to land your patch first and rebase everything on top of it. I will incorporate the default `LabelSelector` change on my current patches if this is a blocker.
> 
> Lets follow-up tomorrow. It's a relatively minor change, testing it on top of this PR in [olliewalsh@8386bff](https://github.com/olliewalsh/glance-operator/commit/8386bff9a356bb8ff88a7cfa0b274b400c26b3b5). Deciding what the default should be might not be straightforward though...

@olliewalsh similarly to what we did for `DistributePods` [1], while looking at [2], I think we should make the following function:

```golang
// GetGlanceAPILabelSelector - Returns a metav1.LabelSelector for a given GlanceAPI
func GetGlanceAPILabelSelector(instance *glancev1.GlanceAPI) metav1.LabelSelector {
	return metav1.LabelSelector{
		MatchExpressions: []metav1.LabelSelectorRequirement{
			{
				Key:      glance.GlanceAPIName,
				Operator: metav1.LabelSelectorOpIn,
				Values: []string{
					fmt.Sprintf("%s-%s-%s", glance.ServiceName, instance.APIName(), instance.Spec.APIType),
				},
			},
		},
	}
}
``` 

a generic one, that can be called from lib-common. This way we can avoid duplicating it across the operators, and we can provide 
```
common.AppSelector, []string{glance.ServiceName`}
```
as input for that function.
You could then call it with something like `topology.GetDefaultTopologySelector(common.AppSelector, []string{glance.ServiceName`}) and pass it down to `topology.EnsureTopologyRef(`.

Do you want to update https://github.com/openstack-k8s-operators/lib-common/pull/597 to add this new function?

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/pkg/glanceapi/statefulset.go#L296
[2] https://github.com/olliewalsh/glance-operator/commit/8386bff9a356bb8ff88a7cfa0b274b400c26b3b5

controllers/glanceapi_controller.go
```diff
@@ -1509,6 +1510,7 @@ func (r *GlanceAPIReconciler) glanceAPIRefresh(
 // newly referenced topology object
 func (r *GlanceAPIReconciler) ensureGlanceAPITopology(
 	ctx context.Context,
+	instance *glancev1.GlanceAPI,
```
fmount, line 1513: we shouldn't require instance here.

controllers/glanceapi_controller.go
```diff
@@ -1545,11 +1547,13 @@ func (r *GlanceAPIReconciler) ensureGlanceAPITopology(
 			tpRef.Namespace = helper.GetBeforeObject().GetNamespace()
 		}
 		// Retrieve the referenced Topology
+		defaultLabelSelector := glanceapi.GetGlanceAPILabelSelector(instance)
```
fmount, line 1550: This should directly call `labels.GetSingleLabelSelector(` and pass down a struct to this call. This is important because if you look at other PRs, where we have multiple Components (manilaAPI, manilaScheduler, manilaShare) we run this same code multiple times for different `instances`.

pkg/glanceapi/funcs.go
```diff
@@ -43,26 +45,25 @@ func GetGlanceEndpoints(apiType string) map[service.Endpoint]endpoint.Data {
 	return glanceEndpoints
 }
 
+// GetGlanceAPILabelSelector - Returns a metav1.LabelSelector for a given GlanceAPI
```
fmount, line 48: we do not require this wrapper, or at least we should pass down `instance`.

fmount: Done

controllers/glanceapi_controller.go

test/functional/base_test.go
olliewalsh: > For example if I have glanceAPI=api0

Not sure I understand. You wouldn't have this, the glanceAPI label includes the api type.

fmount: Done

fmount: Resolving the thread, but we can move the conversation elsewhere if you would like to dig more and brainstorm w/ me more on this! Thank you for the feedback so far!

olliewalsh: Was mostly just asking which is the correct labelselector to use as it's significant for the patches that set the default. For the tests, meh it's just a nit, maybe a comment wouldn't hurt for our future-selves wondering why it is different

olliewalsh (approved):
/lgtm

PR #686, opened 2025-02-10
Update topology reference from lib-common to infra-operator

fmount:
This patch updates the `topology` common function references to `infra-operator`, which is now the only source of truth when we import `Topology` related functions.

Depends-On: https://github.com/openstack-k8s-operators/infra-operator/pull/342


stuggi (approved):
/lgtm

PR #699, opened 2025-03-07
[functional] verify LastAppliedTopology != nil before checking name

fmount:
This patch fixes the way we verify the status of LastAppliedTopology field. In particular, .Name should be dereferenced only when we know that .LastAppliedTopology is not nil.

Related: [OSPRH-14494](https://issues.redhat.com//browse/OSPRH-14494)


stuggi (approved):
/lgtm

PR #702, opened 2025-03-11
Fix cinder dependency when unique_pod_names is true

fmount:
When `unique_pod_names` parameter, injected by `openstack-operator` is `true`, the `Cinder` CR is created with a pseudorandom name, and it's not predictable by `glance-operator` when a cinder backend is added. Other than making sure that `Cinder` exists, we're not doing any other check in glance (it's a decision), hence we can move from `r.Get` to `r.List` to ensure that at least a `Cinder CR` exists in the same `namespace` where `Glance` is deployed.

Jira: https://issues.redhat.com/browse/OSPRH-14637
Jira: https://issues.redhat.com/browse/OSPCIX-725

### DIFF
--- a/controllers/glanceapi_controller.go
+++ b/controllers/glanceapi_controller.go
@@ -695,21 +695,11 @@ func (r *GlanceAPIReconciler) reconcileNormal(
 		backendToken := strings.SplitN(availableBackends[i], ":", 2)
 		switch {
 		case backendToken[1] == "cinder":
-			cinder := &cinderv1.Cinder{}
-			err := r.Get(ctx, types.NamespacedName{Namespace: instance.Namespace, Name: glance.CinderName}, cinder)
+			cinderList := &cinderv1.CinderList{}
+			err := r.List(ctx, cinderList, client.InNamespace(instance.Namespace))
 			if err != nil {
-				if k8s_errors.IsNotFound(err) {
-					// Request object not found, GlanceAPI can't be executed with this config
-					r.Log.Info("Cinder resource not found. Waiting for it to be deployed")
-					instance.Status.Conditions.Set(condition.FalseCondition(
-						glancev1.CinderCondition,
-						condition.RequestedReason,
-						condition.SeverityInfo,
-						glancev1.CinderInitMessage),
-					)
-					return glance.ResultRequeue, nil
-				}
-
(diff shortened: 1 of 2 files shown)


konan-abhi (approved):

controllers/glanceapi_controller.go
```diff
@@ -695,8 +695,8 @@ func (r *GlanceAPIReconciler) reconcileNormal(
 		backendToken := strings.SplitN(availableBackends[i], ":", 2)
 		switch {
 		case backendToken[1] == "cinder":
-			cinder := &cinderv1.Cinder{}
-			err := r.Get(ctx, types.NamespacedName{Namespace: instance.Namespace, Name: glance.CinderName}, cinder)
+			cinderList := &cinderv1.CinderList{}
+			err := r.List(ctx, cinderList, client.InNamespace(instance.Namespace))
```
fmount, line 699: I still need to update this code. Marking as draft

ASBishop: I see where you're going with this! Look for all cinders in this namespace, and there should be only one of them: the one with the unique pod names.

fmount: Correct. The problem is that `uniquePodNames` parameter exists in the `openstack-operator` Glance (and Cinder) sections [1], which means that we're not aware, in the services operators, about the difference between standard names (`ServiceName`) and pseudo random CR names, which is ok I guess, we don't have to take strong assumptions about "names".
Because we do not support multiple Cinder CRs in the same namespace, we know, that only one will be deployed by the openstack-operator, and its existence represents the Glance dependency.
If there isn't a Cinder CR, we should block the deploy, hence the condition I set as an `and` between `err` (the result of listing cinder CRs) and the length of the `.Items` array.
We could, in the future, discuss about introducing a `Cinder` name parameter in the Glance CR, but the scope of such change might be bigger that what we're trying to fix in the scope of this release.

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/main/apis/core/v1beta1/openstackcontrolplane_types.go#L434

fmount:
ok I think this change is in an acceptable state to fix the current CIX. @ASBishop @konan-abhi because you both looked at this patch, would you like to take a look and see if the approach works for you?

ASBishop (approved):
/lgtm

PR #703, opened 2025-03-11
Add envTests for Cinder backend dependency

fmount:
As a follow up of [1], this patch introduces functional tests to check how the `Cinder condition` is updated when `Cinder` is added as a `Glance` backend. The test shows how the condition (defined in the glance-operator) transitions from `False` to `True` as long as a (default) `Cinder CR` (with an arbitrary name) is deployed.

Related: https://issues.redhat.com/browse/OSPRH-14637

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/702


konan-abhi (approved):
Looks good, thank you!!

PR #709, opened 2025-03-25
Add cinder-multistore control plane example

fmount:
This patch reworks the multistore related samples and introduces an example based on cinder multistore (backed by nfs and lvm-iscsi).

Jira: https://issues.redhat.com/browse/OSPRH-15130


config/samples/backends/multistore/cinder/glance.yaml
```diff
@@ -0,0 +1,44 @@
...
+  glance:
+    template:
+      customServiceConfig: |
+        [DEFAULT]
+        enabled_backends = lvm:cinder,nfs:cinder
+        debug = true
+        [glance_store]
+        default_backend = lvm
+        [lvm]
+        rootwrap_config = /etc/glance/rootwrap.conf
```
fmount, line 18: @konan-abhi @rajathere I'm wondering if, based on [1], we have a way to not repeat common parameters.

[1] https://etherpad.opendev.org/p/apr2024-ptg-glance#L187

rajathere: That was surely proposed (inspired from cinder) but not actually implemented. I could not find any spec or implementation related to it so unfortunately we don't have that support yet.

konan-abhi (approved):
Looks good, thank you!

PR #712, opened 2025-03-28
Move cinder multistore backend examples to ontap iscsi and nfs

fmount:


konan-abhi (approved):
Looks good, thank you!!

PR #718, opened 2025-04-14
Fix GlanceAPI SecurityContext with Cinder backend

fmount:
When `Cinder` is selected as a `backend`, we need a `privileged` container to access the underlying resources required by `os-brick`. This patch aligns the security context to deploy a privileged container (as before) with `Cinder` as a `backend`. We missed this in the previous patch series.


konan-abhi (approved):
Looks good!

PR #728, opened 2025-04-30
Add the ability to deploy httpd+proxypass or wsgi mode

fmount:
The goal of this patch is to handle `minor updates` when we move from an older image that does not support `wsgi` to a newer image that must be deployed using the `wsgi` approach.
This patch adds the ability to drive the `Glance` deployment through an annotation that determines the deployment model for a particular image.


pkg/glance/const.go
```diff
@@ -108,6 +108,8 @@ const (
 	ShortDuration = time.Duration(5) * time.Second
 	// NormalDuration -
 	NormalDuration = time.Duration(10) * time.Second
+	// GlanceWSGILabel -
+	GlanceWSGILabel = "wsgi"
```
stuggi, line 112: two things in this. 
* if we add the  const in the api module, we could use it in the openstack-op which imports it.
* iirc we discussed that a user could override it when adding the annotation on the ctlplane cr. for this, this should have a uniq name for glance, as in future other services might have such annotations, too. Maybe something like using the `<group>/<key>`, like `glance.openstack.org/wsgi`?

controllers/glance_common.go
```diff
@@ -300,3 +301,26 @@ func GetServiceLabels(
 		common.OwnerSelector:     instance.Name,
 	}
 }
+
+// GetWSGIAnnotation - returns in the form of a boolean the value associated to
+// the "wsgi" annotation. It determines whether Glance will be deployed in WSGI
+// or ProxyPass mode.
+func GetWSGIAnnotation(
```
stuggi, line 308: might be good to add a generic one in https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/annotations/network.go , which has parameters for
* the annotations map
* the key to check
* the expected default when not found

templates/common/config/glance-api-config.json
```diff
@@ -0,0 +1,78 @@
+{
```
stuggi, line 1: wondering why this kolla config is in common and linked to `glanceapi/config` , while the `glance-httpd-config.json` is in `glanceapi/config`?

pkg/glance/const.go
fmount: > two things in this.
> 
>     * if we add the  const in the api module, we could use it in the openstack-op which imports it.
It makes sense, I'll do that so we can consume it from openstack-operator. Thank you for this, I agree it is a better experience.
> 
>     * iirc we discussed that a user could override it when adding the annotation on the ctlplane cr. for this, this should have a uniq name for glance, as in future other services might have such annotations, too. Maybe something like using the `<group>/<key>`, like `glance.openstack.org/wsgi`?

Ack, atm in the `openstack-operator` side the annotation is only set to the Glance CR in the `glance.go`, but I agree if we want to provide a set of annotations that get propagated to the underlying services, we want to use the group name to make them unique. I'll follow up on this.

templates/common/config/glance-api-config.json
fmount: I think this comes from historical reasons because we went through a refactoring. long term, but out of scope for this patch, I want to only have a `glance/config` that is in "common" by default and used for all the GlanceAPI.

controllers/glance_common.go
fmount: I was thinking about this at the beginning, it might be a good idea. Let me draft something and I will come up with an improved approach for this.

fmount: Done https://github.com/openstack-k8s-operators/lib-common/pull/628

pkg/glance/const.go
fmount: I think I did this. 
I added this field (as a static string though) in the `glance-operator` `API`, so we can consume it when we import `glancev1` in `openstack-operator`.

dprince:
Here is a potential openstack-operator idea to support a minor updates mechanism for this new glance deployment type https://github.com/openstack-k8s-operators/openstack-operator/pull/1439

dprince:
Going to land this. Then we can promote it into the openstack-operator PR and verify minor updates are working correctly

dprince (approved):

PR #730, opened 2025-05-05
Enable MTLS memcached auth

lmiccini:
This commit allows operators to use mtls as an authentication method against Memcached.
Glance controller will detect the presence of a purposely-created mtls certificate (authCertSecret) and use this to configure the [keystone_authtoken] section accordingly.
Additional volumes/volumemounts will be appended to each pod.

Note that this commit switches from MemcachedServersWithInet to MemcachedServers since keystone-middleware now uses pymemcache when tls=true and there is no need to use "[]" to enclose the list of memcached servers even for ipv6.


fultonj:
Looks reasonable to me.

PR #738, opened 2025-05-26
Do not update status when reconciler exits due to panic

ratailor:
Currently we use deferred call to always update the status field when Reconcile call exits, which seems correct for any other error except Reconciler exits due to panic.

This change adds a call to recover function and try to handle panic and log error.

Closes: [OSPRH-16934](https://issues.redhat.com//browse/OSPRH-16934)

### DIFF
--- a/controllers/glance_controller.go
+++ b/controllers/glance_controller.go
@@ -140,6 +140,11 @@ func (r *GlanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 	// Always patch the instance status when exiting this function so we can
 	// persist any changes.
 	defer func() {
+		// Don't update the status, if Reconciler Panics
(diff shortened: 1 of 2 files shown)


fmount:
I'm not sure we have a way to automatically test this patch. The current testing can ensure that we do not introduce regressions, which is fine, but I'm wondering if you recently hit this issue and if you have an evidence of some manual test that fixes it.
In theory, if the reconciliation breaks, or hit a panic, I agree that we shouldn't update the `.Status` of the CR, because we risk to inject wrong data in case something else reads this resource, but in practice I'm wondering why panic happens because there might be a bug we want to fix (I assume we shouldn't panic in the first place!).
@ratailor do you have any insights about this issue?

fmount (approved):
/lgtm

PR #739, opened 2025-05-27
Enable notifications for Glance

fmount:
This patch adds the logic of processing a `transportURL` `CR/secret`. It has been a gap so far compared to the other operators. However, because `Glance` do not need `RabbitMq` to properly work, if a `rabbitmqClusterName` is not passed, the deployment moves forward in any case.
`RabbitMQ` is only used so far for notification purposes, and based on this parameter `oslo_messaging_notifications` is configured.

Jira: https://issues.redhat.com/browse/OSPRH-15845

### DIFF
--- a/config/samples/copy_image/README.md
+++ b/config/samples/copy_image/README.md
@@ -14,7 +14,8 @@ than the default, for example:
 "copy_image": "'public':%(visibility)s"
 ```
 
-For example, use the following configuration to enable `copy-imge` import method:
+For example, use the following configuration to enable `copy-image` import
+method:
 
 ```
 ...
--- a/config/samples/notifications/README.md
+++ b/config/samples/notifications/README.md
@@ -1,0 +1,110 @@
+# Glance Notifications
+
+The OpenStack Image Service (Glance), like other core OpenStack services, can
+generate notifications for various events that occur throughout the image
+lifecycle. These notifications provide valuable telemetry data for auditing,
+troubleshooting, monitoring operations, and integration with other services
+such as Ceilometer for metrics collection and processing.
+While Glance operates independently without requiring RabbitMQ for core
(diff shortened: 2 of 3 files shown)


config/samples/layout/base/glance_v1beta1_glance.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: glance
 spec:
   serviceUser: glance
+  rabbitmqClusterName: rabbitmq
```
fmount, line 7: maybe I can remove this and check that we have no regressions with the current kuttl tests when this parameter is not passed.

fmount:
Let's get some CI results on this one

templates/common/config/00-config.conf
```diff
@@ -57,8 +57,12 @@ project_name=service
 password = {{ .ServicePassword }}
 
 [oslo_messaging_notifications]
-#driver=messagingv2
+{{ if (index . "TransportURL") -}}
+driver=messagingv2
+transport_url = {{ .TransportURL }}
+{{ else -}}
 driver=noop
+{{ end -}}
```
konan-abhi, line 65: Should we also consider retry value which is by default -1

# The maximum number of attempts to re-send a notification message which failed
# to be delivered due to a recoverable error. 0 - No retry, -1 - indefinite
# (integer value)
#retry = -1

https://github.com/openstack/oslo.messaging/blob/master/oslo_messaging/notify/notifier.py#L44

fmount: yeah, it would makes sense. I wouldn't add it by default there because we're currently relying on  a shared rabbit. But I noted this as a follow up that we might have in this area.

konan-abhi (approved):
Looks good and neatly explained in docs as well!!

PR #747, opened 2025-06-16
Update glance-operator docs

fmount:
This patch introduces a `README` in the main `docs` folder and references the relevant parts of the repo.
While there's still some doc that should be updated, the immediate goal is to provide a meaningful outline that references the design decisions and the configuration aspect of the `Glance` features provided by the operator.


konan-abhi (approved):
Looks perfect, thank you!!

PR #751, opened 2025-06-24
Remove unnecessary rootwrap_config

fultonj:
Privileged operations should be managed by securityContext provided by k8s.

### DIFF
--- a/config/samples/backends/cinder/glance-common/glance.yaml
+++ b/config/samples/backends/cinder/glance-common/glance.yaml
@@ -15,7 +15,6 @@ spec:
         [glance_store]
         default_backend = default_backend
         [default_backend]
-        rootwrap_config = /etc/glance/rootwrap.conf
         store_description = Default cinder backend
         cinder_store_auth_address = {{ .KeystoneInternalURL }}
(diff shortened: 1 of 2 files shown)


fultonj:
Let's not merge this until we confirm with @konan-abhi

fmount:
> Let's not merge this until we confirm with @konan-abhi

In previous releases we used also to run glance pointing to `--config-dir /etc/glance/glance.conf.d` so that path is always ignored.
If we merge this patch, we might need to update architectures accordingly [1][2][3].

[1] https://github.com/openstack-k8s-operators/architecture/blob/main/examples/dt/uni05epsilon/service-values.yaml#L103
[2] https://github.com/openstack-k8s-operators/architecture/blob/main/examples/dt/uni07eta/control-plane/service-values.yaml#L46
[3] https://github.com/openstack-k8s-operators/architecture/blob/main/examples/dt/uni06zeta/control-plane/service-values.yaml#L50

fmount:
@konan-abhi confirmed we do not need it. I'm going to land this patch.

fmount (approved):
/lgtm

PR #764, opened 2025-07-16
Add Horizon integration documentation

fmount:
This documentation covers `CORS` configuration for `Horizon`.


docs/horizon.md
```diff
@@ -0,0 +1,71 @@
...
+In general Horizon has different available options.
+
+1. `off`: disables the ability to upload images via Horizon.
+2. `legacy`: enables local file upload by piping the image file through the
+   Horizon’s web server.
+2. `direct` sends the image file directly from the web browser to Glance.
+
+"direct" is the preferred mode, and it requires `CORS` support to be enabled on
+the Glance API service.
+`direct` mode has several benefits:
```
deshipu, line 20: Perhaps it would be prudent to also mention the downside of the direct mode, that is that the Glance endpoint needs to be available outside for it to work.

fmount: maybe we can clarify it, but in general if those are not available `Glance` won't reach a `Setup Complete` state.

konan-abhi (approved):
Looks good!!

docs/horizon.md
deshipu: They could be blocked outside, though? I know that some customers only make the Horizon available to the outside, and not all of the OpenStack APIs. In that case the direct mode wouldn't work.

fmount: Fair enough, but I don't think that the `Route` in k8s is blocked by default, and in order to do that, someone has to write a `NetworkPolicy` to explicitly don't make it reachable from a given `Pod`, which would make any additional note obvious to the human administrator because it might simply result in contradiction.

deshipu: If you think it's obvious, then fine.

PR #785, opened 2025-08-20
docs: Clarify quota behavior when registered limit is 0

fmount:
Improve `Glance` quota `README` explaining that setting a quota limit to `0` means **no limit is enforced for that resource type**. It also includes a `YAML` example showing mixed quota configurations with both limited and unlimited resources.
This reflect the quota strategies defined by the upstream documentation [1].

[1] https://docs.openstack.org/glance/latest/admin/quotas.html#quota-strategies

Related: https://issues.redhat.com/browse/OSPRH-19136
Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/config/samples/quotas/README.md
+++ b/config/samples/quotas/README.md
@@ -44,6 +44,21 @@ The fields above are defined at the top level CR and applied by default to all
 the `GlanceAPI` instances.
 It's not possible setup or override those values for each GlanceAPI instances
 independently.
+
+**Note:** Setting quota limits to `0` has different implications depending on
+the quota type. As per the [OpenStack
+documentation](https://docs.openstack.org/glance/latest/admin/quotas.html#quota-strategies):
+- `imageCountTotal: 0` prevents users from creating any images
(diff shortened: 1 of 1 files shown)


config/samples/quotas/README.md
```diff
@@ -44,6 +44,19 @@ The fields above are defined at the top level CR and applied by default to all
 the `GlanceAPI` instances.
 It's not possible setup or override those values for each GlanceAPI instances
 independently.
+
+**Note:** When a registered limit is set to `0`, it means there is no limit
+enforced for that resource type. This allows unlimited consumption of the
+resource.
+
```
konan-abhi, line 51: I think this will convey wrong message. If imageCountTotal is 0 then user will not be able to create any image at all.
You should replicate the text from upstream documentation as is.

https://docs.openstack.org/glance/latest/admin/quotas.html#quota-strategies

konan-abhi (approved):
Looks good, thank you!!

PR #787, opened 2025-08-20
Automatically configure CORS section based on Horizon CR

fmount:
Add integration with `Horizon` to dynamically set `CORS` `allowed_origin` in `Glance` configuration when `Horizon CR` is present.

In particular:
  - Add Horizon API dependency and RBAC permissions
  - Watch for Horizon CR changes in GlanceAPI controller
  - Fetch Horizon endpoint and inject into config template
  - Update CORS section in glance.conf template
  - Add functional test support for Horizon CRDs

Depends-On: https://github.com/openstack-k8s-operators/horizon-operator/pull/500
Jira: https://issues.redhat.com/browse/OSPRH-19261


test/functional/suite_test.go
```diff
@@ -125,6 +129,7 @@ var _ = BeforeSuite(func() {
 			keystoneCRDs,
 			infraCRDs,
 			cinderCRDs,
+			horizonCRD,
```
konan-abhi, line 132: If need another change, can we rename this variable similar to others here, horizonCRDs

konan-abhi (approved):
This looks good to me, just one suggestion but that is also trivial and can be ignored.

test/functional/suite_test.go
fmount: sure, I will do in the next iteration (when I rebase on top of https://github.com/openstack-k8s-operators/horizon-operator/pull/500)

PR #794, opened 2025-09-08
Fix notificationBusInstance

ratailor:
This change replaces notificationBusInstance references with notificationsBusInstance for consistency with usage in other operators.

### DIFF
--- a/api/bases/glance.openstack.org_glances.yaml
+++ b/api/bases/glance.openstack.org_glances.yaml
@@ -1560,7 +1560,7 @@ spec:
                 description: NodeSelector to target subset of worker nodes running
                   this service
                 type: object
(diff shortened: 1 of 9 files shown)


fmount:
Thank you, but -1 for now.
As I mentioned in https://github.com/openstack-k8s-operators/openstack-operator/pull/1591 we need to coordinate this with a particular release to not break what is already shipped.

fmount:
Clearly the precommit check indicates that we shouldn't change a parameter that we already released in the same API version, and I think I agree. If we decide to go with this, we should override that check, but I need to check this within the team and circle back to it.

fmount:
obsoleted by: https://github.com/openstack-k8s-operators/glance-operator/pull/848

PR #795, opened 2025-09-11
Watch for keystone-overrides

fmount:
This patch implements a watcher over keystone-overrides secrets based on a particular labelSelector. By doing this it is possible to extract and process keystone parameters (auth_url, www_authenticate_uri, region) and override the default information.

- https://github.com/openstack-k8s-operators/lib-common/pull/638
- https://github.com/openstack-k8s-operators/keystone-operator/pull/625


fmount:
Closing as we discussed a different (better) solution

PR #802, opened 2025-09-24
Automatically set s3_store_cacert when S3 and TLS are enabled

fmount:
Add automatic `S3` `CA` certificate injection when `TLS` is enabled, injecting `s3_store_cacert` parameter to point to the CA bundle. This is based on the mechanism provided by `lib-common` `util.ExtendCustomServiceConfig()` that safely adds configuration options without overwriting existing user settings, preserving `customServiceConfig` integrity based on the user input.

Jira: https://issues.redhat.com/browse/OSPRH-14309

### DIFF
--- a/internal/controller/glanceapi_controller.go
+++ b/internal/controller/glanceapi_controller.go
@@ -688,6 +688,7 @@ func (r *GlanceAPIReconciler) reconcileNormal(
 	configVars := make(map[string]env.Setter)
 	privileged := false
 	imageConv := false
+	extConfigOptions := []util.IniOption{}
 
 	//
 	// check for required OpenStack secret holding passwords for service/admin user and add hash to the vars map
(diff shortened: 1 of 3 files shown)


controllers/glanceapi_controller.go
```diff
@@ -758,7 +759,7 @@ func (r *GlanceAPIReconciler) reconcileNormal(
 		}
 	}
 	// iterate over availableBackends for backend specific cases
-	for i := 0; i < len(availableBackends); i++ {
+	for i := range len(availableBackends) {
```
konan-abhi, line 762: I am not sure but range here will not work as len(availableBackends) will return integer.
This should be like

```
for i := range availableBackends {
    backendToken := strings.SplitN(availableBackends[i], ":", 2)
    switch backendToken[1] {
    case "cinder":
        // your code here
    }
}
```

konan-abhi (approved):
Thank you for explaining in detail, looks good to me!!!

PR #812, opened 2025-10-02
Application Credential support

Deydra71:
Jira: [OSPRH-16625](https://issues.redhat.com/browse/OSPRH-16625)

This PR adds end-to-end support for consuming Keystone ApplicationCredentials (AC) in the Glance operator, enabling Glance API pods to use AC-based authentication when available.

Reconcile:

API changes:

Adds an optional authentication field to the Glance API CR:

`spec.auth.applicationCredentialSecret` — name of the Secret that contains the Keystone Application Credential ID and Secret (AC_ID and AC_SECRET).

Reconcile behavior:

Reads `spec.auth.applicationCredentialSecret`
Attempts to load AC_ID / AC_SECRET from the referenced Secret (via the Keystone helper).
If the secret is missing or incomplete, it falls back to password authentication (the AppCred auth is optional, not an error).

Once the AC Secret is ready with valid AC_ID and AC_SECRET fields, templates AC credentials into Glance configuration
Computes hash of Secret contents and stores in configVars to trigger rolling updates when credentials rotate
RBAC permissions:

Added read permissions for secrets resources to allow fetching AC secrets.

Jira: https://issues.redhat.com/browse/OSPRH-22958

Depends-On: https://github.com/openstack-k8s-operators/keystone-operator/pull/567


controllers/glanceapi_controller.go
```diff
@@ -1700,3 +1756,46 @@ func (r *GlanceAPIReconciler) GetHorizonEndpoint(
 
 	return ep, nil
 }
+
+// verifyApplicationCredentials checks if ApplicationCredential secret exists and adds it to configVars
+// The AC secret is created by the keystone-operator's AC controller when the AC is ready.
+// If the secret exists and is valid, we use AC auth. Otherwise, we fall back to password auth.
+func (r *GlanceAPIReconciler) verifyApplicationCredentials(
```
fmount, line 1763: This functions seems very generic to me: if we do not pass `instance *glancev1.GlanceAPI`, but we simply pass a `namespace` or a `serviceName` there would be nothing Glance specific here, and the service could consume it from keystone-operator.
In addition, this might result in a lot of code duplication across the board, so I would invest some time to make it generic and remove the `instance` dependency.

controllers/glanceapi_controller.go
```diff
@@ -1700,3 +1756,46 @@ func (r *GlanceAPIReconciler) GetHorizonEndpoint(
...
+	ctx context.Context,
+	_ *helper.Helper,
```
fmount, line 1765: maybe we should not pass `_ *helper.Helper` if not required

controllers/glanceapi_controller.go
```diff
@@ -1700,3 +1756,46 @@ func (r *GlanceAPIReconciler) GetHorizonEndpoint(
...
+	instance *glancev1.GlanceAPI,
+	configVars *map[string]env.Setter,
+) (ctrl.Result, error) {
+	log := r.GetLogger(ctx)
+
+	// Check if AC secret exists (created by keystone AC controller)
+	acSecretName := keystonev1.GetACSecretName(glance.ServiceName)
```
fmount, line 1772: `ServiceName` is a const that identifies the openstack image service [1] but this does not ensure this code will work if the CR name is different (e.g. `glance-12345678` instead of `glance`).
If we are looking for the current `glanceAPI` name here, we usually expose the `APIName()` function [2], so you can technically pass it to the function and consume it here (this would help to make the function generic and not have dependencies on `instance` type).  

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/pkg/glance/const.go#L34
[2] https://github.com/openstack-k8s-operators/glance-operator/blob/main/api/v1beta1/glanceapi_types.go#L174

controllers/glanceapi_controller.go
```diff
@@ -1700,3 +1756,46 @@ func (r *GlanceAPIReconciler) GetHorizonEndpoint(
...
+	secretKey := types.NamespacedName{Namespace: instance.Namespace, Name: acSecretName}
+
+	hash, res, err := secret.VerifySecret(
+		ctx,
+		secretKey,
+		[]string{"AC_ID", "AC_SECRET"},
+		r.Client,
+		10*time.Second,
```
fmount, line 1780: Note you can reference `10*time.Second` with `glance.NormaDuration` [1]

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/pkg/glance/const.go#L112

test/functional/glanceapi_controller_test.go
```diff
@@ -1286,4 +1289,116 @@ var _ = Describe("Glanceapi controller", func() {
...
+
+				conf := string(cfgSecret.Data["00-config.conf"])
+
+				g.Expect(conf).To(ContainSubstring(
+					"application_credential_id = test-ac-id"),
+				)
+				g.Expect(conf).To(ContainSubstring(
+					"application_credential_secret = test-ac-secret"),
+				)
+			}, 30*time.Second, interval).Should(Succeed())
```
fmount, line 1401: is `timeout` [1] not sufficient here?

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/test/functional/suite_test.go#L70

templates/common/config/00-config.conf
```diff
@@ -41,9 +41,14 @@ default_backend=default_backend
 [keystone_authtoken]
 www_authenticate_uri={{ .KeystonePublicURL }}
 auth_url={{ .KeystoneInternalURL }}
-auth_type=password
+auth_type={{ if .UseApplicationCredentials }}v3applicationcredential{{ else }}password{{ end }}
```
fmount, line 44: to keep things cleaner here I would try:
```
{{ if .UseApplicationCredentials -}}
auth_type = v3applicationcredential
application_credential_id = {{ .ACID }}
application_credential_secret = {{ .ACSecret }}
{{ else -}}
auth_type = password
username={{ .ServiceUser }}
{{- end }}
```
your version is correct, but this would improve readability, other than saving an `if`

templates/common/config/00-config.conf
```diff
@@ -94,11 +107,16 @@ filesystem_store_datadir = /var/lib/glance/os_glance_tasks_store/
 
 [oslo_limit]
 auth_url={{ .KeystoneInternalURL }}
-auth_type = password
+auth_type = {{ if .UseApplicationCredentials }}v3applicationcredential{{ else }}password{{ end }}
```
fmount, line 110: same here

fmount (changes requested):

fmount:
Also @Deydra71 because we deploy keystone in kuttl as a dependency, we should add this scenario to ensure some coverage.

internal/controller/glanceapi_controller.go
```diff
@@ -1285,6 +1297,30 @@ func (r *GlanceAPIReconciler) generateServiceConfig(
...
+			return err
+		}
+		acID, okID := acSecretObj.Data[keystonev1.ACIDSecretKey]
+		acSecretData, okSecret := acSecretObj.Data[keystonev1.ACSecretSecretKey]
+		if !okID || len(acID) == 0 || !okSecret || len(acSecretData) == 0 {
+			Log.Info("ApplicationCredential secret missing required keys", "secret", instance.Spec.Auth.ApplicationCredentialSecret)
+			return fmt.Errorf("%w: %s", ErrACSecretMissingKeys, instance.Spec.Auth.ApplicationCredentialSecret)
+		}
+		templateParameters["UseApplicationCredentials"] = true
+		templateParameters["ACID"] = string(acID)
```
fmount: @Deydra71 I'm ok with this implementation, and in theory we can land as it is.
I have a question though.
In cinder (and probably other operators), we have:
```
templateParameters["ApplicationCredentialID"] = string(acID)
templateParameters["ApplicationCredentialSecret"] = string(acSecretData)
```

In the other implementation we also do not use a boolean (seems redundant), so we assume that if an `ApplicationCredentialID` exists we render the appropriate parameters in the template.

For reference, here's the cinder implementation:

```golang

// Retrieve Application Credential data from Auth section if specified
	Log := r.GetLogger(ctx)
	if instance.Spec.Auth.ApplicationCredentialSecret != "" {
		acSecret, _, err := secret.GetSecret(ctx, h, instance.Spec.Auth.ApplicationCredentialSecret, instance.Namespace)
		if err != nil {
			if k8s_errors.IsNotFound(err) {
				Log.Info("ApplicationCredential secret not found, waiting", "secret", instance.Spec.Auth.ApplicationCredentialSecret)
				return fmt.Errorf("%w: %s", ErrACSecretNotFound, instance.Spec.Auth.ApplicationCredentialSecret)
			}
			Log.Error(err, "Failed to get ApplicationCredential secret", "secret", instance.Spec.Auth.ApplicationCredentialSecret)
			return err
		}
		acID, okID := acSecret.Data[keystonev1.ACIDSecretKey]
		acSecretData, okSecret := acSecret.Data[keystonev1.ACSecretSecretKey]
		if okID && len(acID) > 0 && okSecret && len(acSecretData) > 0 {
			templateParameters["ApplicationCredentialID"] = string(acID)
			templateParameters["ApplicationCredentialSecret"] = string(acSecretData)
			Log.Info("Using ApplicationCredentials auth", "secret", instance.Spec.Auth.ApplicationCredentialSecret)
		} else {
			return fmt.Errorf("%w: %s", ErrACSecretMissingKeys, instance.Spec.Auth.ApplicationCredentialSecret)
		}
	}
```
Should we double check the implementation and make sure we have consistency across the board?

Deydra71: Sadly we have inconsistency with this right now. Some of the App Cred patches merged already with the usage of  boolean flag.  And some of them use the presence of AC data (cinder, manila, octavia, watcher). It's not critical issue, more about inconsistency only. I think as well that the boolean is redundant - I can change in glance. But for other operators that are merged already with the boolean (neutron, barbican, designate...) we can only do follow ups.

PR #878, opened 2026-02-18
Probes configuration with override support

fmount:
Add `ProbeOverrides` interface and `CreateProbeSet()` function from `lib-common` for unified probe management across services. Enable probe customization through `CRD` overrides and remove code duplication. Updates all services to use the new pattern with proper scheme handling and consistent defaults.
In addition, `webhook` validation for probes have been introduced.

Jira: [OSPRH-29162](https://redhat.atlassian.net/browse/OSPRH-29162)


fmount:
recheck

api/v1beta1/glanceapi_types.go
```diff
@@ -19,10 +19,10 @@ package v1beta1
 import (
 	"fmt"
 
+	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
```
konan-abhi, line 22: nit:

is this change on purpose like moving it to top or it doesn't make any difference if we didn't change this?

docs/probes.md
```diff
@@ -0,0 +1,223 @@
...
+      startupProbes:
+        timeoutSeconds: 10
+        periodSeconds: 5
+        initialDelaySeconds: 5
+        failureThreshold: 20
+```
+
+### Disabling Startup Probes
+
+Set to `nil` to disable startup probes (not recommended):
```
konan-abhi, line 153: null?

konan-abhi (approved):
Looks good, few nits noted inline but those can be ignored!!

api/v1beta1/glanceapi_types.go
fmount: this is the `linter` fixing things (like reordering imports)

docs/probes.md
fmount: I'll fix this, thanks for catching. In theory we don't want to disable this so I'm going to remove this section in a follow up patch. we should always rely on the defaults.
https://github.com/openstack-k8s-operators/glance-operator/pull/907

PR #897, opened 2026-04-08
Add application credential finalizer management

Deydra71:
Jira: [OSPRH-27509](https://redhat.atlassian.net/browse/OSPRH-27509)

Application Credential dev-doc: https://github.com/openstack-k8s-operators/dev-docs/blob/main/application_credentials.md

* Tracks the active AC secret name in `Status.ApplicationCredentialSecret`
* Add `openstack.org/glance-ac-consumer` finalizer to the AC secret after service config is rendered
* On AC rotation, move the finalizer from the old secret to the new one
* On CR deletion, remove the consumer finalizer from the AC secret before cleaning up the CR

This ensures that the keystone-operator cannot revoke a rotated AC secret while Glance is still consuming it.

```
2026-04-28T11:49:19Z	INFO	Controllers.GlanceAPI	Added consumer finalizer	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "GlanceAPI": {"name":"glance-default-external","namespace":"openstack"}, "namespace": "openstack", "name": "glance-default-external", "reconcileID": "69f308d9-956f-455e-953a-f87848e4a2e9", "object": "ac-glance-cec48-secret", "finalizer": "openstack.org/glanceapi-default-ac-consumer"}
2026-04-28T11:49:19Z	INFO	Controllers.GlanceAPI	Removed consumer finalizer	{"controller": "glanceapi", "controllerGroup": "glance.openstack.org", "controllerKind": "GlanceAPI", "GlanceAPI": {"name":"glance-default-external","namespace":"openstack"}, "namespace": "openstack", "name": "glance-default-external", "reconcileID": "69f308d9-956f-455e-953a-f87848e4a2e9", "object": "ac-glance-79649-secret", "finalizer": "openstack.org/glanceapi-default-ac-consumer"}

```

Depends-On: https://github.com/openstack-k8s-operators/keystone-operator/pull/685

Assisted-by: Claude Opus 4.6 [noreply@anthropic.com](mailto:noreply@anthropic.com)


internal/controller/glanceapi_controller.go
```diff
@@ -989,6 +995,15 @@ func (r *GlanceAPIReconciler) reconcileNormal(
 	// we can mark the ServiceConfigReady as True and rollout the new pods
 	instance.Status.Conditions.MarkTrue(condition.ServiceConfigReadyCondition, condition.ServiceConfigReadyMessage)
 
+	// Manage AC consumer finalizer, the AC data was already read and rendered to the service config
+	if err := keystonev1.ManageACSecretFinalizer(ctx, helper, instance.Namespace,
```
fmount: should we call this block conditionally? if `instance.Spec.Auth.ApplicationCredentialSecret` is not set, we could skip this block (until L1005) entirely.
I'm also wondering if we should detect issues when `err != nil` and reflect it int the `.Status` through a dedicated condition.

internal/controller/glanceapi_controller.go
```diff
@@ -494,6 +494,12 @@ func (r *GlanceAPIReconciler) reconcileDelete(ctx context.Context, instance *gla
 		return ctrlResult, err
 	}
 
+	// Remove consumer finalizer from the AC secret GlanceAPI was consuming.
+	if err := keystonev1.RemoveACSecretConsumerFinalizer(ctx, helper, instance.Namespace,
```
fmount: The concern here is that you might have a misaligned `.Spec.Auth.ApplicationCredentials` with `.Status.ApplicationCredentialSecret`. On a rotation, if the reconciliation stops before being able to assign the `LastAppliedCredentialSecret`, and you delete the CR (e.g. `oc delete glance glance`), the finalizer might not be removed from the last created secret.
To be honest this is just theory, I didn't test it, but I'm wondering if we should take a note here for a follow up where we introduce or check `ObservedGeneration` in the reconciliation.

internal/glance/const.go
```diff
@@ -106,6 +106,8 @@ const (
 	GlanceCacheCleaner = "/usr/bin/glance-cache-cleaner"
 	// GlanceCachePruner -
 	GlanceCachePruner = "/usr/bin/glance-cache-pruner"
+	// ACConsumerFinalizer is added to AC secrets that GlanceAPI is actively consuming
+	ACConsumerFinalizer = "openstack.org/glanceapi-ac-consumer"
```
fmount: This is weird to me, and I have to check it more. 
In the past I used to add/remove/update finalizers using `controllerutil` (see [1] as an example).
That way you can simply pass to the function exposed by keystone an `helper` and `instance.APIName()`, and you don't need to define a constant because the group is inferred by `controllerutil`.
However I didn't see the keystone function exposed here, so I might be totally wrong.
I'll check that part.


[1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/topology/v1beta1/topology_types.go#L166

internal/controller/glanceapi_controller.go
fmount: @Deydra71 after a better look into this, I think we can save us from setting up a new condition for the `finalizer`, but we could reuse `condition.ServiceConfigReadyCondition`.
The `finalizer` management runs after `ServiceConfigReadyCondition` is marked to `true` (L996). If `ManageACSecretFinalizer` fails, we technically marked the condition as true but the `finalizer` wasn't applied. This might generate a problem, so I was thinking about moving L996 after L1005, and setting `ServiceConfigReady` to `false` if there's a problem at this stage (L1003). 

What do you think?

internal/glance/const.go
fmount: @Deydra71 also using a constant in the form `glanceapi-ac-consumer` might be problematic for two reasons:
1. we use to have <group>/<service> pattern, and we're breaking it
2. assuming you have multiple glance APIs (which is pretty common in DCN, DZ and other custom topologies), you can technically use the same appCred for all of them by referencing the same secret. This can create a problem because you lose track of which api is using that secret from a finalizer perspective, and if I remove one of these APIs, I remove this constant finalizer from the secret for all of them, defeating the purpose of this patch.

Scenario#2 seems a blocker to me, and this is why we should use `instance.APIName()` as a finalizer for a given secret.

fmount (changes requested):

Deydra71:
Hi @fmount ! Thank you the revieww.

 I didn't initially realize that Glance creates multiple GlanceAPI instances per CR, so the original static `ACConsumerFinalizer` was incorrect. Now it's replaced with `ACConsumerFinalizerName(instance.APIName())` which generates `openstack.org/glanceapi-<apiName>-ac-consumer`

As for the guard finalizer block - `ManageACSecretFinalizer` is now wrapped in: 
```
if  instance.Spec.Auth.ApplicationCredentialSecret != "" || instance.Status.ApplicationCredentialSecret != "" 
``` 
so we skip the call when there's no AC secret.

Reordered ServiceConfigReadyCondition - moved `MarkTrue(ServiceConfigReadyCondition)` to after the finalizer block. If ManageACSecretFinalizer fails, we now explicitly set the condition to False with the error.

Removed finalizer from both spec and status secrets on delete - `reconcileDelete` now removes the consumer finalizer from both `Status.ApplicationCredentialSecret` and `Spec.Auth.ApplicationCredentialSecret`, handling the crash edge case. I t hink this is sufficient fix for the delete path for now, without the `ObservedGeneration` check.

I will also incorporate last three mentioned fixes to barbican, manila, cinder and swift operators since it's affecting them the same way.

internal/glance/funcs.go
```diff
@@ -1,13 +1,22 @@
...
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/utils/ptr"
 	"math"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// ACConsumerFinalizerName returns a per-GlanceAPI finalizer name so that
+// multiple GlanceAPI instances (e.g. internal, external) sharing the same
+// AC secret each get their own distinct finalizer on the secret.
+func ACConsumerFinalizerName(apiName string) string {
```
fmount: @Deydra71 It shouldn't change much, but I'm wondering if, instead of having this function here, we should move it to the API layer (see at the very bottom of this file [1]).
This way we could:
1. have the constant `ACConsumerFinalizerPrefix` encoded in the const block of the API
2. we can invoke this function like `instance.ACConsumerFinalizerName()`:

```golang
func (instance *GlanceAPI) ACConsumerFinalizerName() string {
    return fmt.Sprintf("%s%s-ac-consumer", ACConsumerFinalizerPrefix, instance.APIName())
}
```
Moving it to the API type as a method seems cleaner as ties the finalizer to the instance that owns it, and you call it as `instance.ACConsumerFinalizerName()` instead of passing `instance.APIName()`.


[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/api/v1beta1/glanceapi_types.go

Deydra71: Makes perfect sense, will do. Thanks!

internal/controller/glanceapi_controller.go
```diff
@@ -1184,6 +1215,25 @@ func (r *GlanceAPIReconciler) reconcileNormal(
 	)
 	// create ImageCache cronJobs - end
 
+	// Manage the old AC secret's finalizer and status tracking.
+	// On rotation (old != new), only remove the old secret's finalizer after
+	// all sub-services are ready with the new credentials. This prevents
+	// premature revocation during rapid rotations.
+	isRotation := instance.Status.ApplicationCredentialSecret != "" && instance.Status.ApplicationCredentialSecret != instance.Spec.Auth.ApplicationCredentialSecret
+
+	if isRotation {
```
fmount, line 1224: I like this part because even if is a bit verbose the purpose is very clear and is easy to read.

internal/controller/glanceapi_controller.go
```diff
@@ -987,6 +1000,24 @@ func (r *GlanceAPIReconciler) reconcileNormal(
...
+	// Add consumer finalizer to the new AC secret early, before deployment.
+	// The old secret's finalizer is removed later (after all services deploy)
+	// so that rapid rotations don't revoke a credential still in use by pods.
+	if instance.Spec.Auth.ApplicationCredentialSecret != "" {
+		if err := keystonev1.ManageACSecretFinalizer(ctx, helper, instance.Namespace,
+			instance.Spec.Auth.ApplicationCredentialSecret,
+			"",
+			instance.ACConsumerFinalizerName()); err != nil {
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				condition.ServiceConfigReadyCondition,
```
fmount, line 1012: Ok I think it is clear here that in case of failure we rely on `ServiceConfigReadyCondition` and `ServiceConfigReadyMessage`. As per the meaning of `ServiceConfigReady` it is ok to include the result of the keystone API call.

internal/controller/glanceapi_controller.go
```diff
@@ -496,6 +496,19 @@ func (r *GlanceAPIReconciler) reconcileDelete(ctx context.Context, instance *gla
...
 	}
 
+	// Remove consumer finalizer from AC secrets GlanceAPI was consuming.
+	// Check both status and spec to handle the edge case where the reconciler
+	// crashed after adding the finalizer but before updating the status.
+	for _, secretName := range []string{
+		instance.Status.ApplicationCredentialSecret,
+		instance.Spec.Auth.ApplicationCredentialSecret,
+	} {
+		if err := keystonev1.RemoveACSecretConsumerFinalizer(ctx, helper, instance.Namespace,
```
fmount, line 506: I was worried about the double call here but I see the function embeds the check for `secretName == ""`, resulting in a `nop` [1]. +1 on this part as well.

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/main/api/v1beta1/keystoneapplicationcredential.go#L121

api/v1beta1/common_types.go
```diff
@@ -29,6 +29,10 @@ import (
 )
 
 const (
+	// ACConsumerFinalizerPrefix is the base prefix for the per-GlanceAPI AC consumer finalizer.
+	// Each GlanceAPI instance gets its own finalizer: <prefix><apiName>-ac-consumer
+	ACConsumerFinalizerPrefix = "openstack.org/glanceapi-"
```
fmount: nit: this might go to  `api/v1beta1/glanceapi_types.go` but not worth iterating again just for this, because in the end it's the same package.

test/functional/glanceapi_controller_test.go
```diff
@@ -1669,4 +1669,149 @@ var _ = Describe("Glanceapi controller", func() {
...
+
+			Eventually(func(g Gomega) {
+				secret := th.GetSecret(types.NamespacedName{
+					Namespace: glanceTest.Instance.Namespace,
+					Name:      newACSecretName,
+				})
+				g.Expect(secret.Finalizers).To(
+					ContainElement(glancev1.ACConsumerFinalizerPrefix + "default-ac-consumer"))
+			}, timeout, interval).Should(Succeed())
+
```
fmount, line 1777: @Deydra71 do you think it makes sense to add a test scenario where we remove the appCred entirely? In theory if we set `spec.auth.applicationCredentialSecret = ""` we should see that:

1. the secret's `finalizer` gets removed (after services are ready again because of the new Pods rollout)
2. `.status.applicationCredentialSecret` is cleared to an empty string (or not visible entirely because is set to `nil`)
3. the glanceAPI config falls back to password

Is this a valid scenario? If not entirely, it could in any case test the finalizer removal in a use case where no rotation is happening.

test/functional/glanceapi_controller_test.go
```diff
@@ -1669,4 +1669,149 @@ var _ = Describe("Glanceapi controller", func() {
...
+			Eventually(func(g Gomega) {
+				secret := th.GetSecret(types.NamespacedName{
+					Namespace: glanceTest.Instance.Namespace,
+					Name:      acSecretName,
+				})
+				g.Expect(secret.Finalizers).NotTo(
+					ContainElement(glancev1.ACConsumerFinalizerPrefix + "default-ac-consumer"))
+			}, timeout, interval).Should(Succeed())
+		})
+	})
```
fmount, line 1852: @Deydra71 I had an important test in mind, and probably it should go in `glance_controller_test.go`.
Basically we should try to create a CR with at least  two `GlanceAPIs` instances sharing the same AC secret (resulting in `glanceapi-<apiName>-ac-consumer` on the same secret), and verifying that deleting one instance only removes its own finalizer while the other's stays.
We can make this in a follow up if you have no capacity and we need to merge this code very soon, but worth checking if it makes sense to you and if we want to add it in a follow up.

internal/glance/funcs.go
fmount: Thank you

test/functional/glanceapi_controller_test.go
Deydra71: Yes, this and the scenario above are both valid. I will look at them today.

Deydra71: After thinking about this more the `ACConsumerFinalizerName()` uses `APIName()` which returns the shared map key, and not the API name, so it results in 1 shared finalizer `openstack.org/glanceapi-default-ac-consumer`. 

@fmount Does this change makes sense to achieve one finalizer per API type?

```
func (instance GlanceAPI) ACConsumerFinalizerName() string {
	return fmt.Sprintf("%s%s-%s-ac-consumer", ACConsumerFinalizerPrefix, instance.APIName(), instance.Spec.APIType)
}
```

fmount: right, and when you have `type: split` resulting in `glance-default-internal` and `glance-default-external` the finalizer set/unset might be problematic because you have to worry about coordinating it per API. Instead, if we add the `APIType`, it results self-contained and works for each API instance.
See [1] for example, where we use `fmt.Sprintf("%s-%s-%s", glance.ServiceName, instance.APIName(), instance.Spec.APIType)` to identify that particular glanceAPI.
I think we should make this change otherwise you risk that the controller for an API -external removes the finalizer for a -internal API, resulting in inconsistencies in the logic you provided. The best choice here is to let each controller manage the lifecycle of each API type, hence use the pattern [1] to set/unset finalizers.
When you deploy a glanceAPI with `type:split` (the default in 18), you need to end up with:
1. openstack.org/glanceapi-glance-default-internal-ac-consumer
2. openstack.org/glanceapi-glance-default-external-ac-consumer

(technically we could remove from the prefix `glanceapi` since it's redundant).

When `APIType` is `single` (`type: single`) or `edge` (`type:edge`), you need to end up with:

1. openstack.org/glanceapi-glance-default-single-ac-consumer
2. openstack.org/glanceapi-glance-default-edge-ac-consumer

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/internal/glanceapi/funcs.go#L54

api/v1beta1/common_types.go
fmount: @Deydra71 `glanceapi-` seems redundant if we set if by `APIName` (e.g. glance-default-internal / glance-default-externale, glance-az1-edge, etc)

test/functional/glanceapi_controller_test.go
fmount: I think we're almost there. We might want to probably look at this test as an extension of what is already in place, otherwise we should be ready.

Deydra71: I added it together with the latest change - https://github.com/openstack-k8s-operators/glance-operator/pull/897/changes#diff-ed8e9966ce93f97dca999445ada7c33ea237e447141c7a1cafa993d455523a24R1817

Or you had something else in mind to add as a test scenario?

fmount: ah ok sorry I kind of missed that. let me check. I think falling back to password is implicit and it is sufficient verify that appCred is gone from the status.

fmount (approved):
/lgtm

PR #903, opened 2026-04-21
Use OpenStack CR operatorOverrides to scale down glance operator

karelyatin:
Switch from CSV-based scaling to using OpenStack CR operatorOverrides
  feature for scaling down the glance operator. This avoids the need
  to disable the openstack init operator and provides better integration
  with the OpenStack operator management system.

### DIFF
--- a/hack/run_with_local_webhook.sh
+++ b/hack/run_with_local_webhook.sh
@@ -176,16 +176,14 @@ if [ -n "${CSV_NAME}" ]; then
     oc patch "${CSV_NAME}" -n openstack-operators --type=json -p="[{'op': 'replace', 'path': '/spec/install/spec/deployments/0/spec/replicas', 'value': 0}]"
     oc patch "${CSV_NAME}" -n openstack-operators --type=json -p="[{'op': 'replace', 'path': '/spec/webhookdefinitions', 'value': []}]"
 else
-    # Handle operator deployed by Openstack Initialization resource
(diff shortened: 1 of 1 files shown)


hack/run_with_local_webhook.sh
```diff
@@ -176,16 +176,14 @@ if [ -n "${CSV_NAME}" ]; then
...
-
-    printf \
-    "\n\tNow patching openstack operator CSV to scale down deployment resource.
-    To restore it, use:
-    oc patch "${CSV_NAME}" -n openstack-operators --type=json -p=\"[{'op': 'replace', 'path': '/spec/install/spec/deployments/0/spec/replicas', 'value': 1}]\""
+    if CR_NAME=$(oc get openstack -n openstack-operators -o name); then
+        printf \
+        "\n\tNow patching openstack CR to scale down deployment resource.
+        To restore it, use:
+        oc patch ${CR_NAME} -n openstack-operators --type=merge -p '{\"spec\": {\"operatorOverrides\": [{\"name\": \"glance\", \"replicas\": 1}]}}'\n"
```
fmount, line 183: nit: double quote `${CR_NAME}`, while we know what the output would be, so I don't expect this to raise an error.

fmount (commented):
/lgtm

hack/run_with_local_webhook.sh
karelyatin: Yes it works fine

fmount (approved):

PR #925, opened 2026-06-01
Fix cache cleaner/pruner affinity to target exact replica

fmount:
The cache `CronJob` pods used a broad `podAffinity` selector matching all replicas of a `GlanceAPI` (via the `glanceAPI` label), which allowed the scheduler to place a cleaner/pruner pod on a node hosting a different replica than the one whose `RWO` `PVC` it references. This causes "multi-attach" volume errors and pods stuck in `ContainerCreating`. Replace the `glanceAPI` label selector with a `statefulset.kubernetes.io/pod-name` match derived from the `PVC` name, ensuring each `CronJob` is colocated with the exact `StatefulSet` Pod that owns the target volume.

Jira: [OSPRH-30553](https://redhat.atlassian.net/browse/OSPRH-30553)


konan-abhi (approved):
Looks good, thank you!!

fultonj (approved):
/approve
/lgtm

PR #933, opened 2026-06-25
Set service_type in [keystone_authtoken] for access rule validation

Deydra71:
Without service_type configured, keystonemiddleware cannot validate application credentials with custom access rules, causing HTTP 401 for end users.

Closes: [OSPRH-22365](https://redhat.atlassian.net/browse/OSPRH-22365)

### DIFF
--- a/templates/common/config/00-config.conf
+++ b/templates/common/config/00-config.conf
@@ -64,6 +64,7 @@ memcache_tls_keyfile = {{ .MemcachedAuthKey }}
 memcache_tls_cafile = {{ .MemcachedAuthCa }}
 memcache_tls_enabled = true
 {{end}}
+service_type = image
 {{ if (index . "Region") -}}
 region_name = {{ .Region }}
 {{ end -}}


fmount (approved):
/lgtm

templates/common/config/00-config.conf
```diff
@@ -64,6 +64,7 @@ memcache_tls_keyfile = {{ .MemcachedAuthKey }}
 memcache_tls_cafile = {{ .MemcachedAuthCa }}
 memcache_tls_enabled = true
 {{end}}
+service_type = image
```
fmount, line 67: @Deydra71 before merging I was wondering if it makes sense to propagate [1] via `templateParameters`.
In addition looks like `ServiceType` is what we use in the first place to call the keystoneAPI and register a new Service [2].

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/internal/glance/const.go#L36
[2] https://github.com/openstack-k8s-operators/glance-operator/blob/main/internal/controller/glance_controller.go#L476

fmount: @Deydra71 I won't paste the same comment everywhere but worth double checking this part for other operators as well.

Deydra71: Thanks @fmount ! I raised your question in nova --> https://github.com/openstack-k8s-operators/nova-operator/pull/1136#issuecomment-4799674240

With that I think we should keep it hardcoded so it's unified across operators, wdyt?

fmount: ok, let's keep consistency and hardcode it everywhere. we'll folllow up if we really need to!

PR #946, opened 2026-07-16
Allow file backend in multistore configurations

fmount:
The file (`NFS`) backend is now only blocked when it is the **only** store configured with a `split` layout. `Multistore` configurations that include `NFS` alongside other backends (`ceph`, `s3`, etc.) are permitted to support backend migration workflows.

Jira: [OSPRH-32965](https://redhat.atlassian.net/browse/OSPRH-32965)


konan-abhi (approved):
looks good!!

PR #949, opened 2026-07-17
Read admin password from secure.yaml instead of hardcoding it

stuggi:
The install_yamls PR #1158 (https://github.com/openstack-k8s-operators/install_yamls/pull/1158) replaced hardcoded passwords with dynamically generated per-service secrets. The sample scripts used a hardcoded fallback of "12345678" which no longer matches the actual deployment password, breaking the glance_dist_image_import kuttl test.

Read the password from /etc/openstack/secure.yaml (already mounted in the openstackclient pod) instead.

Jira: [OSPRH-32754](https://redhat.atlassian.net/browse/OSPRH-32754)


fmount (approved):
/approve
/lgtm

PR #955, opened 2026-07-24
Remove unnecessary pods RBAC permissions

stuggi:
The workload rbacRules and a kubebuilder RBAC marker granted full CRUD (create/delete/get/list/patch/update/watch) on core Pods, which was never exercised. Remove the unused permission and regenerate config/rbac/role.yaml.

The narrower get;list marker on glanceapi_controller.go is kept, since it reads a Pod directly to associate it with the image-cache PVC.


abays (approved):
/lgtm

PR #969, opened 2026-08-03
Include ssl.conf in httpd configuration

fmount:
The `ssl.conf` rendered by `lib-common` was placed by `kolla` into `conf.d/` but never loaded because the Include `conf.d/*.conf` directive was commented out. This left global `SSL` hardening settings (cipher suite, protocol restrictions, session cache) at `mod_ssl` defaults instead of the operator-managed values.

Jira: https://redhat.atlassian.net/browse/OSPRH-34152


stuggi (approved):
/lgtm

PR #976, opened 2026-08-11
Add transport URL secret rotation with consumer finalizer

lmiccini:
## What this does

When infra-operator rotates a RabbitMQ transport URL (creating a new secret and user), consumer operators must hold a **consumer finalizer** on the old secret until all their pods have rolled out with the new credentials. Without this, infra-operator cleans up the old RabbitMQ user while pods are still connected with the old credentials, causing message-bus outages.

## Approach (updated)

> **Note:** this PR was reworked from the original grace-period design to an event-driven "applied input hash" design. The old approach released the finalizer after a fixed grace period based on the parent `Ready` condition, which was racy against the informer cache and could revoke credentials prematurely. The new approach releases the finalizer only once every child has provably rolled out.

1. Add a consumer finalizer to the current transport URL secret early in reconcile. `Status.TransportURLSecret` is set for first-time setup only (empty or unchanged); during rotation the status is updated solely by `FinalizeSecretRotation` at the end of reconcile.

2. Pass `transportURL.Status.SecretName` directly to sub-CR creation and config generation as a parameter — never read `Status.TransportURLSecret` for sub-CR specs.

3. Each child controller records an `AppliedInputSecretHash` in its status, set only after `statefulset.IsReadyForInput` / `deployment.IsReadyForInput` confirms — via an uncached API read — that the workload is fully rolled out with the expected `CONFIG_HASH`.

4. The parent mirrors a child's Ready condition only when its `Generation == ObservedGeneration` and `AppliedInputSecretHash` matches the current input hash; otherwise it sets the condition to Unknown.

5. Guard: `FinalizeSecretRotation` removes the consumer finalizer from the old secret only when every child reports the expected hash and is ready. The guard is computed from `Conditions.AllSubConditionIsTrue()` (not `IsReady()`), because `Conditions.Init()` resets the `Ready` condition to Unknown on every reconcile.

The same pattern applies to notification transport URL secrets and application-credential secrets where applicable.

## Dependency

Depends on the lib-common `IsReadyForInput` / `FinalizeSecretRotation` helpers (currently pinned via a `replace` to the fork commit while the lib-common PR is in review).


lmiccini:
recheck

internal/controller/glance_controller.go
```diff
@@ -611,11 +647,31 @@ func (r *GlanceReconciler) reconcileNormal(ctx context.Context, instance *glance
 				condition.NotificationBusInstanceReadyRunningMessage))
 			return glance.ResultRequeue, nil
 		}
+
+		// Set status early for first-time setup so PatchInstance persists it
+		// even on early returns. During rotation (old != current), the status
+		// is only updated by FinalizeSecretRotation at end of reconcile.
+		if instance.Status.NotificationBusSecret == "" ||
+			instance.Status.NotificationBusSecret == currentNotificationSecret {
```
stuggi: does this make sense if we set it to the same in the condition? should it just be `instance.Status.NotificationBusSecret == ""` ?

stuggi:
@coderabbitai full review

stuggi (approved):
/lgtm

